Issue #41, opened 2022-08-19
Extension: Build pipeline and base implementation


jurevans:
This issue is to establish the basic underlying functionality of the extension, and ensure that everything is building properly to be tested on Chrome and Firefox.

See related specs: https://github.com/anoma/specs/pull/114

## Requirements

- Extension should build correctly for both Chrome (using `manifest_version: 3`) and Firefox (using `manifest_version: 2`)
- Extension should be able to load wasm functionality into the service worker
- A script should be injected into the page exposing an API which the interface can utilize
- Extension should have messaging in place (`postMessage`) to enable communication between the injected script, content script, and the background service worker/script
- Extension's service worker (or background script in Firefox) should be able to store and retrieve data from IndexedDB
- We should have test scripts available for the extension and related packages

## References

- See `polkadot{js} extension` - This is the experience we wish to replicate <https://github.com/polkadot-js>
- See `keplr-wallet` - We want to be able to query accounts/signers in a similar manner to this (hopefully with a slightly more elegant API of our choosing) <https://github.com/chainapsis/keplr-extension>
- See `keplr-wallet/packages/store` for Key/Value store Promised-based store (IndexedDB) - This can be reused to actually store the data

PR #48, opened 2022-08-31
feat/41 - Extension: Base implementation

jurevans:
See https://github.com/anoma/namada-interface/issues/41

The purpose of this PR is to set up the build and basic implementation of the browser extension. 

See the [https://github.com/anoma/specs/pull/114](specs) as well for more information on the design and implementation details.

## Tasks

- [x] Be able to build extension for Chrome and Firefox (build pipeline should be present for a React app, background and content scripts)
- [x] With the extension loaded, we should have a `window.anoma` class providing methods to a web client
- [x] Calling methods on `window.anoma` should proxy these calls to the `browser.runtime` via `content` scripts
- [x] Methods in `content` scripts should be able to message `background` scripts
- [x] `background` should be able to route messages back to the web client
- [x] All dependencies of the extension from `namada-interface` should be included and building properly
- [x] WASM functionality should work within extension's service worker/background script
  - **note** - To begin, I've split up functionality in `@anoma/wasm` to move all crypto functions to their own lib (`@anoma/crypto`), thus reducing the asset size (the full build of `@anoma/wasm` exceeded 1.2MB). I also began porting aspects of `@anoma/wasm` to `@anoma/shared` to better indicate that this is interfacing with the `namada/shared` code base (which should be its sole responsibility)
  - When moving crypto and shared functionality from `@anoma/wasm` into separate packages, I also updated our code to support `namada` `0.7.1` (as well as tweaked the build script to better support running in a ServiceWorker). As such, the code in `@anoma/shared` does not have the imported MASP types (the `shielded` parameter is unavailable on the `Transfer` struct in `0.7.1`). The `@anoma/wasm` package is still present in its entirety, supporting `namada-interface` with MASP.
  - See https://developer.chrome.com/docs/extensions/whatsnew/#m102-wasm - As of this release, wasm is allowed in manifest v3 extensions in Chrome
- [x] Document API (`window.anoma`), begin defining additional requirements (See <https://github.com/anoma/specs/pull/114>), in addition to documenting the messaging system between client, `content` and `background` scripts

## Notes

Note that Firefox does not support Manifest V3, and Chrome as stopped accepting new Manifest V2 extensions:

> As of January 17, 2022 the Chrome Web Store has stopped accepting new Manifest V2 extensions.

See:
- <https://developer.chrome.com/docs/extensions/mv3/mv2-sunset/>
- <https://developer.chrome.com/docs/extensions/mv2/security/>

As such, the build process in this PR uses a `manifest_version: 2` for Firefox and `manifest_version: 3` for Chrome.

----------------

This PR follows a similar approach to the `keplr-wallet` implementation, where the requirements align. 

See the following `keplr-wallet` packages for more information:

- [extension](https://github.com/chainapsis/keplr-wallet/tree/master/packages/extension)
- [background](https://github.com/chainapsis/keplr-wallet/tree/master/packages/background)
- [provider](https://github.com/chainapsis/keplr-wallet/tree/master/packages/provider)
- [router](https://github.com/chainapsis/keplr-wallet/tree/master/packages/extension)
- [router-extension](https://github.com/chainapsis/keplr-wallet/tree/master/packages/router-extension)

We similarly incorporate the `webextension-polyfill`, and as such, we can only use this polyfill for within the `browser` (or `chrome`) namespace, invoking from within the `content` or `background` scripts of the extension. Therefore, reusable components should be careful to abstract away the calls to any `browser` API for injected Web scripts, and any tests requiring access to this API should make use of `jest-webextension-mock` or `mockzilla-webextension`. As Chrome `manifest_version: 3` requires `background-scripts` to be loaded as a ServiceWorker, we do not have access to the `window` namespace (whereas v2 did), so this should be avoided as it will prevent the ServiceWorker from loading.

## Testing

Run unit tests:

```bash
# Note: There aren't any tests enabled, as I still need to implement mocks for the `ExtensionRequester`, 
# as well as the background services)
yarn test
```

Build extension files:

```bash
yarn build
# See files in ./build/{chrome|firefox}
yarn build:chrome
yarn build:firefox
```

Run development version:

```bash
yarn start
```

### TODO

- Before release, we should take a good look at our injected script sizes. See <https://github.com/chainapsis/keplr-wallet/issues/164> for insight into `keplr-wallet`'s approach. I'm not concerned about the size of the service worker/background scripts at this point, but the `injected.anoma.js` will get added to and executed on each page, and this could create some performance issues if this becomes too large.


apps/extension/src/popup/index.ts
```diff
@@ -0,0 +1,3 @@
+import App from "../App";
```
jurevans, line 1: We may not use a popup, but instead inject our app into the page. This is here mainly to confirm that the build-pipeline is functioning as it should.

apps/extension/src/background/chains/service.ts
```diff
@@ -0,0 +1,100 @@
...
+type ChainRemovedHandler = (chainId: string, identifier: string) => void;
+
+export class ChainsService {
+  protected onChainRemovedHandlers: ChainRemovedHandler[] = [];
+
+  constructor(
+    protected readonly kvStore: KVStore,
+    protected readonly defaultChains: Chain[]
+  ) {
+    this.kvStore = kvStore;
```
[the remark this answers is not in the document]
jurevans: Correct. Thanks!

apps/extension/src/provider/InjectedAnoma.ts
```diff
@@ -0,0 +1,177 @@
...
+import { Anoma } from "./Anoma";
+import { Result } from "../router/types";
+
+type ProxyMethods =
+  | "suggestChain"
+  | "connect"
+  | "getSigner"
+  | "chain"
+  | "chains"
+  | "version";
```
memasdeligeorgakis, line 12: nit:
```suggestion
enum ProxyMethod {
  SuggestChain = "suggestChain",
  Connect = "connect"
}
```
I feel that while needing a few more keystrokes at use site, this makes a way better refactoring experience as one can much more confidently do a global search 

`ProxyMethod.Connect` vs. `connect`

jurevans: I'll take a look. These values map 1-to-1 with the naming used in the `Anoma` and `InjectedAnoma` methods, so I wonder if there's a better mechanism to confirm the passed method arg matches this (though I'm not opposed to wrapping them in an enum, but it would be nice to not have to worry about it while refactoring at all if possible if they could be typed more automatically). If I don't find a better way, I'll move them into an enum

jurevans: As the methods we want to proxy belong to the `Anoma` class, I've simply done this:
```ts
type ProxyMethod = keyof Anoma;
```
Which gives us the type-safety we need, without need for future maintenance.

apps/extension/src/router/Message.ts
```diff
@@ -0,0 +1,35 @@
+import { Env, MessageSender } from "./types";
+
+/**
+ * This messaging system is influenced by cosmos-sdk.
+ * The messages are processed in the following order:
+ * "deserialize message -> approve external -> validate basic -> handler by routing".
+ * This deserializing system has weak polymorphism feature.
+ * Message would be converted to object according to their class and registered type.
+ * But, nested class is not supported. Non primitivie types or array that includes non primitive types in message's fields
+ * can't be decoded to their type properly. In this case, user should set thier prototype manually.
```
memasdeligeorgakis, line 10:
```suggestion
 * can't be decoded to their type properly. In this case, user should set their prototype manually.
```

jurevans: Thanks! I actually rewrote the whole thing to make better sense. Some of the comments have been inherited, so it's good that we update them to our standards and according to our specifications!

apps/extension/src/provider/InjectedAnoma.ts
jurevans: Even better, I've added:
```ts
type ProxyMethod = keyof Omit<Anoma, "version">;
```
As we only want to type-check methods that can actually get proxied. During runtime, this has no effect, but makes it clear to developers _which_ methods should be allowed to pass through.

apps/extension/src/provider/InjectedAnoma.ts
```diff
@@ -0,0 +1,183 @@
...
+        if (result.error) {
+          // TODO: Remove following `any` once structure is known
+          reject(new Error(result.error as any));
+          return;
+        }
+
+        resolve(result.return);
+      };
+
+      this.eventListener.addMessageListener(receiveResponse);
```
memasdeligeorgakis, line 132: This is what I meant, so I was wondering why we keep on adding `receiveResponse` here and then removing it inside itself once it will be called in return?

memasdeligeorgakis:
When walking through I made some notes and just putting them here, so maybe you can correct if my understanding seems incorrect.

**Files loaded by the extension**

* **background.anoma.js** <- **apps/extension/src/background/index.ts** - This contains the core of the extension application.
* **content.anoma.js** <- **apps/extension/src/content/index.ts** - this file injects the "gateway" script to the host page such as (namada.me). It will be loaded when the site is accessed and it will bootstrap the extension application and inject `injected.anoma.js` to the page, this will cause `injected.anoma.js` to load.
* **injected.anoma.js** <- **apps/extension/src/content/injected.ts** - This is being injected to the host web page by `content.anoma.js`. It is importing and initializing `InjectedAnoma`. `InjectedAnoma` contains all those functions that are visible at `windows.anoma`, and this dispatches the messages that are being captured by the core code in `apps/extension/src/background/index.ts`.
* **popup.anoma.js** <- **apps/extension/src/popup/index.ts** - This contains the visible UI of the extension.

packages/types/src/anoma.ts
```diff
@@ -0,0 +1,31 @@
...
+  address: string;
+  publicKey: Uint8Array;
+};
+
+export interface Signer {
+  sign(account: Account, tx: Transaction): Promise<Transaction>;
+  accounts: Account[];
+}
+
+export interface Anoma {
```
memasdeligeorgakis, line 19: I feel that thing name could benefit of being more specific. It is describing the public API of the extension.

apps/extension/src/provider/Anoma.ts
```diff
@@ -0,0 +1,53 @@
...
+import { Anoma as IAnoma, Signer } from "@anoma/types";
+import { Ports, MessageRequester } from "../router";
+import {
+  GetChainMsg,
+  GetChainsMsg,
+  GetSignerMsg,
+  SuggestChainMsg,
+} from "../background/chains";
+
+export class Anoma implements IAnoma {
```
memasdeligeorgakis, line 11: I got the feeling this this might be better if it lived next to all the code the relates to the "core" part of the extension, `/src/background`. It could still be "provided" from here thought, if we want to have consumers get it through a provider.

apps/extension/src/provider/InjectedAnoma.ts
```diff
@@ -0,0 +1,177 @@
...
+  args: any;
+}
+
+export interface ProxyRequestResponse {
+  type: ProxyRequestTypes.Response;
+  id: string;
+  result: Result | undefined;
+}
+
+export class InjectedAnoma implements IAnoma {
```
memasdeligeorgakis, line 32: Same as with `class Anoma implements IAnoma` I feel this might be better living next to the code that is only specific to the host page related class.

apps/extension/src/provider/InjectedAnoma.ts
jurevans: I agree this is a little confusing (and could stand to be refactored), but essentially, we are registering a listener in `requestMethod` to validate the message we are sending back to the client. This is listening for a response from the content-scripts (which is where our API instance exists), and validating it before sending it back to the client (it must be the correct type, have the correct ID, and if any errors were encountered in the background, reject the promise and pass it back to the client).

The `proxyMessage` is constructed with the generated ID, and a handler (`receiveResponse`) is created that knows that ID (as it's in scope when this handler is defined) and can match it, and so we want to listen with that handler just before we send the response (`event.postMessage(proxyMessage)`) This is essentially adding a listener for a specific, expected message, and then immediately removing itself ([Line 117](https://github.com/anoma/namada-interface/blob/80f0e28cf1f34531b225a62bb9c9db22e2bf55a7/apps/extension/src/provider/InjectedAnoma.ts#L117). Any time we send a `ProxyMessage` to the content-scripts, we register a listener to respond to that exact message. Any message that is proxied from the client to the API gets a one-time listener for that message.

1. [L89-L93](https://github.com/anoma/namada-interface/blob/80f0e28cf1f34531b225a62bb9c9db22e2bf55a7/apps/extension/src/provider/InjectedAnoma.ts#L89-L93) - An ID is generated for the requested method to proxy
2. [L104-L133](https://github.com/anoma/namada-interface/blob/80f0e28cf1f34531b225a62bb9c9db22e2bf55a7/apps/extension/src/provider/InjectedAnoma.ts#L104-L133) - We define the listener here, which if it is able to validate that the ID of the requested method matches, and no errors have been returned, will resolve the promise and return the results to the client.

apps/extension/src/router/types/router.ts
```diff
@@ -0,0 +1,44 @@
...
+  browser.runtime.MessageSender,
+  "id" | "url" | "tab"
+>;
+
+export interface Env {
+  readonly isInternalMsg: boolean;
+  readonly requestInteraction: any;
+}
+
+export type FnRequestInteraction = <M extends Message<unknown>>(
```
memasdeligeorgakis, line 13: Is this like a dynamic `Message`? It seems as if it is not used outside of this file for now.

apps/extension/src/extension/utils.ts
```diff
@@ -0,0 +1,31 @@
...
+};
+
+// Determine if content-scripts can be executed in this environment
+export class ContentScriptEnv {
+  static readonly produceEnv = (sender: MessageSender): Env => {
+    const isInternalMsg = sender.id === browser.runtime.id;
+
+    return {
+      isInternalMsg,
+      requestInteraction: () => {
```
memasdeligeorgakis, line 24: What is `requestInteraction` used for? I see that we use Utility types in several occasions to remove this from `Env`. But I do not really see what what we use it for 🤔

jurevans: That's going to be important later when we're requesting connections to the extension from the interface.

apps/extension/src/router/types/router.ts
```diff
@@ -0,0 +1,44 @@
+import { Message } from "../Message";
+
+export type MessageSender = Pick<
+  browser.runtime.MessageSender,
+  "id" | "url" | "tab"
+>;
+
+export interface Env {
```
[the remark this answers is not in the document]
jurevans: I can add a comment. `Env` is a common term for environment, so I'd like to leave it as is, but it could use some context as to what it's used for.

apps/extension/src/router/types/router.ts
jurevans: It's not currently used, but it will be implemented when we are requesting to connect to the extension, and this will need to be in place.

apps/extension/src/router/types/router.ts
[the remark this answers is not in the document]
memasdeligeorgakis: Yeah the word I got, but reading the property names one might get the feeling that this type relates perhaps to a message or request. So I felt that the purpose of this type or what environment we mean here is not very clear.

jurevans:
> Just putting this here for the future reference. This is the current directory structure and I tried to summarise here what I think they do.

I agree! It would be good to better organize this so that related items reside in a common top-level directory. I think this will take some additional work, and a good amount of thought, to find something that will work as it grows. We should keep in mind the next round of features related to integrating this with `namada-interface`. Some of the types you pointed out above will be important to be in place for then, and hooking up the rest of the features may have implications for the overall organization as well.

Since this PR has been open for a bit, there is quite a lot of work based on it (See #71). I would propose that following that PR, I should go in and implement some of these changes, once we have more pieces in place. These changes are easy to address, but will be somewhat of a challenge/headache to merge into #71 when the time comes. Also, we should perhaps come to a consensus on what the structure should be. I definitely want to put more thought into this before implementing a change across the project. Also, I will be putting together some overview graphics to show how the pieces relate to each other, and this may help guide us toward a better structure.

-----------------------------------

### Some additional thoughts:

I see this application containing 3 primary components:

1. The **background** services (we can think of this as a back-end API that the extension pages/scripts can interact with) running in the service worker/background scripts - this can access storage
2. The **injected** API (This is the `InjectedAnoma` script providing the proxy to the API), which exposes public methods to the client
3. The **extension** - What I call the "extension" is everything else that can access the `WebExtension` API (outside of the background services), that has access to the browser runtime within the extension, and this has a few components:
  - The API instance in **content-scripts** (responsible for injecting the proxied methods into a webpage for clients to connect to)
  - The Router responsible for relaying and validating messages between content-scripts and background
  - The React application delivered via `popup.html`, which provides access to the background services, but this could be any number of pages that exist within the extension, e.g., a post-installation setup page

Another way we could look at is solely by which browser/window APIs each has access to (as they'll need to be targeting these environments correctly):

| Extension Component | Global API |
| :-- | :-: |
| **background**  | `browser` |
| **injected**        | `window` |
| **extension**     | `browser` + `window` |
 
We should also keep in mind the build pipelines as we organize this as well, perhaps organizing around the targets to make it easy to reason about when something needs to be changed. This may make it more clear where dependencies can be shared. It's important that we're not importing unnecessary or incompatible dependencies within this project, or attempting to access an invalid API (for example, in the `page` and `content-scripts` runtime, you have both the `browser` and a limited version of the `window` APIs, but in the `background` you have no `window`, but full access to `browser` - sharing a dependency that targets `window` will crash the service worker, etc.). In this PR I've made sure that this doesn't happen, but looking at the top-level structure, it isn't clear which domain each top-level folder belongs to. 

Just for reference, these are the current Webpack targets (as of this PR) and their responsibilities - As this is ultimately what gets delivered, it may be a good idea that these are accessible from the top of the hierarchy:

- **background**
  - Service Worker (Chrome & other v3 manifests), Background Script (Firefox & other v2 manifests)
  - Responsible for key management, authentication, and interaction with the storage API
  - Is able to receive messages from an extension `page` (such as popup) or `content-scripts` via the router
  - This also has access to our wasm functionality (serializing/signing transactions, key derivation, etc.)
- **content**
  - Responsible for injecting the proxied API instance (`InjectedAnoma`)
  - Responsible for routing messages from the web page to 
- **injected**
  - The script that executes on the client's web page
- **popup**
  - The React app for managing keys
- **manifest.json**

Other projects tend to take it a step further with modularization, and many components of their extensions are actually abstracted into individual packages, perhaps for re-use in different contexts. I don't think that's important to us until we come across that use-case, but it is something to note. Just as an example, the [keplr-wallet extension](https://github.com/chainapsis/keplr-wallet/tree/master/packages) has separate packages within their mono-repo, including (but not limited to) the following:

```
packages/
  background/
  content/
  extension/
  provider/
  router/
  router-extension/
```

I've followed a somewhat similar hierarchy in parts. I like having individual services listed under `background`, with each following a similar structure from one to the next (each service should contain similarly named components within them, making it clear what their purpose is)

memasdeligeorgakis:
> Since this PR has been open for a bit, there is quite a lot of work based on it (See https://github.com/anoma/namada-interface/pull/71). I would propose that following that PR, I should go in and implement some of these changes, once we have more pieces in place.

Yeah, let's just do them after that #71. I created this https://github.com/anoma/namada-interface/issues/85 task where we can collect some thoughts for the upcoming work on this.

memasdeligeorgakis (approved):
Left this task to follow up on this https://github.com/anoma/namada-interface/issues/85.

Issue #66, opened 2022-09-16
Staking, Governance, Public Goods Funding - unified the fake data fetching in Staking


memasdeligeorgakis:
* Get all fake data through actions
* Clean up the types

jurevans:
Closing as this is done

PR #67, opened 2022-09-16
#66 Unifying data types and fake data in staking

memasdeligeorgakis:
# What does this PR do

* cleaned up staking-related types
  * It nests `Validator` in staking data 
  * cleans up the naming
* changed all fake data to go through actions
  * removes the fake data from the component's directory
  * all fake data runs through actions and redux
  

# FAQ
**Q:** Why is the data drilled down through a few layers?
**A:** While it is quite many, I feel that this view is very likely to change, or even be completely reshuffled for desktop/mobile. So this way maintaining several views for this data will be quick to refactor. So I feel that doing any changes in these stateless views will be very cheap.

image

image


memasdeligeorgakis:
I did not add tests for the actions as they are just placeholders, likely when we fetch real data we have to add some logic to the actions.

apps/namada-interface/src/App/Staking/StakingOverview/StakingOverview.tsx
```diff
@@ -154,14 +154,15 @@ export const StakingOverview = (props: Props): JSX.Element => {
...
-        data={myBalancesData}
+        data={myBalances}
         tableConfigurations={myBalancesConfigurations}
       />
 
       {/* my validators */}
       <Table
         title="My Validators"
-        data={myValidatorData}
+        // data={myValidatorData}
```
jurevans, line 164: Definitely not a big deal, but this comment could be removed.

jurevans (approved):
Looks good! I only left a couple very minor comments, just cleaning up old comments

Issue #60, opened 2022-09-15
Extension: Account creation


jurevans:
User should be able to:

- Generate a mnemonic
- Verify the mnemonic
- Create a password
- Generate an initial account (derive account from master seed)

See [Polkadot-js extension](https://polkadot.js.org/extension/) - This is the UX we're going for.

**NOTE** - Polkadot-js has the following additional features that we will likely want to incorporate as well:

- Create new account (this generates a new mnemonic, sets a new password)
- Account addresses can be used on any chain, unless otherwise specified (if specified, the encoding is set for the target chain)
- Derive from account
  - Choose existing account
  - Enter password for said account
  - Derivation path (defaults to `//1`) -> Unlock to edit

PR #71, opened 2022-09-19
feat/60 - Extension: Account Creation

jurevans:
See https://github.com/anoma/namada-interface/issues/60

**NOTE** Our legacy Rust library (`@anoma/wasm`) which is being used on `namada-interface`, now fails to build. Since this will be retired soon with `@anoma/shared` supporting the latest `namada` release, I've added a properly built artifact so it will continue to work in PR environments and `namada.me`. 

This PR is to implement the account creation flow.

- [x] User can create and validate a mnemonic
- [x] User can set and verify a password
- [x] User can derive accounts from the seed
  - [x] A default initial account derivation path is derived from mnemonic
  - [x] Provide a default derivation path
  - [x] User can specify a custom derivation path (so long as it hasn't already been used) See the UI for <https://github.com/polkadot-js/extension> and Keplr-Wallet (+Add Account -> This PR is following that UI)
- [x] Stored mnemonic seed and private keys are encrypted

## Testing

As authentication is not currently implemented (this is being done in https://github.com/anoma/namada-interface/pull/84), the password set during the Setup process will only be valid unless the Service Worker/Background Script is refreshed, therefore, if you are using the dev tools (`yarn start:chrome`, etc.), you won't be able to add an account if, for example, you make a change in any background script causing it to be refreshed, and you'll get an error (which is expected if you attempt an action requiring a password and one isn't set).

It may be easiest to test via `yarn build:chrome` and `yarn build:firefox`, as nothing will trigger a refresh. In https://github.com/anoma/namada-interface/pull/84, if the user is not authenticated, they will see a login form to unlock the wallet. You can still test with `yarn start:chrome` and `yarn start:firefox`, just keep this caveat in mind if you intend to poke around in the background scripts while the dev-server is running.

## Notes

- To reuse the components and theme utilities from `namada-interface`, I began porting the following:
  - `namada-interface/src/components` -> `@anoma/components`
  - `namada-interface/src/utils/theme` -> `@anoma/utils/theme`
- As per the `keplr-wallet` wallet workflow, the user can launch the Create Account process in a new tab, and once an initial account is created, the user can view all accounts in the extension popup. It would be nice to have this launch on install, which is doable.
  - This adds another entry point to `webpack.config.js`, `setup/index.ts`, which is a React app containing the setup process (AccountCreation flow). This is nearly identical to what exists on `namada-interface`, just cleaned up a bit, and with the addition of user feedback during the completion process.
  - The popup retains its original entry point in the config, and is a separate React app containing the derived accounts overview (and eventually settings and editing features). If no accounts exist in the popup, the user is directed to the setup page. The account creation process cannot happen here (like in Keplr), as the user needs to be able to navigate away to handle storing their mnemonic for safe-keeping, which in the case of a popup, destroys the state forcing them to start over. 
- We aren't too concerned about design in the extension (it's much more important that it works correctly and can be integrated with `namada-interface`), so long as functionality can be demonstrated
- The accounts listing in the popup would do well to follow the Polkadot-js extension. Unlike Keplr, for now, we're onlyconcerned with Namada addresses here. The user should be able to derive additional accounts with either **a)** a provided derivation path (incremented from previous path), or **b)** a custom path (which must be validated)
  - One thing to note about Polkadot's extension, in contrast to Keplr: You can generate a new mnemonic/account in Polkadot within the popup. Keplr allows this within their Account setup page, but not inline with the accounts overview.  Keplr requires a password to access the popup. The password behaves differently in Polkadot, where it is required when you want to perform any actions on an account, but it isn't required to _see_ the accounts in the popup, and a password is set for each top-level account. We might consider which of these ways suits our needs best.
  - I think an amalgamation of the two might be best
    - Like Polkadot, the user should see both `Create new account` and `Derive from account`
    - `Create new account` should launch the setup process for account creation, making it possible to do mnemonic verification like in Keplr.
- For the purposes of this PR:
    - the user will be able to generate a mnemonic and set a password once, as we have it on `namada-interface`, and then can derive additional accounts from that seed. When requirements are clear, given the above options, we can implement a more comprehensive flow. That said, the background scripts of this extension _do_ support storage for multiple mnemonics, it's a matter of exposing that to the user.
  - Key derivation essentially follows what is done in the `namada wallet` CLI - ed25519 key pairs are derived from a seed, and an `ImplicitAddress` type is displayed to the end user (along with alias and derivation path). I will add the feature to initialize this account to receive an `EstablishedAddress` (we already have this functionality in `namada-interface`) in a follow-up PR, as this one is already _huge_.
  - For now, I'm hard-coding `0'` as the coin-type, but ultimately this will be a value in https://github.com/satoshilabs/slips/blob/master/slip-0044.md (whatever the Slip44 for NAM is).
  - See:
     - https://github.com/bitcoin/bips/blob/master/bip-0032.mediawiki (BIP32 HD Wallets)
     - https://github.com/bitcoin/bips/blob/master/bip-0044.mediawiki (BIP44 paths)
     - https://github.com/satoshilabs/slips/blob/master/slip-0044.md (Slip-0044 Coin Types)


apps/extension/src/setup/index.ts
```diff
@@ -0,0 +1,3 @@
+import Setup from "../Setup";
+
+export default Setup();
```
memasdeligeorgakis, line 3: I got problem here as we have 2 directories that only differ in casing
```bash
/src/setup
/src/Setup
```
also I have compiler warning in `/src/setup/index.ts` that I expect to be due to this: `'default' implicitly has type 'any' because it does not have a type annotation and is referenced directly or indirectly in its own initializer.ts(7022)`

I see all the files in those 2 directories listed in one folder in VSCode, Sublime Text and in Terminal. So I expect this to be system wide on MacOS. I think it would be better to place all files truly in the same directory or have name that do not only differ in casing.

jurevans: I removed `src/setup`, and the build directly links to `src/Setup`. I did the same for the `popup` build pipeline, which was importing a function that renders the `src/App` React app, but this is not needed either.

apps/extension/src/background/keyring/keyring.ts
```diff
@@ -1,68 +1,217 @@
...
-/*
- Keyring stores keys in persistent backround.
+/**
+ * Keyring stores keys in persisted backround.
  */
 export class KeyRing {
-  private _loaded: boolean;
+  private _mnemonicStore: IStore<MnemonicState>;
+  private _accountStore: IStore<AccountState>;
+  private _password: string | undefined;
```
memasdeligeorgakis, line 39: Nit: I assume that the underscores are in all of these just for the consistency, while we actually only need it here for `status` which has a getter? I see this also in other places. Best we decide a consistent rule.

apps/extension/src/background/keyring/types.ts
```diff
@@ -0,0 +1,32 @@
...
+  id: string;
+  alias?: string;
+  phrase: Uint8Array;
+};
+
+export type DerivedAccount = {
+  id: string;
+  address: string;
+  alias?: string;
+  bip44Path: Bip44Path;
```
memasdeligeorgakis, line 17: nit: listing optionals last
```suggestion
  bip44Path: Bip44Path;
  alias?: string;
```

apps/extension/src/background/keyring/types.ts
```diff
@@ -0,0 +1,32 @@
...
+  establishedAddress?: string;
```
memasdeligeorgakis, line 18: nit: listing optionals last
```suggestion
  id: string;
  address: string;
  bip44Path: Bip44Path;
  alias?: string;
  establishedAddress?: string;
```

apps/extension/src/background/keyring/types.ts
```diff
@@ -0,0 +1,32 @@
+export type Bip44Path = {
+  account: number;
+  change: number;
+  index: number;
+};
+
+export type MnemonicState = {
+  id: string;
+  alias?: string;
+  phrase: Uint8Array;
```
memasdeligeorgakis, line 10: nit: listing optionals last
```suggestion
  id: string;
  phrase: Uint8Array;
  alias?: string;
```

apps/extension/src/App/Accounts/AddAccount.tsx
```diff
@@ -0,0 +1,193 @@
...
+    const { index } = account.bip44Path;
+    if (index > maxIndex) {
+      maxIndex = index;
+    }
+  });
+
+  return maxIndex + 1;
+};
+
+enum Status {
```
memasdeligeorgakis, line 64: nit: I feel that if we do state management in React components, we might like to do some small util/share something like this type. We only now have this 3 times, but likely will have it several times soon.

apps/extension/src/App/Accounts/AddAccount.tsx
```diff
@@ -0,0 +1,193 @@
...
+const AddAccount: React.FC<Props> = ({ accounts, requester, setAccounts }) => {
+  const navigate = useNavigate();
+  const [alias, setAlias] = useState("");
+  const [account, setAccount] = useState(0);
+  const [change, setChange] = useState(0);
+  const [index, setIndex] = useState(findNextIndex(accounts));
+  const [bip44Error, setBip44Error] = useState("");
+  const [isFormValid, setIsFormValid] = useState(true);
+  const [formError, setFormError] = useState("");
+  const [formStatus, setFormStatus] = useState<Status>();
```
memasdeligeorgakis, line 78: nit: I am wondering if we should have also Idle. As now we have 4 states, `undefined`, `Pending`, `Failed`. We have no `Completed` but this component disappears as soon as it it success?

apps/extension/src/background/keyring/types.ts
jurevans: See my comment at https://github.com/anoma/namada-interface/pull/71/#discussion_r988716978 - I think there are different ways of looking at this.

apps/extension/src/App/Accounts/AddAccount.tsx
jurevans: I agree with this, but would first like to refer to the subsequent PRs that are out, as some of this has been moved around or removed. When using the `ExtensionRequester` `sendMessage` function, we should have one set of states that it could be in, so I definitely agree these should be consolidated.

That said, I don't think there is a good use case for doing state-management outside of what React provides, as it these are still very simple. While it _may_ grow somewhat in the future, I still think it will be very manageable with only using `react-router` and local state (or context for any globals)

apps/extension/src/background/keyring/keyring.ts
jurevans: The `_` prefix is for any private variables, which I have been doing consistently. I see different approaches across other extensions and wallet apps, and it just comes down to what we want to do for consistency. Some are very strict about prefixing `_` to _any_ private variable (See polkadot-js), whereas I rarely see that done in Keplr. I think scoping certain things as private and indicating it as such is important, but that's just an opinion. I think it's better for the developer (or anyone new to the code) to know that a method or value is only accessible from inside a class, and it's a very common convention. I also think it's good form to scope these things with intention, only allowing public access to public values, etc., otherwise it just gets sloppy and hard to reason about. I like the Rust approach, where everything is private by default, and you must be explicit about making something public (which is the opposite of TS/JS, where everything is public by default), but that's an opinion, and we should determine what works best for us!

apps/extension/src/background/keyring/messages.ts
```diff
@@ -75,3 +80,120 @@ export class CheckPasswordMsg extends Message<boolean> {
     return CheckPasswordMsg.type();
   }
 }
+
+export class GenerateMnemonicMsg extends Message<string[]> {
+  public static type() {
+    return MessageTypes.GenerateMnemonic;
+  }
```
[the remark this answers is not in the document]
jurevans: I think this is a better approach! I was thinking something similar. This portion was inherited, with the exception of adding an enum with all possible values, though I don't think there is any use in `type()` being a function, and switching it to a field would be much simpler and cleaner (but this will need to be checked thoroughly to make sure all instances are caught)

packages/components/src/Input/types.ts
```diff
@@ -0,0 +1,6 @@
+export enum InputVariants {
```
memasdeligeorgakis, line 1: nit:
```suggestion
export enum InputVariant {
```

apps/extension/src/background/keyring/keyring.ts
```diff
@@ -1,68 +1,217 @@
...
+
+  // Concatenate any number of args onto the name parameter
+  args.forEach((arg) => {
+    uuidName = `${uuidName}::${arg}`;
+  });
+
+  return uuid(uuidName, UUID_NAMESPACE);
+};
+
+enum KeyRingStore {
```
memasdeligeorgakis, line 28: nit: would it be more correct to call these `<name>StoreKey` as they are not the store, but a key to a store. The method they are passed also names the parameter `key`.

```suggestion
enum KeyRingStoreKey {
```

apps/extension/src/background/keyring/types.ts
```diff
@@ -0,0 +1,32 @@
...
+  parentId: string;
+  private: Uint8Array;
+  public: Uint8Array;
+};
+
+export enum KeyRingStatus {
+  NOTLOADED,
+  EMPTY,
+  LOCKED,
+  UNLOCKED,
```
memasdeligeorgakis, line 31: nit: all caps feels inconsistent against the other other enums
```suggestion
  NotLoaded,
  Empty,
  Locked,
  Unlocked,
```

apps/extension/src/background/keyring/keyring.ts
```diff
@@ -1,68 +1,217 @@
...
-    this._loaded = false;
-    this._keyStore = null;
+  constructor(kvStore: KVStore) {
+    this._mnemonicStore = new Store(KeyRingStore.Mnemonic, kvStore);
+    this._accountStore = new Store(KeyRingStore.Account, kvStore);
   }
 
   public isLocked(): boolean {
-    return this.mnemonic == null || this._password == null;
+    return !this._password;
```
memasdeligeorgakis, line 48: Likely we would never allow a password which is an empty string`""`, but that is also falsy. It feels as if `this._password === undefined` would be more correct here (and on line 136) as we indicate the lack of a password with `undefined`?

apps/extension/src/background/keyring/messages.ts
```diff
@@ -1,11 +1,16 @@
-import { Message } from "../../router";
+import { PhraseSize } from "@anoma/crypto";
+import { Message } from "router";
 import { ROUTE } from "./constants";
-import { KeyRingStatus } from "./keyring";
+import { Bip44Path, DerivedAccount, KeyRingStatus } from "./types";
 
 enum MessageTypes {
```
memasdeligeorgakis, line 6: nit: Not part of this PR but maybe better to keep enums in singular naming

```suggestion
enum MessageType {
```

apps/extension/src/background/keyring/messages.ts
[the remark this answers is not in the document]
jurevans: I started this issue to cover refactoring the `Message` type: #108 - I took a quick look at it while linting, and it will take a bit more work to consolidate all messages into a type and move `type()` to a field.

memasdeligeorgakis (approved):

apps/extension/src/background/keyring/keyring.ts
jurevans: @memasdeligeorgakis This enum has been removed entirely. In https://github.com/anoma/namada-interface/pull/97 (storage updates), everything resides under a single store, so there's no need to differentiate. Instead, there's a single constant defining the store's key (see https://github.com/anoma/namada-interface/blob/1bcb3230792fb0d72f27e962f462b6272668783a/apps/extension/src/background/keyring/keyring.ts#L30-L40).

Issue #61, opened 2022-09-15
Extension: Authentication


jurevans:
User should be able to:

- Lock (or "log out of") the extension
- Unlock (or "log in to") the extension

Other criteria:

- Password will be checked against encrypted mnemonic in storage to validate, at which point it (the password) is persisted in the KeyRing state (similar to Keplr). In the case of Chrome, any time the service worker goes inactive (after, say, 5 minutes), the session is expired and the user would need to again authenticate. The UI should reflect this, querying the state of the keyring to determine where to direct user when the extension is opened. In the case of Chrome, it should be considered a bug if the session expires (the Service Worker becomes inactive) before we determine it should (via session timeout on inactivity) - and this has been logged in #92)
- `namada-interface` (or any connected client) should not be able to access any protected items in storage if the keyring is locked. An attempt to connect to the wallet while locked should show the user the login screen (This will be handled in #95)

PR #84, opened 2022-09-26
feat/61 - Extension: Authentication

jurevans:
See issue https://github.com/anoma/namada-interface/issues/61

The PR is to implement password verification, locking and unlocking of extension, and approving requests to connect to extension.

- [x] User can lock the extension
- [x] User can authenticate using their password

_Note_ I also improved naming conventions used in `@anoma/crypto`, match what is generally expected (i.e., for mnemonics we want to show the protocol in the module along with the exported struct, e.g., `bip39::Mnemonic`. Another example is `bip32::HDWallet`, where `HDWallet` utilizes BIP32 to derive accounts using BIP44 derivation paths, etc.)


apps/extension/src/App/Accounts/AddAccount.tsx
```diff
@@ -139,21 +139,24 @@ const AddAccount: React.FC<Props> = ({ accounts, requester, setAccounts }) => {
                 {[bip44Prefix, coinType].join("/")}/
               </Bip44PathDelimiter>
               <Bip44Input
-                type="text"
+                type="number"
```
jurevans, line 142: Note that this form has changed a quite a bit in #97, to match closely match the form on polkadot-js extensino: https://github.com/anoma/namada-interface/blob/1bcb3230792fb0d72f27e962f462b6272668783a/apps/extension/src/App/Accounts/AddAccount.tsx#L142-L167 Until we have a more specific requirement, it makes sense to do it this way as it enforces _valid_ derivation paths under the mnemonic seed.

jurevans: Specifically, when deriving an address from an account on polkadot, the user can specify `change` and `index`. `change` can only be a 0 (external address) or 1 (internal), and `account` can be any unused value (though it's preferable to be sequential) See https://github.com/bitcoin/bips/blob/master/bip-0044.mediawiki#Change

apps/extension/src/App/Accounts/AddAccount.tsx
```diff
@@ -112,8 +112,8 @@ const AddAccount: React.FC<Props> = ({ accounts, requester, setAccounts }) => {
     e: React.ChangeEvent<HTMLInputElement>,
     callback: (value: number) => void
   ): void => {
-    const result = e.target.value.replace(/\D/g, "") || "0";
-    callback(parseInt(result));
+    const result = e.target.value || "0";
+    callback(parseInt(result, 10));
```
memasdeligeorgakis, line 116: nit: what if this fails? should we handle it in a graceful way?

jurevans: I'm not sure it could fail - the input comes from `<input type="number" />` which should only allow numerical input. The value is returned as a string, though, so it needs to be parsed, but I'm not sure it's possible to trigger a failure in this case.

apps/extension/src/App/App.tsx
```diff
@@ -1,57 +1,88 @@
...
+export enum Status {
   Completed,
   Pending,
   Failed,
 }
 
 export const App: React.FC = () => {
   const theme = getTheme(true, false);
+  const navigate = useNavigate();
   const [status, setStatus] = useState<Status>();
```
memasdeligeorgakis, line 39: nit: This went in already earlier but now that it is being used outside of this file, I feel now that it is being used outside of this file, this name is even more ambiguous as one cannot easily read the code to get the whole context of "what status this is".

```suggestion
  const [accountsLoadingStatus, setAccountsLoadingStatus] = useState<Status>();
```

apps/extension/src/App/App.tsx
```diff
@@ -1,57 +1,88 @@
...
+    } catch (e) {
+      console.error(e);
+      setError(`An error occurred while loading extension: ${e}`);
+      setStatus(Status.Failed);
+    } finally {
+      setStatus(Status.Completed);
+    }
+  };
+
+  const checkIsLocked = async (): Promise<void> => {
```
memasdeligeorgakis, line 60: nit: the naming makes it feel as if it checks if something is locked, while it also navigates.

apps/extension/src/App/LockWrapper/LockWrapper.tsx
```diff
@@ -0,0 +1,44 @@
...
+import { ButtonContainer, LockExtensionButton } from "./LockWrapper.components";
+import { LockKeyRingMsg } from "background/keyring";
+import { Ports } from "router";
+import { Status } from "App/App";
+import { TopLevelRoute } from "App/types";
+import { ExtensionRequester } from "extension";
+
+type Props = {
+  requester: ExtensionRequester;
+  setStatus: (status?: Status) => void;
```
memasdeligeorgakis, line 14: nit: left a comment about the ambiguous naming of this 
>nit: This went in already earlier but now that it is being used outside of this file, I feel now that it is being used outside of this file, this name is even more ambiguous as one cannot easily read the code to get the whole context of "what status this is".

https://github.com/anoma/namada-interface/pull/84/files#r992316411

apps/extension/src/App/App.tsx
jurevans: Since these are pretty quick changes, I'll add them in my current PR (https://github.com/anoma/namada-interface/pull/114) to avoid potential conflicts.

Issue #86, opened 2022-09-26
Staking, Governance, Public Goods Funding - Staking/Unstaking Views


memasdeligeorgakis:
# Description

This should implement the view components for staking and unstaking. These should be displayed inside the `Modal` component. The state management should go the same way as any other state management under Staking parts of the UI.

PR #87, opened 2022-09-26
#86 staking and unstaking

memasdeligeorgakis:
# What does this PR do
It adds staking and unstaking views and the state management for them. The intention is to be very plain. It contains a naive field validation that should be replaced with a central one in the utils.

image

*initial view, confirm disabled*

image

*valid staking amount*

image

*valid unstaking amount*

image

*invalid amount*

TODO:
* likely the modal should be hugging the content while having a max height. Now it takes a constant height even if the content would not require so much.


apps/namada-interface/src/App/Staking/Staking.tsx
```diff
@@ -68,9 +99,11 @@ export const Staking = (props: Props): JSX.Element => {
   } = props;
 
   // these 2 are needed for validator details
-  const stakingPositionsWithSelectedValidator = myStakingPositions.filter(
-    (validator) => validator.validatorId === selectedValidatorId
-  );
+  const stakingPositionsWithSelectedValidator =
+    myStakingPositions &&
```
mateuszjasiuk, line 103: `myStakingPosition` is of type `MyStakingPosition[]`, so either `myStakingPosition` is always true or the type has to change. Kinda weird that typescript does not show it as "always true"

mateuszjasiuk (approved):
Looks good to me! \
Left one comment regarding type.

Also not connected to this PR directly, when I was looking at redux state:
`stakingOrUnstakingState` shows `0 | 1 | 2`. Maybe it would be nice to change enum to have string values, so it's more clear when looking at state :)

apps/namada-interface/src/App/Staking/Staking.tsx
memasdeligeorgakis: Odd that it was there, but now I removed it and TS seems to be happy about it as expected.

apps/namada-interface/src/slices/README.md
```diff
@@ -0,0 +1,17 @@
+# The state management of namada-interface
+
+## Description
+This directory contains the state management for the application. It is divided by main topics (staking and governance, transfers, ...). Each directory contains the following files:
+* **actions.ts** - this contains all the actions for the state. All the side effects (network calls, calls to libs, ...) are being performed from here.
+* **reducers.ts** - all reducers of the state are here. This is where the state changes are being committed. Commonly all actions have 3 reducers associated with them:
+    * pending - this is being triggered with the initial dispatch call (mostly originating from the UI or view Components). The payload to this is same as the action was dispatched with.
```
jurevans, line 7: Very minor suggestion: This should likely be `**pending**` to match the others

jurevans (approved):
Looks good! Just one minor comment in the slices README

Issue #94, opened 2022-09-29
Extension: Improve storage


jurevans:
We should be following other extensions to some degree when it comes to handling safely storing keys. 

We should take a look at Keplr Wallet, which makes use of AES (Counter Mode) with Scrypt KDF, and Polkadot, which utilizes Argon2 for password-based key derivation.

More details _TBD_

Issue #4, opened 2022-04-21
Generic Todos/Improvements


memasdeligeorgakis:
# Description
This is for collecting generic TODOs and improvement ideas in the codebase.

# Outcomes
* comments under this ticket listing the TODOs. When the work on them starts a new task should be created for each.

memasdeligeorgakis:
Move all accounts state management to the directory that is currently being called `AccountsNew`.

memasdeligeorgakis:
Think about the folder/component hierarchy and structure:

```bash
/App
  /StakingAndGovernance
  # 👆 only this is accessing the state management
  /Staking
    /StakingOverview
    /ValidatorDetails
    /NewStakingPosition
    # 👆 we need a type there that is re-imported from `/Apps/StakingAndGovernance` 
    # and defined in `/slices/StakingAndGovernance`
/slices
  /StakingAndGovernance
  # the state management stuff could well live next to the main consumer
  # to be more "contained"    
```

memasdeligeorgakis:
Create proper validation utils '(as part of `<Input />`, likely the best)

memasdeligeorgakis:
Thinking about the naming of the shared packages

```bash
/packages
  /components # ui-components
  /crypto # if this is purely some abstract utils, could it live under utils?
  /integrations # external-wallet-intergrations
  /masp-web # likely this could move to live under the directory containing other namada specific functionality
  /rpc # networking this is to abstract communication, not rpc protocol
  /seed-management
  /session # wallet-session
  /shared #namada-shared and do we need this independently are we using this directly from our ts code?
  /storage
  /tx # this would be covered by the one wrapping namadas core types?
  /types # would these not be shared from the directory containing the actual functionality for these types
  /utils
  /wallet # would this functionality live in location managing the keys 
  /wasm # namada-shared
```

memasdeligeorgakis:
Check out the generics in `KVStore` 
```ts
// namada-interface/packages/storage/src/store/types.ts

// from
export interface KVStore {
  get<T = unknown>(key: string): Promise<T | undefined>;
  set<T = unknown>(key: string, data: T | null): Promise<void>;
  prefix(): string;
}

// to aligned
export interface KVStore<T> {
  get(key: string): Promise<T | undefined>;
  set(key: string, data: T | null): Promise<void>;
  prefix(): string;
}
```
https://github.com/anoma/namada-interface/pull/97#issuecomment-1277650205

memasdeligeorgakis:
[...]

PR #97, opened 2022-09-29
WIP: feat/94 - Extension storage updates

jurevans:
See https://github.com/anoma/namada-interface/issues/94

### DIFF
--- a/apps/extension/src/background/keyring/types.ts
+++ b/apps/extension/src/background/keyring/types.ts
@@ -1,7 +1,19 @@
+import { ScryptParams, Argon2Params } from "@anoma/crypto";
+
 export type Bip44Path = {
   account: number;
   change: number;
   index: number;
+};
+
+enum KdfTypes {
+  Argon2 = "argon2",
+  Scrypt = "scrypt",
+}
+
+export type Storage = {
+  kdf: KdfTypes;
+  params: ScryptParams | Argon2Params;
 };
 
 export type MnemonicState = {
--- a/packages/crypto/lib/src/crypto/aes.rs
+++ b/packages/crypto/lib/src/crypto/aes.rs
@@ -1,0 +1,34 @@
+// TODO: Finish implementation
+// use aes::Aes256;
+use aes::cipher::{
+    // BlockCipher, BlockEncrypt, BlockDecrypt, KeyInit,
+    generic_array::GenericArray,
+};
+use wasm_bindgen::prelude::*;
+
+#[wasm_bindgen]
+pub struct AES {
+    key: Vec<u8>,
+}
+
+#[wasm_bindgen]
+impl AES {
+    #[wasm_bindgen(constructor)]
+    pub fn new() -> Self {
+        let key: &[u8]  = &GenericArray::from([0u8; 16]);
+
+        Self {
+            key: Vec::from(key),
+        }
+    }
+
+    pub fn key(&self) -> Vec<u8> {
+        self.key.clone()
+    }
+}
+
+impl Default for AES {
+    fn default() -> Self {
+        Self::new()
+    }
+}
--- a/packages/crypto/lib/src/crypto/argon2.rs
+++ b/packages/crypto/lib/src/crypto/argon2.rs
@@ -1,0 +1,176 @@
+use argon2::{
+    self,
+    password_hash::{
+        rand_core::OsRng,
+        PasswordHash, PasswordHasher, PasswordVerifier, SaltString
+    },
+    Params,
+};
+use serde::{Serialize, Deserialize};
+use gloo_utils::format::JsValueSerdeExt;
+use wasm_bindgen::prelude::*;
+
+#[wasm_bindgen]
+#[derive(Serialize, Deserialize)]
+pub struct Argon2Params {
+    m_cost: u32,
+    t_cost: u32,
+    p_cost: u32,
+}
+
+#[wasm_bindgen]
+impl Argon2Params {
+    #[wasm_bindgen(constructor)]
+    pub fn new(m_cost: u32, t_cost: u32, p_cost: u32) -> Self {
+        Self {
+            m_cost,
+            t_cost,
+            p_cost,
+        }
+    }
+}
+
+#[wasm_bindgen]
+pub struct Argon2 {
+    password: Vec<u8>,
+    params: argon2::Params,
+}
+
+/// Argon2 password hashing
+#[wasm_bindgen]
+impl Argon2 {
+    #[wasm_bindgen(constructor)]
+    pub fn new(
+        password: String,
+        m_cost: Option<u32>,
+        t_cost: Option<u32>,
+        p_cost: Option<u32>,
+    ) -> Result<Argon2, String> {
+        let password = Vec::from(password.as_bytes());
+        let default_params = Params::default();
+
+        let m_cost = match m_cost {
+            Some(m_cost) => m_cost,
+            None => default_params.m_cost(),
+        };
+
+        let t_cost = match t_cost {
+            Some(t_cost) => t_cost,
+            None => default_params.t_cost(),
+        };
+
+        let p_cost = match p_cost {
+            Some(p_cost) => p_cost,
+            None => default_params.p_cost(),
+        };
+
+        let params = Params::new(m_cost, t_cost, p_cost, None)
+            .map_err(|err| err.to_string())?;
+
+        Ok(Argon2 {
+            password,
+            params,
+        })
+    }
+
+    pub fn to_hash(&self) -> Result<String, String> {
+        let salt = SaltString::generate(&mut OsRng);
+        // Argon2 with default params (Argon2id v19)
+        let argon2 = argon2::Argon2::default();
+        let bytes: &[u8] = &self.password;
+        let params = &self.params;
+
+        // Hash password to PHC string ($argon2id$v=19$...)
+        let password_hash = argon2.hash_password_customized(
+            bytes,
+            None, // Default alg_id = Argon2id
+            None, // Default ver = v19
+            params.to_owned(),
+            &salt,
+        ).map_err(|err| err.to_string())?.to_string();
+
+        Ok(password_hash)
+    }
+
+    pub fn verify(&self, hash: String) -> Result<(), String> {
+        let argon2 = argon2::Argon2::default();
+        let bytes: &[u8] = &self.password;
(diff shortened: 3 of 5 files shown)


memasdeligeorgakis:
Not a part of this PR, but putting here for later as I thought it might be relevant as it is dealing with storage and this interface, we have that 

```ts
// namada-interface/packages/storage/src/store/types.ts
export interface KVStore {
  get<T = unknown>(key: string): Promise<T | undefined>;
  set<T = unknown>(key: string, data: T | null): Promise<void>;
  prefix(): string;
}
```
but I was wondering if we would not like to rather have it, to ensure we get and set the same type. 
```ts
export interface KVStore<T> {
  get(key: string): Promise<T | undefined>;
  set(key: string, data: T | null): Promise<void>;
  prefix(): string;
}
```
A good example of this is in `class ChainsService` we perform several calls like

```ts
this.kvStore.set<Chain[]>(KVKeys.Chains, savedChains);
this.kvStore.get<Chain[]>(KVKeys.Chains);
```

mateuszjasiuk (approved):
LGTM! Not claiming to understand everything :) Left two small comments.

memasdeligeorgakis:
Not a part of this PR, but putting here for later and also as this part of the storage. `T` should likely enforce `{id: string}` as it is very likely expected to be present when calling `update` and `remove`. saved for later [here](https://github.com/anoma/namada-interface/issues/4#issuecomment-1278604991).
```ts
// namada-interface/apps/extension/src/background/types/Store.ts
export interface IStore<T> {
  set(state: T[]): Promise<void>;
  get(): Promise<T[]>;
  append(state: T): Promise<void>;
  update(id: string, state: Partial<T>): Promise<void>;
  remove(id: string): Promise<void>;
}
```

apps/extension/src/App/Accounts/AddAccount.tsx
```diff
@@ -21,24 +25,23 @@ import {
...
-    newAccount.change,
-    newAccount.index,
-  ].join("/");
+  const newPath = [account, newAccount.change, newAccount.index].join("/");
   let isValid = true;
   accounts.forEach((a: DerivedAccount) => {
-    const { bip44Path } = a;
-    const { account, change, index } = bip44Path;
+    const { path } = a;
+    const { account, change, index } = path;
```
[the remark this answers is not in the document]
jurevans: That's correct, I definitely missed that :)

jurevans:
> Not a part of this PR, but putting here for later and also as this part of the storage. `T` should likely enforce `{id: string}` as it is very likely expected to be present when calling `update` and `remove`. saved for later

Good call! That is what I meant to do :) I updated my current PR with this change https://github.com/anoma/namada-interface/blob/07cf93e66d7c2f43a2bf5ead9544a5310e758e87/packages/storage/src/store/Store.ts#L3-L16 - That requirement is defined on the actual class `Store`, but I had forgot to add it to the interface.

jurevans:
> Not a part of this PR, but putting here for later as I thought it might be relevant as it is dealing with storage and this interface, we have that

I had actually spent some time on that when improving the types in https://github.com/anoma/namada-interface/pull/105 - I ran into some issues at that point, but it's something I plan on revisiting. Some of the types in `@anoma/storage` have been pulled out in #105. removing anything that doesn't get used. I'll take another look at it

jurevans:
> Not a part of this PR, but putting here for later as I thought it might be relevant as it is dealing with storage and this interface, we have that

I've updated this in my current PR, #114:
Interface:
https://github.com/anoma/namada-interface/blob/086d02d7debb6d1fd5d58d55b2a76f8b555ad86d/packages/storage/src/store/types.ts#L1-L5

Usage in Store:
https://github.com/anoma/namada-interface/blob/086d02d7debb6d1fd5d58d55b2a76f8b555ad86d/packages/storage/src/store/Store.ts#L30-L36

Usage in service:
https://github.com/anoma/namada-interface/blob/086d02d7debb6d1fd5d58d55b2a76f8b555ad86d/apps/extension/src/background/keyring/service.ts#L12-L14

apps/extension/src/App/App.tsx
```diff
@@ -83,6 +84,11 @@ export const App: React.FC = () => {
     }
   }, [status, accounts]);
 
+  const parent = accounts.find(
+    (account) => account.type === KeyStoreType.Mnemonic
+  );
+  const { account: parentAccount = 0 } = parent?.path || {};
```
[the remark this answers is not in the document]
jurevans: It is a lot of `?`, but in this case I think it's a lot more clear the way you propose! I've updated this. I think it's a good convention to use in these cases.

memasdeligeorgakis:
Not directly part of this PR, but `namada-interface/packages/storage/src/store/MemoryKVStoreProvider.ts` is named by a class in it, but not the exported class.

apps/extension/src/App/App.tsx
```diff
@@ -83,6 +84,11 @@ export const App: React.FC = () => {
     }
   }, [status, accounts]);
 
+  const parent = accounts.find(
+    (account) => account.type === KeyStoreType.Mnemonic
+  );
```
memasdeligeorgakis, line 89: Q: We have just single one with `KeyStoreType.Mnemonic`, right? Is that a direct `parent` to all derived ones?

apps/extension/src/App/Accounts/AddAccount.tsx
```diff
@@ -137,19 +146,12 @@ const AddAccount: React.FC<Props> = ({ accounts, requester, setAccounts }) => {
             <p>HD Derivation Path</p>
             <Bip44PathContainer>
               <Bip44PathDelimiter>
-                {[bip44Prefix, coinType].join("/")}/
+                {[bip44Prefix, coinType, parentAccount].join("'/")}&apos;/
```
memasdeligeorgakis, line 149: nit: I feel it would be cleaner to use template string
```suggestion
                {`${bip44Prefix}'/${coinType}'/${parentAccount}'/`}
```

apps/extension/src/App/Accounts/AccountListing.tsx
```diff
@@ -6,43 +7,59 @@ import {
...
       <Details>
-        {alias && <Alias>{alias}</Alias>}
         <DerivationPath>
-          m/44'/{coinType}/{`${bip44Path.account}'`}/{bip44Path.change}/
-          {bip44Path.index}
+          {type !== KeyStoreType.Mnemonic &&
+            `${parentAlias} /${path.account}'/${path.change}${
+              typeof path.index === "number" ? "/" + path.index : ""
+            }`}
         </DerivationPath>
```
memasdeligeorgakis, line 40: nit: I feel that moving logic out of "html" adds readability and makes it less error-prone
```suggestion
  const navigate = useNavigate();
  
  // if we have a non-mnemonic account, we display the path
  const isNonMnemonicAccountType = type !== KeyStoreType.Mnemonic;
  const maybePathIndex = typeof path.index === "number" ? "/" + path.index : "";
  const formattedDerivationPath = isNonMnemonicAccountType
    ? `${parentAlias} /${path.account}'/${path.change}${maybePathIndex}`
    : "";
  return (
    <AccountListingContainer>
      <Details>
        <DerivationPath>
          {formattedDerivationPath}
        </DerivationPath>
```

apps/extension/src/App/Accounts/AccountListing.tsx
```diff
@@ -6,43 +7,59 @@ import {
...
-import { DerivedAccount } from "background/keyring";
+import { DerivedAccount, KeyStoreType } from "background/keyring";
+import { TopLevelRoute } from "App/types";
 
 type Props = {
+  // The parent Bip44 "account"
+  parent?: number;
+  // The child account
   account: DerivedAccount;
+  parentAlias: string;
```
memasdeligeorgakis, line 21: very nit: `parentAlias` gets a value that is based on `alias` which is optional in `DerivedAccount`. It would be cleaner to treat it as such until it is being utilised in this file and remove the logic to turn a possible undefined to `""` in the parent. Because now in the parent we do not know why the logic is there without reading this file.

memasdeligeorgakis (approved):
Left some follow ups [here](https://github.com/anoma/namada-interface/issues/119) and [here](https://github.com/anoma/namada-interface/issues/4#issuecomment-1278793284)

apps/extension/src/App/App.tsx
jurevans: Technically we could have as many as we want, but right now, it's just the first one that gets created, so for now it's just one. This will be made a bit more robust if that need ever arises, which it may.

jurevans:
> Not directly part of this PR, but `namada-interface/packages/storage/src/store/MemoryKVStoreProvider.ts` is named by a class in it, but not the exported class. I think it should be `MemoryKVStore.ts`.

That class has been removed entirely in #105, as we don't use it, or have a use for it.

apps/extension/src/background/keyring/crypto.ts
```diff
@@ -0,0 +1,65 @@
...
+    aes.free();
+
+    return {
+      alias,
+      address,
+      id,
+      path,
+      crypto: {
+        cipher: {
+          type: "aes-256-gcm",
```
leontiad, line 33: is there a technical reason we switched to aes-256-gcm? I would rather prefer XChaChaPoly as the default symmetric cipher mostly due to a rare but existing attack on aes-256-gcm: https://www.usenix.org/system/files/sec22summer_albertini.pdf

jurevans: Just for historical reference, `namada-interface` is actually using this package for encrypting stored keys: https://www.npmjs.com/package/redux-persist-transform-encrypt, and using `orion::aead` (which uses `XchaChaPoly1305`) only for the seed phrase (which is stored in `window.localStorage`). I've opened https://github.com/anoma/namada-interface/issues/124 to move our encrypted storage to use `XChaChaPoly`. In the extension, the stored mnemonic and private keys share the same encryption cipher (unlike how it is in `namada-interface` currently), and once the interface is loading accounts from the extension, it will no longer need to store any data itself. All accounts in the extension are stored in IndexedDB.

leontiad: can we keep somewhere an up to date document with the workflow of that protocol?Is that related with that? https://github.com/anoma/namada-interface/commit/85c80072468411950268b637a91cd6843d394162 and line 23? It is not very clear what is meant "encrypting stored keys".  Can we short describe it somewhere? as in the link above?

PR #100, opened 2022-09-30
fix(theme, colors): theme changes to dark after refreshing the app

### LINKED ISSUE #26: BUG - User's selected theme is lost on page refresh
The theme is reverted to the default (light mode) any time the page is refreshed. This isn't a severe issue, but is indeed annoying and likely straight-forward to resolve.

This occurs during development as well. Any time a non-React file is modified, the page is reloaded and dark mode reverts to light mode.

mateuszjasiuk:
Resolves #26

### DIFF
--- a/apps/namada-interface/src/App/App.tsx
+++ b/apps/namada-interface/src/App/App.tsx
@@ -6,7 +6,12 @@ import { AnimatePresence } from "framer-motion";
 import { ThemeProvider } from "styled-components/macro";
 
 // internal
-import { getTheme } from "utils/theme";
+import {
+  getTheme,
+  loadColorMode,
+  storeColorMode,
+  ColorMode,
+} from "utils/theme";
 import { TopLevelRoute, locationToTopLevelRoute } from "./types";
 
 import { TopNavigation } from "./TopNavigation";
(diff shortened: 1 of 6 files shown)


apps/namada-interface/src/App/App.tsx
```diff
@@ -54,15 +59,22 @@ const getShouldUsePlaceholderTheme = (location: Location): boolean => {
...
 function App(): JSX.Element {
-  const [isLightMode, setIsLightMode] = useState(true);
+  const initialColorMode = loadColorMode();
+  const [colorMode, setColorMode] = useState<ColorMode>(initialColorMode);
   const location = useLocation();
   const [password, setPassword] = useState<string>();
   const [store, setStore] = useState<AppStore>();
   const [persistor, setPersistor] = useState<Persistor>();
-
+  const isLightMode = colorMode == "dark";
```
mateuszjasiuk: Unfortunately this is "correct" for now :) I will fix it in the next PR.

apps/namada-interface/src/utils/theme/theme.ts
```diff
@@ -333,6 +334,17 @@ export enum ThemeName {
   Placeholder,
 }
 
+export type ColorMode = "light" | "dark";
```
mateuszjasiuk, line 337: I went with union type instead of enum here. There are few upsides. It does not produce new object when compiled. Potentially you can "compare" to it without importing it explicitly. You can export it as a type to avoid potential circular dependencies.

apps/namada-interface/src/App/App.tsx
```diff
@@ -78,7 +90,8 @@ function App(): JSX.Element {
             <TopSection>
               <TopNavigation
                 isLightMode={isLightMode}
-                setIsLightMode={setIsLightMode}
+                toggleColorMode={toggleColorMode}
+                setColorMode={setColorMode}
```
mateuszjasiuk, line 94: This one has to be passed, so we change theme to light when navigating to governance/staking.

jurevans (approved):
This looks good to me! Works perfectly in the demo at https://pull-request-100--wallet-development-heliax-dev.netlify.app/. Would probably be a good idea for @memasdeligeorgakis to look at it too, as he worked on the theme functionality

PR #103, opened 2022-10-04
refactor(theme, colors): replace isLightMode with colorMode in regular components

mateuszjasiuk:
components

Changes:
- nothing visual should change
- `isLightMode` passed as a props is replaced by `colorMode`
- I've removed few unused files

In the next PR I will replace `isLightMode` in the styled components that refer to theme :)
I've a draft PR with more changes but I figured out, that smaller ones will be nicer to review


jurevans (approved):
Looks good to me!

apps/namada-interface/src/App/TopNavigation/topNavigation.tsx
```diff
@@ -315,7 +315,10 @@ function TopNavigation(props: TopNavigationProps): JSX.Element {
               </LogoContainer>
               <ColorModeContainer>
                 {topLevelRoute !== TopLevelRoute.StakingAndGovernance && (
-                  <Toggle checked={isLightMode} onClick={toggleColorMode} />
+                  <Toggle
+                    checked={colorMode == "dark"}
```
memasdeligeorgakis, line 319: I think it would always best to default to strict equality when there is no special need for the abstract equality. I saw this slipped in already at the previous PR, so maybe would be still good to clean that up.
```suggestion
                    checked={colorMode == "dark"}
```

memasdeligeorgakis (approved):

Issue #92, opened 2022-09-29
Extension: Bug - Service worker goes inactive while user is logged in


jurevans:
When the extension is running as `manifest_version: 3`, with a service worker, Chrome does not guarantee that the service worker will remain active during the user's session, which can lead to deauthentication, even after a short period of inactivity. Unlike persistent `background` pages in Firefox (`manifest: 2`) service workers will inevitably terminate (this is also true in Firefox `manifest: 3`).

See <https://stackoverflow.com/questions/66618136/persistent-service-worker-in-chrome-extension> for information on how to work around this, and to achieve better control over the lifetime of the service worker. This is not an issue in Manifest v2 extensions, therefore, Firefox is unaffected by this.

This bug can be observed by running the extension in a Chrome-based browser, as of https://github.com/anoma/namada-interface/pull/84

Steps to recreate:
1. Create an account in the extension (or log in to an existing account)
2. Navigate to `Derive new account` screen
3. Wait until the service worker becomes inactive (this can be viewed from the `Manage Extensions` page)
4. Attempt to derive a new address, and observe that it fails (which is due to the service worker no longer having the user's credentials)

It may be good in the interim to periodically check that the extension is authenticated before performing further actions, but the best solution is to keep the service worker alive until 1) the user manually locks the extension, or 2) a period of inactivity that we have determined has passed

**NOTE** - This is resolved when the content-scripts establish a port to the background (service-worker) using the `browser.runtime.connect()` API. This should be performed either on `Anoma.connect("chainId")`, or upon successful login. Either of those will re-establish the port and keep the service worker alive for a default of 5 minutes, which is reasonable for a session length.
- It may be worth it at this point to update the `ExtensionRouter` class to establish all connections over a port with the `runtime.connect()` API, which would, by default, maintain a connection to the service worker. This may be the cleanest long-term solution.

memasdeligeorgakis:
In google's [documentation for migrating from manifest v2 to v3](https://developer.chrome.com/docs/extensions/mv3/migrating_to_service_workers/#events) it is stated: *"service workers are a special execution environment that are started to handle events they're interested in and are terminated when they're no longer needed"*

[...]

Issue #63, opened 2022-09-15
Extension: Sign transactions (Transfer & IBC)


jurevans:
The extension should be able to sign a transfer transaction.

- User is logged in to the extension and has connected to `namada-interface`
- For a given account, they can initiate a transfer
- The extension will accept tx parameters (account address, recipient, token, amount, gas fee) and return a serialized and signed, wrapped transaction that can be broadcast from `namada-interface` to the ledger
  - `namada-interface` will need the Tx hash, as well as encoded bytes (a base64 string), in order to submit an JSON RPC request to broadcast to the ledger.
  - The private key (secret) is only accessed within the service worker (Chrome) or background scripts (Firefox) for the given account's public key (address)

The extension could also provide the functionality to sign arbitrary data and verify signatures given a stored keypair, which would be useful for non-Namada transactions in the future. This is relatively straight-forward to implement, as we could assume we know nothing about how data is serialized. Something to think about, but not necessarily critical for this specific issue.

### Notes
- The shared Rust libs in `@anoma/shared` could be updated to accept Borsh-serialized data (for 3 types, `transfers`, `ibc-transfers`, and `shielded-transfers`) for which we can define shared classes between the extension and `namada-interface`, and the signer, which will be accessed via querying the extension from `namada-interface`, will serialize all transaction data before submitting via `postMessage` to proxy into the extension's background services, where the transaction will be signed. The resulting data passed back from the extension should be the final result of wrapping, encrypting, and signing the transaction, which should return two values: the transaction hash, and the transaction data (`base64`-encoded). The private key for the account will only be decrypted long enough to sign the transaction within the background service.

jurevans:
Closing as irrelevant, and handled by SDK

PR #105, opened 2022-10-04
feat/63 - Extension: Sign transaction data with extension

jurevans:
See https://github.com/anoma/namada-interface/issues/63

_NOTE_ as part of this PR, I've updated all occurrences of linting errors and warnings (per discussion around https://github.com/anoma/namada-interface/pull/71) , which added about 20 files to this PR, mostly very minor changes.

- [x] Tx args should be serialized before passing to extension
- [x] `@anoma/shared` should be updated to accept serialized (binary) transaction data
- [x] Signing functionality should be accessible to `namada-interface`
  - [x] The interface connected to the extension should be able to request the signer for an account
  - [x] The interface should be able to construct the necessary transaction data needed to pass to the extension
  - [x] Signer should return the signed `hash` and `bytes` of the transaction
- [x] As this PR updates `InjectedAnoma`, do some refactoring to better-type the proxy and `requestMethod` functionality
- [x] Run linter on `apps/extension` and update accordingly

## Testing

To test this functionality, you can interact directly with the extension via the browser console.

**NOTE** Address encoding has been fixed in https://github.com/anoma/namada-interface/pull/114/commits/6e2a31d14fd2068ced44a342154106de6aaa9ab8 - where the encoding now matches what you would expect to see.

**NOTE** Until Issue #92 is fixed, it may be much easier to test this in Firefox. If you do test in Chrome, make sure you have authenticated the extension before running the following:

1. Create an account on the extension
2. Derive any number of addresses  (optional - you can use the default root account to sign transactions too)
3. Log-in to the extension, and copy one of the addresses (for this example, I'll use: `A51B6CEEC4D0C3EE08F80E9D105D11ADD0C83063`)
4. In the console, issue the following commands:

```javascript
// Copy any address from your wallet and set it to `address`:
const address = "A51B6CEEC4D0C3EE08F80E9D105D11ADD0C83063";

// You can copy and past the following, or update parameters as you see fit:
const signer = anoma.getSigner("anoma-test.XXXXXXXXXXX");

const transferArgs = {
  source:  "atest1v4ehgw368ycryv2z8qcnxv3cxgmrgvjpxs6yg333gym5vv2zxepnj334g4rryvj9xucrgve4x3xvr4",
  token: "atest1v4ehgw36x3prswzxggunzv6pxqmnvdj9xvcyzvpsggeyvs3cg9qnywf589qnwvfsg5erg3fkl09rg5",
  target: "atest1v4ehgw36xvcyyvejgvenxs34g3zygv3jxqunjd6rxyeyys3sxy6rwvfkx4qnj33hg9qnvse4lsfctw",
  amount: 1000,
};

const txData = await signer.encodeTransfer(transferArgs);

// Set up transaction parameters:
const txProps = {
    token: "atest1v4ehgw36x3prswzxggunzv6pxqmnvdj9xvcyzvpsggeyvs3cg9qnywf589qnwvfsg5erg3fkl09rg5",
    epoch: 5,
    feeAmount: 1000,
    gasLimit: 1000000,
    txCode: new Uint8Array([]),
};

// Double-check that you're logged into the extension (that it's unlocked), then do the following:
const { hash, bytes } = await signer.signTx(address, txProps, txData);

console.log(hash, bytes);

// This should have succeeded. Note that it may fail if the Chrome service worker goes inactive (just log-in again)
```

## Notes

- I updated the `@anoma/shared` types for `Transfer` and `IbcTransfer` to follow what is done on Namada for consistency and precision (see: <https://github.com/anoma/namada/blob/main/shared/src/types/token.rs#L16-L34>)
- In this PR, I've updated the structs in `@anoma/shared` to accept `u64` values where required by types in Namada, and the values for these structs is serialized using Borsh. We should handle values for `amount` and `epoch` with `u64` (especially in the case of `amount`, we want to preserve precision, which is lost using `f32`, as we did before, where we'd _convert_ back to `u64` in Rust). Also, as a consequence of Borsh serialization, transaction details are encoded before being submitted over `postMessage`, which isn't a bad thing. The libs in `@anoma/shared` are updated to accept encoded messages, and handle deserialization on the backend (service worker). The workflow should follow this:
  1. The interface creates an instance of a `Message` (e.g, `TransferMsg`), and hands it off to the `Signer`
  6. The `Signer` serializes this data before proxying to the extension Service-Worker
  7. The Service-Worker is responsible for decrypting the associated private key from storage and signing the transaction (via the `@anoma/shared` wasm functionality for `Transfer`), then returning the resulting data to the `Signer`
  8. The `Signer` returns the resulting Tx `hash` and `bytes` to the interface, which at this point can broadcast to the ledger
- As such, I've updated the following types to `u64` in `@anoma/shared` to match what is required by Namada:
  - `amount`
  - `fee_amount`
  - `gas_limit`
  - `epoch`
- I've kept the decrypting of the private key as close to storage as possible, and should only occur when signing a Tx and never leaving the confines of the service worker. All access to private keys is performed within `background/keyring`, which is the only class with access to the stored keys
- It's possible to sign a transaction with the root/default account (`m/44'/1'/0'/0`). The two types of stored accounts are `Mnemonic` and `PrivateKey` - A mnemonic stores an encrypted seed phrase, which can be used to derive a private key for signing a transaction. A `PrivateKey` type account is a normal, derived account (e.g., `m/44'/1'/0'/0/1`) which contains an encrypted private key, so this can be directly decrypted to sign a transaction.
- `coinType`, for BIP44 derivation paths, is now being set in the default chain config in the extension. In the config, this value is set to `1`, which is the designation for _testnet_ chains (all coins). See: https://github.com/satoshilabs/slips/blob/master/slip-0044.md


apps/extension/src/provider/InjectedAnoma.ts
```diff
@@ -1,176 +1,83 @@
...
+  public async accounts(chainId: string): Promise<DerivedAccount[]> {
+    return await InjectedProxy.requestMethod<string, DerivedAccount[]>(
+      "accounts",
+      chainId
+    );
+  }
 
-  public async connect(chainId: string): Promise<void> {
-    return await this.requestMethod("connect", chainId);
+  public getSigner(chainId: string): ISigner<Account> | undefined {
```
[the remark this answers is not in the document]
jurevans: The `Account` type should only be a subset of `DerivedAccount`. This is because an account returned to the interface should only have _some_ of the fields, such as alias, address, and likely public-key at some point, but shouldn't return anything else. The signer will only return those few fields to the interface, not the full stored account data.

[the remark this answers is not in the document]
jurevans: @memasdeligeorgakis Also, that is correct that `namada-interface/packages/tx/src/Account.ts` is unrelated to the extension. That package is used by `namada-interface`, and until the extension is integrated, that will need to remain.

packages/types/src/signer.ts
```diff
@@ -0,0 +1,24 @@
...
+import {
+  IbcTransferProps,
+  InitAccountProps,
+  SignedTx,
+  TransferProps,
+  TxProps,
+} from "./tx";
+
+export interface Signer<T = Account> {
+  accounts: () => Promise<T[] | undefined>;
```
memasdeligeorgakis, line 11: 1
```suggestion
export interface ISigner<T = Account> {
  accounts: () => Promise<T[] | undefined>;
```
It seems that we are aliasing this at a few imports as `ISigner` as `Signer` class is implementing it. As we have a class of same name it would make searching for their usage much easier to have different names for them.


2
Also I am wondering why do we not speak about just `Account[]` here instead of `T[]`. To me it seems that we are using this `interface Account` in 2 occasions:
* in one of them we use `Account` for the generic type
* in the other occasion we just omit it, so it defaults this this `Account`

Or are going to have some future use case where `T` will have to be something different than `Account`?

```suggestion
export interface Signer {
  accounts: () => Promise<Account | undefined>;
```

packages/types/src/signer.ts
```diff
@@ -0,0 +1,24 @@
...
+  signTx(
+    signer: string,
+    txProps: TxProps,
+    encodedTx: string,
+    txData: string
+  ): Promise<SignedTx | undefined>;
```
memasdeligeorgakis, line 17: Is `encodedTx: string` intentional here? When I do a global search for `encodedTx` this is the only hit I get. Due to the structural type system we are allowed to omit this in the implementing class and we are omitting it in `class Signer implements ISigner` in `namada-interface/apps/extension/src/provider/Signer.ts`.
```suggestion
  signTx(
    signer: string,
    txProps: TxProps,
    txData: string
  ): Promise<SignedTx | undefined>;
```

packages/types/src/signer.ts
jurevans: Yeah we could probably just use `Account` and not have it be generic. I think I had originally thought that there may be additional types that could be used here, but I don't think a generic is necessary.

apps/extension/src/config/chains.ts
```diff
@@ -27,7 +27,9 @@ const defaultChain: Chain = {
   chainName: "Namada Testnet",
   stakeCurrency: ibcCurrency,
   bip44: {
-    coinType: 9999,
+    // coinType = testnet (all coins) - Slip-0044
+    // See: https://github.com/satoshilabs/slips/blob/master/slip-0044.md
+    coinType: 1,
```
memasdeligeorgakis, line 32: We would need to support multiple coin types per chain here, right?

memasdeligeorgakis: ...or this will be "namada" once we register it there, and the currencies the chain supports are listed under `currencies`.

jurevans: We just need a `slip-0044` address for NAM, not for each coin-type per chain, so once that is registered, we can use that for production. `1` here denotes "testnet"/all-tokens. Having `NAM` registered on [slip-0044](https://github.com/satoshilabs/slips/blob/master/slip-0044.md) will help ensure a user's addresses are unique to our network. That said, we should have a "development" and "production" chain config (in a future PR) with the value set accordingly.

memasdeligeorgakis (approved):
I created a follow up ticket to more document some key components https://github.com/anoma/namada-interface/issues/125

Issue #3, opened 2022-04-20
UI Todos/Improvements


memasdeligeorgakis:
# Description
This is for collecting UI related TODOs and improvement ideas in the app.

# Outcomes
* comments under this ticket listing the TODOs. When the work on them starts a new task should be created for each.

memasdeligeorgakis:
It is just a placeholder so might not matter, but adjust the layout for mobile size, to not waste space

<img width="480" alt="image" src="https://user-images.githubusercontent.com/2773320/191051971-8e60d5b4-8650-45f4-99e8-16720a6b6114.png">

memasdeligeorgakis:
Left align the breadcrumbs, as it looks a bit awkward like it is now (right chevron at center).
<img width="348" alt="image" src="https://user-images.githubusercontent.com/2773320/191056303-928dc0f2-f113-4734-9ae5-e337ab09b593.png">

jurevans:
@memasdeligeorgakis I think we can probably close this issue, but I can leave it open for now in case we want to update it with recent issues.

memasdeligeorgakis:
Improve the naming of `themeConfigurations` object in `DesignConfiguration`

memasdeligeorgakis:
This component expects that the address is there. Otherwise crashes and the whole page does not render. It should work, but now this is coupled with and depending on the logic of whichever component is using this.

<img width="659" alt="image" src="https://user-images.githubusercontent.com/2773320/203545871-f867bfe2-ee0e-4538-b328-c08ea5078d46.png">

memasdeligeorgakis:
Let's alias the `string` or add a comment to inform what they represent.

<img width="655" alt="image" src="https://user-images.githubusercontent.com/2773320/203548174-c28b1bbb-9158-4167-944b-97f72b06558b.png">

memasdeligeorgakis:
rendering a whole object at token sending 
<img width="473" alt="image" src="https://user-images.githubusercontent.com/2773320/206463077-c5479214-aaec-4419-acbd-3b837474eba5.png">

memasdeligeorgakis:
<img width="265" alt="image" src="https://user-images.githubusercontent.com/2773320/207804637-a2de8c97-a6a8-4416-ba96-cf9280c4c3c3.png">

Sizing/styling for this

jurevans:
Closing, will address issues as they arise after implementing new designs

PR #107, opened 2022-10-04
refactor(theme, colors): add colorMode to the styled components

mateuszjasiuk:
Replaces isLightMode in components that refer to theme.
Additionally fixed one label color in dark mode :)


apps/namada-interface/src/utils/theme/theme.ts
```diff
@@ -101,7 +101,10 @@ export type DesignConfiguration = {
   spacers: Spacers;
   borderRadius: BorderRadius;
   typography: Typography;
-  themeConfigurations: { isLightMode?: boolean; themeName: ThemeName };
+  themeConfigurations: {
```
jurevans, line 104: Just a thought, and I realized this was not introduced in this PR, so this comment can be ignored! But thinking about the naming convention here, it's usage requires:

`DesignConfiguration.themeConfigurations` and `DesignConfigurations.themeConfigurations.themeName`, which feels quite redundant and hard to read. I see this in a lot of places in our code, and feel like the following is adequately clear:

```typescript
export type DesignConfiguration = {
  ...
  theme: {
    colorMode: ColorMode;
    name: ThemeName;
  }
}
```
Instead of https://github.com/anoma/namada-interface/blob/360c09425f0799b7ffe67899d94e7f572d2cd1b5/apps/namada-interface/src/utils/theme/theme.ts#L99-L108

I think the hierarchical structure of this type makes it clear what each thing is. Like I said, not important to this PR, but just a thought I had! I'm curious what others think.

jurevans: I also realize it's a big hassle to change, now that we're loading the theme in styled-components props all over the place, so definitely not a concern for this PR!

jurevans (approved):
Looks good to me! I left a comment (unrelated to the work in this PR, but something to bring up to think about maybe for later for theme configs)

apps/namada-interface/src/utils/theme/theme.ts
memasdeligeorgakis: I introduced this some time ago for the purpose that we might need these constants to define exceptions in the use sites where we use the theme.

styled components has this root level `theme` object by default and we defined it's type in `/src/utils/theme/styled.d.ts` by `export interface DefaultTheme extends DesignConfiguration {}`.

I feel that the whole thing is theme, this is just part of the theme, so maybe a more specific name could be more descriptive. But I agree let's do this seperately, I added a note https://github.com/anoma/namada-interface/issues/3#issuecomment-1271288269.

so we use it like:
```ts
// now
const { colorMode } = theme.themeConfigurations;

// maybe one of these
const { colorMode } = theme.configurations;
const { colorMode } = theme.constants;
const { colorMode } = theme.parameters;
```

apps/namada-interface/src/App/AccountOverview/DerivedAccounts/DerivedAccounts.components.ts
```diff
@@ -1,19 +1,26 @@
...
-import { DesignConfiguration } from "utils/theme";
+import { ColorMode, DesignConfiguration } from "utils/theme";
 
 enum ComponentColor {
   ScrollbarThumb,
 }
 
 const getColor = (
-  color: ComponentColor,
+  toggleColor: ComponentColor,
```
memasdeligeorgakis, line 9: nit: I feel that the naming here might be `forComponent`, but the old `color` from me was also too ambiguous.

apps/namada-interface/src/components/NavigationContainer/NavigationContainer.tsx
```diff
@@ -32,9 +32,10 @@ export const NavigationContainer = (
 ): React.ReactElement => {
   const { children, onBackButtonClick } = props;
   const themeContext = useContext(ThemeContext);
-  const backButtonIconStrokeColor = themeContext.themeConfigurations.isLightMode
-    ? themeContext.colors.utility2.main80
-    : "black";
+  const backButtonIconStrokeColor =
+    themeContext.themeConfigurations.colorMode == "light"
```
memasdeligeorgakis, line 36:
```suggestion
    themeContext.themeConfigurations.colorMode === "light"
```

memasdeligeorgakis (approved):

PR #112, opened 2022-10-07
feat: notifications

mateuszjasiuk:
Notifications!
I've added them only in one place for now.
When you create new shielded account there is a
- info toast with information that account is being created
- success toast after success


apps/namada-interface/src/slices/AccountsNew/actions.ts
```diff
@@ -34,13 +63,27 @@ type ShieldedKeysAndAddressesWithNewAccountDetails = {
...
 >(
   ADD_ACCOUNT_TO_LEDGER,
-  async (newAccountDetails: NewAccountDetailsWithPassword) => {
+  async (
+    newAccountDetails: NewAccountDetailsWithPassword & { notify?: boolean },
+    thunkAPI
+  ) => {
+    const { chainId, alias, password, notify } = newAccountDetails;
+
+    if (notify) {
```
mateuszjasiuk, line 75: I do not really like this notify flag. Added it because `createShieldedAccount` is used for both adding new accounts and loading initial ones. I think the best solution would be to split those actions into two.

apps/namada-interface/src/components/Toast/toast.components.ts
```diff
@@ -0,0 +1,60 @@
...
+
+export const Wrapper = styled(motion.div)<{ index: number }>`
+  display: flex;
+  justify-content: space-between;
+  position: fixed;
+  transition: top 0.5s;
+  top: ${(props) => 100 + props.index * 70}px;
+  right: 15px;
+  width: 360px;
+  z-index: 10;
```
mateuszjasiuk, line 12: Was not sure about value :)

memasdeligeorgakis: Yeah, we have not defined them in the design system. With `z-index: 10;`, the select element goes on top of this so, maybe to play it safe as long as we define them we could set it to:
```suggestion
  z-index: 9999;
```

memasdeligeorgakis:
I saw that when getting notifications when creating the accounts, it seemed a bit laggy, but it is due to the action for creating the account which is blocking the main thread, not the notification or the animations. Running in isolation it is fine.


https://user-images.githubusercontent.com/2773320/194604527-8eb8c23a-13c5-4d9a-a05c-579db48a4f4c.mov

memasdeligeorgakis (approved):
Looks good!

mateuszjasiuk:
> I saw that when getting notifications when creating the accounts, it seemed a bit laggy, but it is due to the action for creating the account which is blocking the main thread, not the notification or the animations. Running in isolation it is fine.

Yeah, tested it on it's own and it works fine.

PR #114, opened 2022-10-11
feat/95 - Extension address encoding, type improvements

jurevans:
This PR was originally intended to implement a solution to #92, but I've removed that functionality as it isn't a good long-term solution, and could be better handled in other ways.

This PR adds the following:

- Proper encoding for addresses (Bech32m)
- Improve styling of addresses (Bech32m addresses are very long, so I've add a function to truncate in a similar fashion to Keplr)
- Improve type definitions
- Move Message definitions to appropriate parts of the application
- Adds autofocus and Enter key binding to forms


memasdeligeorgakis:
In google's [documentation for migrating from manifest v2 to v3](https://developer.chrome.com/docs/extensions/mv3/migrating_to_service_workers/#events) it is stated: "service workers are a special execution environment that are started to handle events they're interested in and are terminated when they're no longer needed"

It also [says](https://developer.chrome.com/docs/extensions/mv3/migrating_to_service_workers/#state): "One of the main things to get used to when adopting service workers is that they are short-lived execution environments. In more practical terms, an extension's service worker will start up, do some work, and get terminated repeatedly throughout a user's browser session. This poses a challenge to extension developers accustomed to long-lived background pages as application data is not immediately available in global variables."

I assume that this PR is addressing this "terminated when they're no longer needed"? If that is the case I am wondering if we are trying to persist a state in with/in service workers in a way that is not intended 🤔

apps/extension/src/extension/Session.ts
```diff
@@ -0,0 +1,120 @@
...
+    if (requester) {
+      this._requester = requester;
+    }
+
+    this.port.onMessage.addListener(this.handleOnMessage);
+    this.port.onDisconnect.addListener(this.handleOnDisconnect);
+
+    this._timeout = setTimeout(() => {
+      const duration = Date.now() - this._timestamp;
+      if (duration < SESSION_DURATION) {
```
memasdeligeorgakis, line 61: Can this condition ever be true (and cause us to update the session)? as there are always a few ms from where we set `this._timestamp` to where we call `setTimeout`.

The purpose of this method is to start a session and at the end of the timeout automatically send `SessionMsgType.Disconnect` to terminate the session or it should also be keeping it alive?

apps/extension/src/extension/Session.ts
```diff
@@ -0,0 +1,120 @@
...
+/**
+ * Maintain an open port to the background until a certain limit has been reached.
+ */
+export class Session {
+  protected port: Runtime.Port | undefined;
+  private _requester: ExtensionRequester | undefined;
+  private _timestamp: number = Date.now();
+  private _timeout: Timeout | undefined;
+
+  public start(requester?: ExtensionRequester): Runtime.Port {
```
memasdeligeorgakis, line 43: nit: I feel that a short description of the task of this method would be especially helpful here.

apps/extension/src/background/session/handler.ts
```diff
@@ -0,0 +1,31 @@
+import browser, { Runtime } from "webextension-polyfill";
+import { KeyRingService } from "../keyring";
+import { SessionMsg, SessionMsgType } from "extension";
+
+export class SessionHandler {
```
memasdeligeorgakis, line 5: Do we actually need a class here? It seems that the only site where we initiate this is not persisting it, we just end up calling the constructor where we set the listeners.

jurevans:
> I assume that this PR is addressing this "terminated when they're no longer needed"? If that is the case I am wondering if we are trying to persist a state in with/in service workers in a way that is not intended

The only state we're trying to persist is the password within the extension. If a user logs into the extension, and no ports are active to the service worker, it will terminate in 30 seconds. I figured that this would be unacceptable UX, but maybe that isn't the case. Maybe any need for the password in a v3 extension should always prompts the user (e.g., when signing a transaction). This seemed like it would be a nightmare for testing if the extension deauthenticated after 30 seconds. Either way, it isn't particularly difficult to remove the Session functionality, as it sort of sits on the side and doesn't actually communicate much over the port it establishes. It's sort of a short term solution until we have approvals in place, where the user is prompted for a password to authorize a transaction. That's a separate issue though, and maybe when that is addressed, we simply remove the Session functionality at that point. It still seems like clunky UX, having to authenticate every single time you view the extension, but that's what Manifest v3 forces us to do without nasty workarounds.

apps/extension/src/background/session/handler.ts
jurevans: This doesn't necessarily need to be a class.

apps/extension/src/extension/ExtensionRequester.ts
```diff
@@ -59,4 +65,8 @@ export class ExtensionRequester {
 
     return result.return;
   }
+
+  startSession(): void {
```
memasdeligeorgakis, line 69: We call `startSession()` from several places where we want to post any messages using requester. Could it be an issue or leave to unintended behavior if several sessions are "started" in parallel? I see it ends up calling this

```ts
const SESSION_PORT = "session-port";
browser.runtime.connect({
  name: SESSION_PORT,
});
```

apps/extension/src/extension/Session.ts
jurevans: This has been removed.

memasdeligeorgakis (approved):

PR #117, opened 2022-10-13
feat(spinner): add pending actions to state

### LINKED ISSUE #69: Create Loader
# Description
The application needs a unified loader implementation. This is for the situations when the UI is performing a long-lasting action. It could be as simple as: 
- displaying a spinner component
- having a global state indicating the status
- all necessary interactive components could use context to become disabled for the time of loading

This depends on:
- being able to display [notifications](https://github.com/anoma/namada-interface/issues/70), to inform the user of failed loading or after a timeout.

mateuszjasiuk:
Resolves #69 

You can test it while transfering tokens!
ui+state


apps/namada-interface/src/slices/notifications.ts
```diff
@@ -51,6 +60,22 @@ export const NotificationsSlice = createSlice({
       state.toasts = newToasts;
     },
   },
+  extraReducers: (builder) => {
+    builder.addMatcher(isAsyncThunkAction, (state, action) => {
```
mateuszjasiuk, line 64: This one is pretty generic but it works quite well!
It matches `asyncThunkActions` and based on that either adds action as a pending one or removes it from state.
We  can then use it like that:
```ts
const isTransferSubmitting = useAppSelector<boolean>((state) =>
    state.notifications.pendingActions.includes(
      actionTypes.SUBMIT_TRANSFER_ACTION_TYPE
    )
  );
  ```

Comment (jurevans) [reply apps/namada-interface/src/slices/notifications.ts:64]: That's nice!

**Review by jurevans — APPROVED** [review]
Looks good to me!

**Comment by mateuszjasiuk** [comment]
Noticed same issue as with account creation toasts. Will fix it soon! :)

File: apps/namada-interface/src/slices/notifications.ts (hunk 1 above)
Comment (memasdeligeorgakis) [reply apps/namada-interface/src/slices/notifications.ts:64]: Looks good! nit: I guess there might be cases where we want to indicate something to be pending and reflect that in the UI with this mechanism, but not need a notification. In that sense having this under `state.notifications` might not be the perfect place, but that is no big deal to change at any time.

File: apps/namada-interface/src/slices/notifications.ts [hunk 2]
```diff
@@ -51,6 +60,22 @@ export const NotificationsSlice = createSlice({
...
+      const [actionType] =
+        action.type.match(/^.*(?=\/(pending|fulfilled|rejected)$)/) || [];
```
memasdeligeorgakis, line 66: nit: even though it is obvious what we are after here, it might be pleasant to have a few practical examples. Alternatively, we could have a follow up task to put this to it's own folder and have a small test for it.
```suggestion
        // match example "actionIdentifier/fulfilled" => ["actionIdentifier"]
        // match example "stakingAndGovernance/FETCH_VALIDATORS/pending" => ["stakingAndGovernance/FETCH_VALIDATORS"]
        action.type.match(/^.*(?=\/(pending|fulfilled|rejected)$)/) || [];
```

memasdeligeorgakis (approved):

apps/namada-interface/src/slices/notifications.ts
mateuszjasiuk: Agree, we can move it at anytime though :)

apps/namada-interface/src/slices/notifications.ts
mateuszjasiuk: Added tests

memasdeligeorgakis (approved):

jurevans (approved):
Looks good!

Issue #118, opened 2022-10-13
Extension: Update derivation and storage to support zip-0032


jurevans:
The extension should be updated to allow users to derive `zip-0032` addresses. This needs to work with the latest MASP testnet (see [#477](https://github.com/anoma/namada/pull/477)).

- When choosing to derive an account, user can choose between `Shielded` and `Transparent`
- Derivation path (and associated previews) should reflect derivation scheme (`bip-0044` vs `zip-0032`)
  - [master key generation](https://zips.z.cash/zip-0032#orchard-master-key-generation)
  - [child key derivation](https://zips.z.cash/zip-0032#orchard-child-key-derivation)
- Storage should reflect that this is a shielded account
- As shielded accounts have the following attributes, these could be serialized to string (e.g., `JSON.stringify(shieldedAccount)`, or via some other means) prior to encryption, then subsequently parsed/deserialized (e.g., `JSON.parse(decrypted)`):
  - Payment Address
  - Viewing Key
  - Spending Key 
- The API available to a connected interface should be able to request shielded accounts

## References

- [zip-0032 Shielded Hierarchical Deterministic Wallets](https://zips.z.cash/zip-0032)
- See [MASP Primitives](https://github.com/anoma/masp/tree/main/masp_primitives) for related types
- [masp-web - Shielded Accounts](https://github.com/anoma/namada-interface/blob/main/packages/masp-web/lib/src/shielded_account/mod.rs)

jurevans:
Closing as this is done

PR #120, opened 2022-10-17
WIP: feat/118 - Support zip32 derivation

jurevans:
Issue #118 

The extension should support shielded addresses using `zip-0032` derivation, using the `masp/masp_primitives` crate.

- [ ] Update build to support [unmerged MASP testnet](https://github.com/anoma/namada/tree/ray/masp-testnet-n002)
- [ ] Add library support for `zip-0032` derivation from 64-bit seed (mnemonic -> seed)
  - [ ] `ChildIndex`
  - [ ] `Coin-type`
  - [ ] `Account`
- [ ] Update `Transfer` transaction to accept additional required parameters (`key: String` and `shielded: masp_primitives::transaction::Transaction`)

More details _TBD_


apps/extension/src/background/keyring/keyring.ts
```diff
@@ -133,8 +150,65 @@ export class KeyRing {
...
+
+  public static deriveShieldedAccount(
+    seed: Uint8Array,
+    path: Bip44Path,
+    parentId: string
+  ): DerivedAccountInfo {
+    const { index = 0 } = path;
+    const id = getId("shielded-account", parentId, index);
+    const zip32 = new ShieldedHDWallet(seed);
+    const account = zip32.derive_to_serialized_keys(index);
```
mateuszjasiuk, line 190: i'm curious if it is possible(and makes sense) to return those those `masp::` types from this function? Because at the moment(if I understand correctly :)) we are serializing `xsk` `xfvk` and `payment_address` and deserializing them few lines below in the constructor of `ExtendedSpendingKey` and other classes.

apps/extension/src/background/keyring/keyring.ts
```diff
@@ -155,45 +229,55 @@ export class KeyRing {
...
-        alias,
-        address,
-        id,
-        password: this._password,
-        path,
-        text: derivedAccount.private().to_hex(),
-        type,
-      });
-      this._keyStore.append(keyStore);
+      if (type === AccountType.ShieldedKeys) {
```
mateuszjasiuk, line 243: To get rid off this big `if` and let statements you could do something like:
```ts
const accountDerivationFn = type === AccountType.ShieldedKeys ?  KeyRing.deriveShieldedAccount : KeyRing.deriveTransparentAccount
```

mateuszjasiuk (approved):
LGTM! Added few comments/questions.

apps/extension/src/background/keyring/keyring.ts
jurevans: I will take a look! They will need to be converted to a JSON serializable value before returning to TS (string or byte array), but I'll see if this can be cleaned up a bit.

packages/types/src/tx/schema/shielded.ts
```diff
@@ -0,0 +1,86 @@
...
+  txid: Uint8Array;
+  data: Uint8Array;
+
+  constructor(properties: ShieldedProps) {
+    this.txid = properties.txId;
+    this.data = properties.data;
+  }
+}
+
+export const ShieldedTransferMsgSchema = new Map([
```
memasdeligeorgakis, line 15: nit: let's evaluate again in near future if these would be just best living on Rust side, as the types likely anyway come from there and we get all these conversions automatically correct and for free.

memasdeligeorgakis (approved):

packages/types/src/tx/schema/shielded.ts
jurevans: Unfortunately we can't pass in `u64` types from TS natively, hence the schema here. But yeah, we can revisit it.

PR #149, opened 2022-11-22
feat(144): add test for extension message passing system

mateuszjasiuk:


apps/extension/src/test/init.ts
```diff
@@ -0,0 +1,76 @@
+import { KVStore } from "@anoma/storage";
+import { Chain } from "@anoma/types";
```
jurevans, line 2: I would remove this import if it's not going to be used in this file, same for `KeyStore` below

apps/extension/src/extension/ExtensionMessenger.ts
```diff
@@ -0,0 +1,23 @@
...
+export class ExtensionMessenger implements Messenger {
+  addListener(callback: Callback): void {
+    browser.runtime.onMessage.addListener(callback);
+  }
+
+  removeListener(callback: Callback): void {
+    browser.runtime.onMessage.removeListener(callback);
+  }
+
+  sendMessage(payload: any) {
```
jurevans, line 20: can we add a return type here?

jurevans (approved):
This looks good to me! I left a couple minor notes. My only suggestion really is to run it through the linter under `apps/extension` and clean up any errors & warnings. We could even just add `// eslint-disable-next-line` in cases where we intend to use an `any`, with a note stating intent.

memasdeligeorgakis (approved):

Issue #140, opened 2022-11-07
Interface: Update chain configs to specify extension integration configurations 


jurevans:
The chain configuration should allow settings for each chain we will integrate with.

For this issue, we should account for the following:

- Namada configuration specific to extension
- Chain configurations for integration with Keplr (Cosmos, Osmosis, etc.)
- Chain configurations for integrations with Metamask (Ethereum)
- Configurations should work for local, testnet, and live chains

This should be refactored in order to be straightforward to update in the future. For each possible chain that we integrate with, we should be able to instruct the user to connect to the appropriate extension when that chain is selected.

Issue #18, opened 2022-07-06
Basic implementation for integrating with extensions (Keplr, Metamask)


memasdeligeorgakis:
# Description
This should likely be chopped down to smaller pieces based on https://github.com/anoma/namada-interface/issues/5

# Outcomes
* Create PRs for implementing basic functionality to initiate incoming transfers from Keplr and Metamask

jurevans:
I've implemented a generic `Integrations` class in https://github.com/anoma/namada-interface/pull/153, which essentially achieves this. Each integration instance (Keplr, Anoma, etc.) is a class which implements the same interface, and in these classes we can have different implementations specific to the extension.

Another feature of this is that the accounts returned from Keplr can be mapped to our `Account` type, allowing the interface state to support it without additional logic. We would still need to implement logic specific to the extension for things like balance queries, but as far as the interface state is concerned, there should be no difference.

jurevans:
Closed as basic implementation is complete

PR #153, opened 2022-11-28
feat/140 - Multi-chain configurations

jurevans:
See #140 

This PR is to add multiple chain definitions shared by both the extension and namada-interface.

- [x] Create a chain type definition to cover all requirements for the interface (replacing usage of Keplr's `ChainInfo` type)
- [x] Import all chain configs on `namada-interface`, and import Namada chain config into `extension`
- [x] Allow overrides specific to environment
- [x] Configs should identify which _extension_ is required for connecting accounts (e.g., `anoma`, `keplr`, `metamask`, etc.)

__NOTE__ See  `apps/namada-interface/.env.sample` for update environment config values (I've added variables for Cosmos & Osmosis as well) - We can use defaults, but for local testing, you may need to update these. These `.env` values are also imported into the `webpack.config.js` build which makes them available in the extension. They aren't required, but will likely need to be updated for local testing.

__NOTE__ The Keplr integration is currently a bit wonky, but that is to be addressed in a separate issue (see #18 and #5).

## New variable overrides for environments

All of the following defaults can be overridden via the `.env` file in `apps/namada-interface` - these also apply to the extension, which loads the same `.env` (we only need one `.env` with the extension's `webpack.config.js` pointing to this one)

```bash
# NAMADA
REACT_APP_NAMADA_ALIAS=Namada Testnet
REACT_APP_NAMADA_CHAIN_ID=namada-1.5.32ccad5356012a7
REACT_APP_NAMADA_URL=http://127.0.0.1:26657
REACT_APP_NAMADA_BECH32_PREFIX=atest

# COSMOS
REACT_APP_COSMOS_ALIAS=Cosmos Testnet
REACT_APP_COSMOS_CHAIN_ID=cosmos-testnet.123412341234
REACT_APP_COSMOS_URL=http://127.0.0.1:12345

# OSMOSIS
REACT_APP_OSMOSIS_ALIAS=Osmosis Testnet
REACT_APP_OSMOSIS_CHAIN_ID=osmosis-testnet.123412341234
REACT_APP_OSMOSIS_URL=http://127.0.0.1:54321
```


packages/rpc/src/SocketClient.ts
```diff
@@ -25,7 +25,11 @@ class SocketClient extends RpcClientBase {
   private _client: WebsocketClient | null = null;
 
   public connect(): void {
-    this._client = new WebsocketClient(this.endpoint, (e: Error) => {
```
jurevans, line 28: NOTE: This file is going away completely, so I am just parsing the RPC HTTP URL here to force it into the appropriate websocket URL, but we won't be using websockets soon, this is just temporary.

packages/chains/src/chains/cosmos.ts
```diff
@@ -0,0 +1,32 @@
+import { Chain, Extensions } from "@anoma/types";
+
+const DEFAULT_ALIAS = "Cosmos Testnet";
+const DEFAULT_CHAIN_ID = "cosmos-75a7e12.69483d59a9fb174";
```
memasdeligeorgakis, line 4: Do we have any way to override these conveniently?

memasdeligeorgakis: But likely that is not MVP

memasdeligeorgakis (approved):

packages/chains/src/chains/cosmos.ts
jurevans: We can provide overrides in the `.env` file, which needs to be there when we want to point to a local testnet (for example) - these get loaded from `process.env` just below:

PR #167, opened 2022-12-09
feat/32 Remove WebSocket usage from checking when Tx gets applied

memasdeligeorgakis:
# What does this PR do
* uses polling instead of WebSocket to check whether a Tx got applied after broadcasting it
* Increased the timeout for the polling functionality to 40sec, this is as the time to get a response increases when the chain contains more blocks


memasdeligeorgakis:
This was also done in https://github.com/anoma/namada-interface/pull/166.

PR #169, opened 2022-12-13
feat: move integrations to the integrations service

### LINKED ISSUE #168: Move integrations context to the service
So it is easier to use extension provided object in slices.

mateuszjasiuk:
Resolves #168 

Added a service that does few things.
- It returns Integrations context and provider
- it defines hooks to access provider(s) and additional one to run functions "with" connection to extension
- It defines getters to access integration(s) inside slices if needed

### DIFF
--- a/apps/namada-interface/src/App/AccountOverview/AccountOverview.tsx
+++ b/apps/namada-interface/src/App/AccountOverview/AccountOverview.tsx
@@ -1,4 +1,4 @@
-import { useContext, useState } from "react";
+import { useState } from "react";
 import { useNavigate } from "react-router-dom";
 
 import { chains } from "@anoma/chains";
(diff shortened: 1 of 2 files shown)


apps/namada-interface/src/slices/transfers.ts
```diff
@@ -235,9 +237,11 @@ const createTransfer = async (
   const txCode = await fetchWasmCode(TxWasm.Transfer);
 
   // Invoke extension integration
-  const anoma = new Anoma(chains[chainId]);
-  anoma.connect();
-  const signer = anoma.signer();
+  const integration = getIntegration(chainId);
+  integration.detect();
+  //TODO: We have to treat this as anoma Signer for now
```
jurevans, line 242: This is fine, as use with any other extension means we'll need to use IBC actions, which will be handled separately. Only Namada will use Signer this way

jurevans (approved):
Looks very good!

memasdeligeorgakis (approved):

PR #173, opened 2022-12-16
feat: implement async client trait

mateuszjasiuk:
So we can use `RPC.shell()` and `RPC.vp()` calls!


jurevans (approved):
Looks good to me!

Issue #141, opened 2022-11-08
Extension: Prompt for password when unlocking the keyring


jurevans:
The extension needs to be update to incorporate the following:

1. When the user requests to derive a key from their seed, they should be prompted for their password (if the wallet is locked)
2. When the user requests to sign a transaction, they should be prompted for their password (if the wallet is locked)
3. The default state of the extension should show public addresses for now (like Polkadot). This provides better UX in Manifest v3 extensions with service workers, as we cannot rely on the state of the service worker to persist any data (such as a password). Any time we need to unlock the keyring (which is currently only when deriving new addresses from the seed, or signing a transaction), this is the point at which the user should be prompted for their password.

For a connected interface, the public account addresses are available even when the extension is not "authenticated" (i.e., when the password is not in state). We can do the same for the default view of the extension, and only authenticate when a stored value is to be decrypted. The service worker should be treated as stateless, as opposed to manifest v2 background pages (which we should not target). Essentially, we don't need to store the password in state.

PR #174, opened 2022-12-22
feat/141 - Prompt for password only when needed

jurevans:
See #141

This updates the extension web app with behavior similar to Polkadot extension, where public addresses are displayed by default, and features needing access to private keys in storage will first prompt for a password (if the keyring storage is locked - the issue here is that we currently require Chrome extension users to constantly authenticate due to Manifest 3 service workers)

- [x] Public addresses display by default (this process is simple in case we want to follow what Keplr does instead)
- [x] If adding an account, check to see if keyring is locked, if so, redirect to login first

 __NOTE__ I've added some preliminary code for handle Tx approvals, but to keep this PR smaller, I'm moving the rest of it to another PR.

*TODO* (separate PR for #200):

- [ ] Submitting a transfer should check if keyring is locked, and if so, prompt user for the password to unlock, prompt user to approve transaction. If keyring is unlocked, simply display the transfer approval form

__NOTE__ For approvals/unlocking from interface, we will need to trigger the opening of the extension web app from within `content-scripts` upon receiving the signing request from the interface, loading the app from the correct route. The password is needed here to decrypt the signing key from storage, and an approval will be needed to resolve the promise in the interface for completing the signing of a transaction.

### DIFF
--- a/apps/extension/src/App/Accounts/AddAccount.tsx
+++ b/apps/extension/src/App/Accounts/AddAccount.tsx
@@ -120,7 +120,7 @@ const AddAccount: React.FC<Props> = ({
         new CheckIsLockedMsg()
       );
       if (isKeyChainLocked) {
-        navigate(TopLevelRoute.Login);
+        navigate(`${TopLevelRoute.Login}?redirect=${TopLevelRoute.AddAccount}`);
       } else {
         setIsLocked(false);
       }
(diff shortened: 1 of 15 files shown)


mateuszjasiuk (approved):
Tested and LGTM! :)

PR #189, opened 2023-01-11
feat: load accounts from metamask

mateuszjasiuk:
Allows you to load accounts from Eth Goerli test network into interface.
It loads accounts into state but It won't show balances as rpc calls are not working.
We should most likely try to unify the signers for all the extensions.


jurevans (approved):
This looks good! We can handle things like balance queries and interface display in other PRs, since more thought needs to go into handling that across all supported extensions.

PR #193, opened 2023-01-16
feat: display message about extenions connections status and

### LINKED ISSUE #191: Integrations - On "Connect", display message to user if an extension isn't installed
When a user loads the interface, they should see a `Connect` button for any chain selected, or, if that extension isn't installed, we can display a message to the user with a link to where it can be installed.

Look in `namada-interface/packages/integrations` at `Keplr.detect()` - For Keplr, this is essentially checking that `window.keplr` is defined. For `Anoma.detect()`, it checks that `window.anoma` is defined.

For Metamask, we can do something similar in `Metamask.detect()`:

```
typeof window.ethereum !== 'undefined'
```

__NOTE__ The chain configs should have a `url` defined where we can link the user for each of the chains (we can use a placeholder URL for `Namada` for now)

mateuszjasiuk:
availability

Resolves #191

- I've changed the integrations so they are not looking for window[extension] in the constructor - because it is undefined most of the times :)
- Added a hook that waits until extension attaches itself so we can start using it
- Made sure that it is impossible to click button during loading phase

### DIFF
--- a/apps/namada-interface/src/App/AccountOverview/AccountOverview.tsx
+++ b/apps/namada-interface/src/App/AccountOverview/AccountOverview.tsx
@@ -26,7 +26,10 @@ import {
   TotalHeading,
 } from "./AccountOverview.components";
 import { formatCurrency } from "@anoma/utils";
-import { useIntegrationConnection } from "services";
+import {
+  useIntegrationConnection,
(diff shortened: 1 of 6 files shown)


jurevans (approved):
Nice! This works very well for me

PR #198, opened 2023-01-20
feat: Sdk integration

mateuszjasiuk:
### General
- [x] Make basic bond tx with SDK
- [x] Check how to store first created account
- [x] Change SubmitBondMsg to take single msg argument
- [x] Fix submitBond return type
- [x] Cleanup extern rpc call
- [x] Cleanup console.logs
- [x] Check cargo check and lint issues
- [ ] ~~Missing traits implementations~~ -> *Move to another issue*
- [ ] Cleanup `shared/Cargo.toml` - after having a tokio fix
- [ ] Fix added TODOs or create new issues

### Masp
- [x] Fix keyring to support shielded accounts creation
- [x] FIx masp errors


packages/shared/lib/src/rpc_client.rs
```diff
@@ -175,4 +74,24 @@ impl Client for HttpClient {
...
+        R: SimpleRequest,
+    {
+        let request_body = request.into_json();
+        let response = self
+            .fetch(&self.url[..], "POST", &request_body)
+            .await
+            .map_err(|e| {
+                let e = stringify(&e).expect("Error to be serializable");
+                let e_str: String = e.into();
+                RpcError::server(e_str)
```
mateuszjasiuk, line 89: Should add comment about server error.

apps/extension/src/background/index.ts
```diff
@@ -12,30 +13,54 @@ import {
...
+
+  const router = new ExtensionRouter(
+    ContentScriptEnv.produceEnv,
+    messenger,
+    extensionStore
+  );
+  router.addGuard(ExtensionGuards.checkOriginIsValid);
+  router.addGuard(ExtensionGuards.checkMessageIsInternal);
+
+  // TODO: Stored data is not encrypted in any way
```
mateuszjasiuk, line 44: Probably because I'm passing None as password :D I think it can be fixed in the next issue.

jurevans (approved):
Looks good to me per our discussion!

memasdeligeorgakis:
Looks good

memasdeligeorgakis (approved):

packages/types/src/signer.ts
```diff
@@ -16,13 +17,12 @@ export interface Signer {
     txData: string
   ): Promise<SignedTx | undefined>;
   encodeBonding(args: BondingProps): Promise<string | undefined>;
+  submitBond(args: SubmitBondProps): Promise<void>;
```
memasdeligeorgakis, line 20: I started to think about this and I am wondering if SubmittingBond is best off here? Like in the future if we keek on adding more stuff like SubmitTransfer, SubmitProposal, and more here, it feels as Signer, do not sound like a most natural place to contain them 🤔

apps/extension/src/provider/Signer.ts
```diff
@@ -76,6 +79,18 @@ export class Signer implements ISigner {
...
   }
 
+  /**
+   * Submit bond transaction
+   */
+  public async submitBond(args: SubmitBondProps): Promise<void> {
+    const msgValue = new SubmitBondMsgValue(args);
+
+    const msg = new Message<SubmitBondMsgValue>();
+    const encoded = msg.encode(SubmitBondMsgSchema, msgValue);
```
memasdeligeorgakis, line 89: We are encoding this here using Borsh schema, but I thought that this should be a responsibility of the SDK?

memasdeligeorgakis: ... this https://github.com/anoma/namada/blob/e40e91a0ae78f33870482f48535a7ae12d94e3f2/shared/src/ledger/tx.rs#L725

packages/types/src/signer.ts
mateuszjasiuk: It was just the easiest thing to do for now. We will still have to think about how Signer looks, because it's different in different extensions.

apps/extension/src/provider/Signer.ts
mateuszjasiuk: Not sure. I think it is done like that because CLI does not need to serialize structs.

apps/namada-interface/src/slices/StakingAndGovernance/actions.ts
```diff
@@ -188,25 +191,21 @@ export const postNewBonding = createAsyncThunk<
...
-    await rpcClient.broadcastTxSync(bytes);
-  } else {
-    throw new Error("Invalid transaction!");
-  }
+  const integration = getIntegration(chainId);
+  const signer = integration.signer() as Signer;
+  signer.submitBond({
+    source: change.owner,
+    validator: change.validatorId,
+    amount: Number(change.amount) * 1_000_000,
```
jurevans, line 199: Not a big deal, but we do have a `amountToMicro` function in `@anoma/utils` that does this conversion

packages/types/src/signer.ts
jurevans: Yeah, exactly. In Keplr, the signing client also has functionality to submit a transaction. Might not make the most sense here, but I think it's fine for now!

apps/namada-interface/src/slices/StakingAndGovernance/actions.ts
mateuszjasiuk: Yeah yeah :D I even used it for transparent tx.

Issue #111, opened 2022-10-07
Namada-Interface & Extension - Share Components


jurevans:
The extension and the interface should share components from `@anoma/components` to avoid duplicating any code.

Also, we now have the `@anoma/utils`, which contains `theme`, for both `namada-interface` and `extension` to make use of.

We need to port the latest `namada-interface/src/utils/theme` over to `@anoma/utils`, and update the imports across `namada-interface`.

This could be done at the same time as moving all shared components to `@anoma/components`.

As a part of this process, we will need to consolidate the approach to the build process, moving `namada-interface` to a pure Webpack configuration to allow for greater customization, and to keep inline with the extension build.

PR #202, opened 2023-01-31
feat/111 - Consolidate build process, share components

jurevans:
See #111 

This PR is to consolidate the build process between the `extension` and `namada-interface`, and set up in a way where all components in `packages/components` can be shared across both.

- [x] Set up `webpack` for `namada-interface`
- [x] Update component imports in `namada-interface`
- [x] Move any remaining components into `packages/components` where it makes sense to
  - Everything in `namada-interface/src/components` should instead be in `packages/components`
- [x] Ensure no runtime errors are present
- [x] Set up development server (`webpack-dev-server`) on port 3000
  - **Note** this handles hot module replacement on file change

The goal is that development is identical to before, only we have shared components, and much easier way to configure `namada-interface` for our needs.

DEMO: https://pull-request-202--wallet-development-heliax-dev.netlify.app/

_NOTE_ I couldn't move the `Toast` component to the shared `packages/components`, as it has dependencies on the RTK functionality. For now, I've put it under `App` along with files having similar dependencies.

_NOTE_ I've updated `apps/namada-interface/.env.sample` as we no longer need to specify `GENERATE_SOURCEMAPS=false`, as this is now handled in `webpack`, so we don't need to do any hacks around `create-react-app` issues, which is nice.

_NOTE_ I ran into issues in CI due to a mismatch in the `wasm-bindgen` version, so this has been updated from `0.2.83` to `0.2.84` in the `crypto` and `shared` `Cargo.toml` and `Cargo.lock` files.


mateuszjasiuk (approved):
Scrolled through it :) Looks good!

memasdeligeorgakis (approved):

PR #204, opened 2023-02-09
feat: use sdk for transparent transfer tx

mateuszjasiuk:


apps/extension/babel.config.json
```diff
@@ -0,0 +1,4 @@
+{
+  "presets": [],
+  "plugins": ["@babel/plugin-transform-modules-commonjs"]
```
mateuszjasiuk, line 3: Added this one because of the javascript files inside shared rust lib. When we test we set wasm target to node and those js files can't be imported correctly unless we transform them to CommonJs

apps/namada-interface/src/slices/transfers.ts
```diff
@@ -128,207 +122,6 @@ type TxIbcTransferArgs = TxArgs & {
   portId: string;
 };
 
-/* type ShieldedAddress = string; */
```
mateuszjasiuk, line 131: Just removed bunch of unused lines.

packages/shared/lib/src/sdk.rs
```diff
@@ -22,28 +24,21 @@ use crate::rpc_client::HttpClient;
 
 const STORAGE_PATH: &str = "";
 
-pub struct WebWallet {}
+#[derive(Default, Debug, BorshSerialize, BorshDeserialize, Clone)]
+pub struct WebShieldedUtils {}
 
-impl WalletUtils for WebWallet {
```
mateuszjasiuk, line 27: Because of Murisi's change we could get rid off the WebWallet utils :tada:

packages/types/src/tx/schema/bond.ts
```diff
@@ -1,35 +1,7 @@
 import BN from "bn.js";
 import { Schema } from "borsh";
-import { SubmitBondProps, Tx } from "../types";
-
-//TODO: move somewhere else
-export class TxMsgValue {
```
mateuszjasiuk, line 6: I've moved tx specifix schemas to the separate file as it's reusable.

Issue #205, opened 2023-02-13
Interface: IBC - Cosmos ecosystem to Namada integration


jurevans:
We need to support IBC transfers from Keplr chains (e.g., Cosmos, Osmosis, etc.) to Namada.

Currently, the interface has only tested submitting IBC transfers when the source chain is Namada. We will need to provide functionality in the Bridge page to submit a transfer _to_ a Cosmos chain using the relevant _source_ account and signer, with Namada as a target. We will need to access accounts and signing functionality in Keplr, and submit over RPC an IBC Transfer message using `cosmjs`.

- User should be able to submit IBC transfer to Cosmos with Namada as a destination chain
- User of interface should be ableKeplr chains to load Keplr accounts with which to transfer from
- IBC transfers can be signed by account in Keplr

It may make sense to extend the `Integration` class so that each chain instance can have its own internal, extension-specific implementation of a common API.

PR #209, opened 2023-02-14
Keplr Integration - Display All Accounts

jurevans:
This PR addresses parts of #205 

In this PR, we now show all accounts from the extension, even if it has a zero balance, and instead of listing by token balance, it lists all accounts, and the user can choose to view the breakdown of token balances underneath their account. An issue with loading accounts from Keplr has been resolved.

*NOTE* I've also patched up a couple small things in the `webpack.config` for `namada-interface`

- [x] Can view accounts loaded from Keplr extension
- [x] Can choose to submit IBC transfer to a chain in Cosmos ecosystem

TODO: Follow-up PR:
- [ ] Can sign IBC transfer with Keplr account

This PR introduces a `submitBridgeTransfer` function to `Integration` based classes, which can accept either `IbcTransferProps` or `BridgeTransferProps` (in the case of Metamask, which will use the Ethereum bridge). This will be implemented in a future PR.

### DIFF
--- a/apps/namada-interface/src/App/AccountOverview/DerivedAccounts/DerivedAccounts.components.ts
+++ b/apps/namada-interface/src/App/AccountOverview/DerivedAccounts/DerivedAccounts.components.ts
@@ -197,7 +197,7 @@ export const TokenBalance = styled.li`
   width: 100%;
   list-style: none;
   list-style-type: none;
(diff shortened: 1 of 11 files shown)


mateuszjasiuk (approved):
Tested! Looks good.

PR #225, opened 2023-03-07
Balance Query & IBC with Keplr Accounts (@cosmjs)

### LINKED ISSUE #205: Interface: IBC - Cosmos ecosystem to Namada integration

jurevans:
Resolves #205 

This PR extends our `Integration` functionality to abstract the call for querying balances, and implements `submitBridgeTransfer` for Cosmos.

- [ ] Submit IBC transfer (Cosmos->Namada)
  - _TODO_ - We need the working environments to test this 
- [x] Query balance via CosmJS for Keplr account
- [x] Provide Keplr accounts in IBC Bridge interface

## TODO
- Implement Gas price setting in Bridge page, based on `gasPriceStep` in chain configs

## Screenshots

namada-bridge-page

## NOTES

This is most easily tested using Keplr as installed from the web extension store, which means it is pointing to Cosmos mainnet. For this to work with the interface, the `.env` file in `namada-interface` shouldn't have any overrides, and should use the default `Cosmos` chain config instead, which points to mainnet. If this doesn't match what Keplr is using, we will get no accounts from Keplr.

### Local Gaia->Namada IBC set-up

See https://github.com/heliaxdev/namada-ibc-docker

As we are currently dependent on the SDK feature branch (which is based on Namada `0.12.0`), we'll need to set the Namada version appropriately in `namada/Dockerfile`.

### Keplr-Wallet development build setup

First, update the configuration in []() to point to Gaia RPC (or Cosmos Hub testnet)

_NOTE_ You will need `protoc` installed (CLI for compiling protobuf)

_NOTE_ Set-up for Gaia testnet _or_ local Gaia testnet should be relatively the same. I am testing this, and will update these instructions based on what I find.

- Clone: [keplr-wallet](https://github.com/chainapsis/keplr-wallet/) repo
- In `packages/extension/src`, you will need two config files present:
  - `config.ui.var.ts`
  - `config.var.ts`
   _See the example files to generate these with your own settings:_
  - `config.ui.var.example.ts`
  - `config.ui.example.ts`

What we need is a `config.var.ts` that looks like this:
```ts=
import { AxiosRequestConfig } from "axios";

export const COSMOS_RPC_ENDPOINT =
  "https://rpc.sentry-01.theta-testnet.polypore.xyz:443";
export const COSMOS_RPC_CONFIG: AxiosRequestConfig | undefined = undefined;
export const COSMOS_REST_ENDPOINT =
  "https://rest.sentry-01.theta-testnet.polypore.xyz:443";
export const COSMOS_REST_CONFIG: AxiosRequestConfig | undefined = undefined;
```

- Run `export NODE_OPTIONS=--openssl-legacy-provider`, then:
- `yarn build`

The extension files which you will need to point your browser to are located at:
```
extension/prod
```

Then, in `namada-interface/apps/namada-interface/.env`, update the file to match this:

```
REACT_APP_COSMOS_ALIAS=Cosmos Testnet
# Following is optional (just use default), but this needs to match what is specified on Keplr!
REACT_APP_COSMOS_CHAIN_ID=cosmoshub-4
# The following can be a local IP as well (for local testing):
REACT_APP_COSMOS_CHAIN_URL=https://rpc.sentry-01.theta-testnet.polypore.xyz
```

_NOTE_ Even though the testnet's chain ID is `theta-testnet-1`, which should still specify (or omit and use the default) of `cosmoshub-4` so that we can pull accounts from the Keplr extension. Otherwise, we would need to build Keplr with the testnet chain ID updated in `packages/src/extension/src/config.ts`, which is more tedious.

_NOTE_ We can query the Cosmos testnet to ensure our endpoint is up via:

```
curl https://rpc.sentry-01.theta-testnet.polypore.xyz/status | jq
```

Optional node endpoint:

```
https://rpc.earth.theta-testnet.polypore.xyz
```


mateuszjasiuk (approved):
LGTM after reviewing it together

PR #227, opened 2023-03-09
fix: Discard persisted ChainId when it is not part of defined chains

mateuszjasiuk:


jurevans (approved):
Works perfectly for me!

PR #229, opened 2023-03-13
bug: conversion rates do not update for specific accounts

### LINKED ISSUE #224: BUG: Conversion rates not updating for individual tokens
A bug was introduced recently that removes the balance in fiat currency for individual tokens (the total balance in fiat is still converted), resulting in `$0.00` for the individual tokens.

This should be a straight-forward fix, as this is something that previously worked.

mateuszjasiuk:
Resolves #224


jurevans (approved):
LGTM!

PR #233, opened 2023-03-15
fix: send token view does not display ui after refresh

mateuszjasiuk:
Part of the UI does not display for send token view after refreshing. 
This fixed it for scenario when we are connected to extension and we can query node for balances.


jurevans (approved):
LGTM!

PR #235, opened 2023-03-16
feat: add zxcvbn validation

mateuszjasiuk:
Based on the Audit report:
Issue D: Usage of Strong Passwords Not Enforced

Added `zxcvbn` validation which allows passwords with strength of 3 or more, which usually means password at least 11 characters long. It also checks for other properties like most common passwords and usage of recursive words.

Validation is disabled for `NODE_ENV=development`, although feedback messages are still present.


jurevans (approved):
This looks good!

PR #238, opened 2023-03-22
feat: non blocking masp transfers in chrome

mateuszjasiuk:
- [x] Send shielded transfer with WebWorker in Chrome
- [x] Send shielded transfer with WebWorker in FF
- [x] Keep service worker alive for the time transfers are going on
- [x] Clear offscreen document when not needed
- [x] Clear WebWorkers when not needed
- [x] Send message to the extension/interface with tx done info
- [ ] ~~Make sure that it works for MASP only~~ - Moved to the new issue
- [ ] ~~Fix transfer completed toasts~~ - Moved to the new issue

### DIFF
--- a/apps/extension/src/background/web-workers/submit-transfer-web-worker.ts
+++ b/apps/extension/src/background/web-workers/submit-transfer-web-worker.ts
@@ -2,9 +2,13 @@ import { Sdk } from "@anoma/shared";
 import { init as initShared } from "@anoma/shared/src/init";
 import { IndexedDBKVStore } from "@anoma/storage";
 import { fromBase64 } from "@cosmjs/encoding";
-import { SDK_KEY } from "background/keyring";
+// import { SDK_KEY } from "background/keyring";
 import { KVPrefix } from "router";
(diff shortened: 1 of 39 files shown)


apps/extension/src/background/web-workers/submit-transfer-web-worker.ts
```diff
@@ -0,0 +1,37 @@
+import { Sdk } from "@anoma/shared";
+import { init as initShared } from "@anoma/shared/src/init";
+import { IndexedDBKVStore } from "@anoma/storage";
+import { fromBase64 } from "@cosmjs/encoding";
+import { SDK_KEY } from "background/keyring";
+import { KVPrefix } from "router";
+import { INIT_MSG } from ".";
+
+const DEFAULT_URL =
```
jurevans, line 9: Minor thing, but we could simply pull the URL as specified in the `namada` chain config in `@anoma/chains`, which will handle any overrides in the `.env` as well

jurevans (approved):
Looks good to me!

PR #246, opened 2023-03-29
feat: pass routerId directly to the requester

mateuszjasiuk:
To make it a little bit more straightforward,

### DIFF
--- a/apps/extension/src/App/Requester.tsx
+++ b/apps/extension/src/App/Requester.tsx
@@ -1,0 +1,52 @@
+import browser from "webextension-polyfill";
+import { ExtensionKVStore } from "@anoma/storage";
+import {
+  ExtensionMessenger,
+  ExtensionRequester,
+  getAnomaRouterId,
+} from "extension";
+import { createContext, useContext, useEffect, useState } from "react";
+import { KVPrefix } from "router";
+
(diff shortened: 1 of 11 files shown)


apps/extension/src/App/Requester.tsx
```diff
@@ -0,0 +1,52 @@
+import browser from "webextension-polyfill";
+import { ExtensionKVStore } from "@anoma/storage";
+import {
+  ExtensionMessenger,
+  ExtensionRequester,
+  getAnomaRouterId,
+} from "extension";
+import { createContext, useContext, useEffect, useState } from "react";
```
jurevans, line 8: Not a big deal at all, but generally I try to keep third-party imports separate from local (e.g., `react`, etc.) - at least I've tried to do that and be relatively consistent. Do you have an opinion on that, or possibly think it isn't useful?

jurevans (approved):
Nice, looks good!

apps/extension/src/App/Requester.tsx
mateuszjasiuk: So by separate you mean to group them and separate by empty line right?
Yeah I would do it too. I think I just got lazy with autoimports :D Will fix!

PR #248, opened 2023-04-04
feat: keep track of the transfer using msgId

mateuszjasiuk:
Resolves [#251](https://github.com/anoma/namada-interface/issues/251)

- Added TransferStartedEvent - sent by the background script, that informs content script that stuff is happening :)- 
Added custom "anoma_transfer_completed" event, which is being send from content script to the interface via `window.dispatchEvent`
- Added service which listens to the window events and handles them accordingly. In this case by displaying correct toasts.


jurevans (approved):
Nothing to add, looks good to me!

PR #257, opened 2023-04-14
feat: zeroize secrets in WASM memory

### LINKED ISSUE #243: Investigate memory zeroization in the JS

To make sure we are overriding memory that contains secret information.

emccorson:
This PR adds more careful handling of sensitive data to try to guard against sensitive data being left in memory.

The strategy is:

1. Wrap sensitive values in structs.
a. Wrapped values will be automatically zeroized in memory when dropped.
b. Wrapped values can be passed to JS without loading the sensitive value into JS memory.
c. If required, the value can be read from JS by reading directly from WASM memory, which gets around wasm-bindgen making a temporary value in memory that we can't zeroize.
2. Local variables in Rust are zeroized after use if they hold sensitive data.
3. Use `#[wasm_bindgen(skip_typescript)]` to hide methods from JS that don't wrap sensitive data. (Ideally I would use `#[wasm_bindgen(skip)]` but it [doesn't currently support](https://github.com/rustwasm/wasm-bindgen/pull/1410#issuecomment-1517584716) skipping methods.)

I've added some comments for the review at parts I wasn't sure about.

Fixes #243.

### DIFF
--- a/packages/crypto/lib/src/crypto/aes.rs
+++ b/packages/crypto/lib/src/crypto/aes.rs
@@ -4,6 +4,8 @@ use aes_gcm::{
 };
 use thiserror::Error;
 use wasm_bindgen::prelude::*;
+use zeroize::Zeroize;
+use crate::crypto::pointer_types::VecU8Pointer;
 
 #[derive(Debug, Error)]
 pub enum AESError {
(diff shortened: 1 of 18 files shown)


packages/crypto/lib/src/crypto/aes.rs
```diff
@@ -22,35 +24,41 @@ pub struct AES {
...
             cipher: Aes256Gcm::new(&key),
             iv,
-        })
+        };
+        key.zeroize();
+        Ok(aes)
     }
 
-    pub fn encrypt(&self, text: &str) -> Result<Vec<u8>, String> {
+    pub fn encrypt(&self, mut text: String) -> Result<Vec<u8>, String> {
```
emccorson, line 48: Not sure about the change from `&str -> String` in the `text` parameter here, but I think it's right because we want to zeroize the text after we've received it so the method should gobble up the string. wasm-bindgen seems to handle both the same anyway.

packages/crypto/src/__tests__/crypto.test.ts
```diff
@@ -8,12 +8,21 @@ import {
...
+  readVecU8Pointer
+} from "../utils";
+
+// __wasm is not exported in crypto.d.ts so need to use require instead of import
+const memory = (require("../crypto/crypto")).__wasm.memory;
+
 const KEY_LENGTH = 32;
 const SEED_LENGTH = 64;
-const SHIELDED_KEY_LENGTH = 96;
+const SHIELDED_CHILD_KEY_LENGTH = 169;
```
emccorson, line 25: I don't know where this number comes from and if it is stable to use, but I just took it from the actual output so the tests pass.

mateuszjasiuk: Had no idea, but it seems to be correct :)
Summing all the bytes in:
```
pub struct ExtendedFullViewingKey {
    depth: u8,
    parent_fvk_tag: FvkTag,
    child_index: ChildIndex,
    chain_code: ChainCode,
    pub fvk: FullViewingKey,
    dk: DiversifierKey,
}
```
Returns 169

packages/crypto/lib/src/crypto/bip39.rs
```diff
@@ -44,21 +47,16 @@ impl Mnemonic {
     }
 
     pub fn from_phrase(phrase: String) -> Result<Mnemonic, String> {
-        Mnemonic::validate(&phrase)?;
-
-        Ok(Mnemonic {
-            phrase
-        })
-    }
+        if let Err(e) = Mnemonic::validate(&phrase) {
```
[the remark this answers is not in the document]
emccorson: Oops, there should be a call to zeroize in there. I have fixed it now.

jurevans (approved):
Nice! LGTM. Only possible thing to think about is adding more comments in our code-base, but that is something we need to do universally, and is not specific to this PR

PR #258, opened 2023-04-14
WIP: feat/254 - Support multiple top-level accounts

### LINKED ISSUE #254: Extension: Enable multiple top-level accounts
__NOTE__ This blocks #253 

In the extension, currently, we can only create one mnemonic seed. The storage has been set up to allow for any number of mnemonics, so we need to give the user the option to generate and switch to another account. In order to give the user the option to link the extension with the Ledger device, this issue needs to first be implemented.

- [ ] User can elect to create a new (top-level) account from within the extension
- [ ] For "create account", the user should see:
  - `Create Account` (generate a new mnemonic seed)
  - `Connect Ledger` (see #253)
  - `Import Seed` (or something similar to restore from mnemonic phrase - disabled for now)
- [ ] Clicking `Create Account`, user can generate a new mnemonic phrase, and is given an initial account (same flow as before)
- [ ] Clicking `Connect Ledger` takes the user to the Ledger hw wallet flow (to be implemented in #253)
- [ ] Within the extension, we need an option to view accounts with the ability to switch accounts, as well as `Create Account`

## Notes

The storage will need to be slightly modified such that for every Mnemonic the user generates, the initial path will be set (e.g., `/0'/0/0`) for the first account the user derives from that seed, which means we can't use that `account` index to identify the current active parent-account (as they will all start from `0`). This may be easily solved by generating a unique ID for each top-level mnemonic stored, which the child accounts can point to. Changing the active account will set an ID in storage, and load all related accounts, _or_ load all accounts from the connected Ledger wallet.

#### Create Account

keplr-create-new-account

#### Select Account

keplr-select-account

### Metamask

metamask-create-account

jurevans:
Resolves #254 

In order to support a user flow where a user may connect and utilize accounts from their Ledger HW wallet, we need to _first_ support accessing multiple accounts from the extension. This has been previously implemented to a degree, where we can technically store any number of mnemonic seeds and the associated accounts, but this needs to be completed in this issue.

### DIFF
--- a/apps/extension/src/App/Accounts/AddAccount.tsx
+++ b/apps/extension/src/App/Accounts/AddAccount.tsx
@@ -33,8 +33,7 @@ import { TopLevelRoute } from "App/types";
 import { useAuth } from "hooks";
 
 type Props = {
-  // The parent Bip44 "account"
-  parentAccount: number;
+  accountIndex: number;
   accounts: DerivedAccount[];
   requester: ExtensionRequester;
   setAccounts: (accounts: DerivedAccount[]) => void;
@@ -43,12 +42,12 @@ type Props = {
 };
 
 const validatePath = (
-  account: number,
+  accountIndex: number,
   newAccount: { change: number; index: number },
   accounts: DerivedAccount[],
   accountType: AccountType
 ): boolean => {
-  const newPath = [account, newAccount.change, newAccount.index].join("/");
+  const newPath = [accountIndex, newAccount.change, newAccount.index].join("/");
   let isValid = true;
   accounts
     .filter((derivedAccount) => derivedAccount.type === accountType)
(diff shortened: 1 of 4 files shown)


mateuszjasiuk (commented):
Looks good!
One thing did not work for me though. After loading new extension and after creating new account I could not see addresses. I think it happens because check for active account happens only once in `keyring` in the `_init` function. So it starts working after reloading the extension.

Do you think it makes sense to store `activeAccountId` right after we store mnemonic on account creation? I think this way we do not have to fall back on the mnemonic to get the id in the `_init`.

jurevans:
> Looks good! One thing did not work for me though. After loading new extension and after creating new account I could not see addresses. I think it happens because check for active account happens only once in `keyring` in the `_init` function. So it starts working after reloading the extension.
> 
> Do you think it makes sense to store `activeAccountId` right after we store mnemonic on account creation? I think this way we do not have to fall back on the mnemonic to get the id in the `_init`.

Good catch! I will take a look, and that is probably the best thing to do. Thanks!

mateuszjasiuk (approved):
LGTM!

mateuszjasiuk (approved):
Looks good! :)

PR #259, opened 2023-04-19
feat/254 - Events for extension account changes

### LINKED ISSUE #254: Extension: Enable multiple top-level accounts

jurevans:
Resolves #254 

- [x] Extension dispatches events when the top-level account has changed
- [x] Interface responds to this by reloading accounts

__NOTE__ I've added placeholder handlers for Keplr & Metamask as these will also need to respond to the relevant extension events - We'll need a follow up issue to address the proper way to subscribe to account changes in these extensions as we are with our own.

__NOTE__ This also only works if the interface is the current `active` tab. Another follow up task is determine best way to broadcast events (such as this, where no sensitive data is being transmitted) across all tabs for any interface that wishes to subscribe to extension events. This could also be handled better (and more precisely) by storing the tab ID of any connected interface within the extension, and broadcasting events _only_ to that tab. See #262


mateuszjasiuk (approved):
Looks good!

PR #262, opened 2023-05-02
feat/261 - Extension events improvements

### LINKED ISSUE #261: Extension<>Interface events improvements
Following #259, it would be a substantial improvement to incorporate the following:

- [ ] Extension tracks the `tabId` of any interface connected to it
- [ ] Events from extension which need to trigger updates on any connected interface should be broadcasted to each of those `tabId`s

An example of this may be, like in #259, the user selects a different top-level (parent) account in the extension, and any connected interface will automatically update to be in sync with the extension so that the transaction submitting flows work correctly. In this case, only the active tab will update (currently), but knowing all connected tabs by ID will allow us to update any connected interface without needing to broadcast globally.

jurevans:
Resolves #261 

- [x] Tracks connected interfaces by `senderTabId` in extension storage
- [x] Changing top-level account broadcasts to all connected interfaces in tab ID store

__NOTE__ I created a type to track `tabId` along with a timestamp of when it first connected, so we can choose not to store these indefinitely. Currently, these do not expire. I first tried this with an in-memory store, but as soon as the service worker goes inactive, the tabs are lost, and it's a bit tedious to require the interface to reconnect each time in order to receive events.

__NOTE__ In this PR, I needed to address one issue: The Integration `Anoma.connect()` is _not_ invoked if the interface automatically connects to the extension (if the user has connected once, this process is automated, but should still invoke `.connect()`, as it could be from a different tab perhaps. If the user has already approved a connection to the extension, this could happen automatically, but still needs to happen. I added a call to `.connect()`, within `.accounts()`, as it needs to occur first in order to keep track of tab IDs. I'll take a look to see if there is any more efficient way of handling this, but I think it's safe to assume that if the consumer wants to query `accounts`, then they also intend to `connect` to the extension (one day there will be an approval process for the initial connection). I believe this is similar to how Keplr handles this with their `Keplr.enable(chainId)` call.

### Testing

- Load the extension: `yarn start:chrome`
- Create multiple top-level accounts (Settings Icon -> `Add Account`)
- Launch the `namada-interface` dev server: `yarn dev:local`
- Launch multiple tab instances of `localhost:3000` (click `Connect to Extension` if prompted)
- In extension, click settings icon, then choose a different accounts - this should cause each open `namada-interface` tab to update
- You can test persistence by restarting the extension, and then switching the top-level account. The results should be the same.

### TODO

We should possibly look into when a tab _should_ expire. We could take any expired tab IDs and broadcast a message to that tab, if it responds correctly we could reset the timestamp, or remove the tab ID from storage. I feel like this storage should not grow indefinitely. It may make sense to ping the tab to ensure it's still alive, and remove any reference to it if not. It would be easier to just eject tab IDs after a set period, then broadcast a final message indicating that it needs to reconnect to the extension.


mateuszjasiuk (approved):
Tested and LGTM!
It did not work for me in the beginning because my anoma chain id is:
```
""
```
And for `chainId` check we cast string to boolean. Probably not a big issue as it is hard to imagine that `chainId` will be an empty string in real life scenario :D

PR #266, opened 2023-05-05
refactor/265 - Move React hooks to package

### LINKED ISSUE #265: Refactor: Move hooks into shared public package
It would be useful to move all hooks to a dedicated, public package within our repo.

Currently, we have the following hooks, and will likely be creating more:

- `useUntil`
- `useIntegration`
- `useIntegrationConnection`
- `useEvent`

In some instances, the extension may make use of hooks that currently exist only on `namada-interface`, but in general, it would help third-party apps wishing to integrate with our code-base to have the ability to re-use our hooks from the public repo packages. If nothing else, it will be easier to find them to import via `@anoma/hooks`.

For the `provider`s specific to `namada-interface`, I think it makes sense to keep these in place for now (e.g., `extensionEvents/provider` and `integrations/provider`), but we can always investigate any benefits to making these common as well.

jurevans:
Resolves #265 

This PR is simply to move any hooks to a shared package within our repo (`@anoma/hooks`)


emccorson (approved):
Looks good!

mateuszjasiuk (approved):
Looks good and makes sense :)

PR #270, opened 2023-05-09
bug/263 - Fix for extension popup error

### LINKED ISSUE #263: Bug: Extension pop up throws error after service worker goes inactive
Any time the service worker in the extension goes inactive, opening the extension popup will display the `Setup` button, as it has thrown an error when attempting to load accounts from the service worker. The service worker at this point is still in the process of starting up, and isn't ready by the time the popup issues that request.

To replicate: 
- With the extension loaded and set-up, wait for the status of the Service Worker to go `inactive`, then click to open the extension popup.
- Note that when attempting to open the popup again, it will display accounts properly

We need some kind of way for the popup to know that the service worker is active before attempting to issue requests to it. This could be an event dispatched after the service worker initializes, that the popup app listens for, or it could be similar to our `useUntil` hook in `namada-interface`. The former may be quicker and simpler to implement, as the latter would need that hook moved to a public package for import, and used in the extension to make attempts until successful.

__NOTE__ to discover this as the issue, in the `App.tsx` of the extension popup, I added a `setTimeout` around the initial `fetchAccounts` function, and it opens correctly every time if delayed (at least in my test) by `300` ms.

jurevans:
Resolve #263 

This PR makes use of `useUntil`, where the predicate will return `true` when the requester is able to resolve a "fetch accounts" `sendMessage` promise to the extension, and once successful, will load those accounts into state.

__NOTE__ This method will cause many errors to be logged, and I have some ideas on how to address that in #271. The idea is to introduce better and more useful error handling within the messaging of the extension. Currently, we throw an exception in any case that the requester is unable to resolve a promise, and it would perhaps be much more useful to reject the promise and return data to whichever method requested it for better handling on the client end.

### DIFF
--- a/apps/extension/src/App/App.tsx
+++ b/apps/extension/src/App/App.tsx
@@ -7,11 +7,12 @@ import { DerivedAccount } from "@anoma/types";
 import { getTheme } from "@anoma/utils";
 import { ExtensionKVStore } from "@anoma/storage";
 import { Icon, IconName } from "@anoma/components";
+import { useUntil } from "@anoma/hooks";
 
 import { ExtensionMessenger, ExtensionRequester } from "extension";
 import { KVPrefix, Ports } from "router";
 import { QueryAccountsMsg } from "provider/messages";
-import { GetActiveAccountMsg } from "background/keyring";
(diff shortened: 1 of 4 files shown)


mateuszjasiuk (approved):
Looks good to me :)

PR #272, opened 2023-05-09
Extension: Update tab storage to account for closed tabs

### LINKED ISSUE #269: Extension - tab tracking should forget tabs when they are no longer available
When the user closes a tab, the extension should more gracefully handle how it broadcasts events. If the tab is no longer available, storage should be updated before broadcasting so we don't pollute the console with errors.

- Extension should query tabs and compare to what is in storage, and remove any unavailable tabs before broadcasting
- Note that this may negate the need to track timestamps for tab connections, unless we want to expire a "session" after a set period

Currently, this will throw an error every single time events are broadcast as it will keep attempting to communicate to tabs that no longer exist.

jurevans:
Resolves #269 

This PR adds a utility to synchronize the tabs tracked within the `KeyRingService`. It compares what is stored against what is available via `browser.tabs.query`, and updates the storage, returning the updated _current_ tabs before attempting to broadcast to them. This prevents errors that occur in the requester when an attempt is made to communicate with a non-existent tab.

### DIFF
--- a/apps/extension/src/background/keyring/utils.ts
+++ b/apps/extension/src/background/keyring/utils.ts
@@ -18,15 +18,11 @@ export const syncTabs = async (
   chainId: string
 ): Promise<TabStore[]> => {
   const tabs = await requester.queryBrowserTabIds();
-  const storedTabs = await connectedTabsStore.get(chainId);
+  const storedTabs = (await connectedTabsStore.get(chainId)) || [];
   const currentTabs: TabStore[] = [];
 
-  if (storedTabs?.length === 0) {
-    return [];
-  }
-
   storedTabs?.forEach((tab: TabStore) => {
-    if (tabs.indexOf(tab.tabId) > -1) {
+    if (tabs.includes(tab.tabId)) {
       currentTabs.push(tab);
     }
   });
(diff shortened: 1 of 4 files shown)


apps/extension/src/background/keyring/utils.ts
```diff
@@ -0,0 +1,67 @@
...
+  const tabs = await requester.queryBrowserTabIds();
+  const storedTabs = await connectedTabsStore.get(chainId);
+  const currentTabs: TabStore[] = [];
+
+  if (storedTabs?.length === 0) {
+    return [];
+  }
+
+  storedTabs?.forEach((tab: TabStore) => {
+    if (tabs.indexOf(tab.tabId) > -1) {
```
mateuszjasiuk, line 29: You can also use `includes` . As a side not it is quite funny how long it took to add it to JS  :D

mateuszjasiuk (approved):
LGTM! Added two small comments.

apps/extension/src/background/keyring/utils.ts
jurevans: ah yes, thanks for the reminder!

PR #273, opened 2023-05-11
chore: fix wasm-bindgen-cli version to v0.2.84 in CI

emccorson:
When v0.2.85 was released, it broke the CI build because wasm-bindgen-cli versions are required to match, so this commit fixes the globally installed version to v0.2.84 as a temporary measure until we upgrade our packages to use v0.2.85.


jurevans (approved):
LGTM!

PR #275, opened 2023-05-12
feat: update to namada 16

mateuszjasiuk:
Make sure to:
- install protobuf-compiler
- copy namada wasms to correct path `$HOME/.local/share/namada`
- set correct chainId in envs and recreate extension accounts

Looks like transfer fee ATM is 100NAM. Take this into consideration when transfering :)


jurevans (approved):
This looks good!

PR #276, opened 2023-05-12
docs: initial sdk docs

### LINKED ISSUE #178: Document Namada SDK
Subtask of  #177

In order to know what to develop in the client it should be defined what are the types and methods available to the consumer of the SDK.

Outcomes:
* simple documentation of what the SDK covers

mateuszjasiuk:
Resolves #178 

Initial docs for the Sdk.

Let me know if some things are missing or maybe something can be phrased better :)

### DIFF
--- a/packages/shared/lib/src/sdk/mod.rs
+++ b/packages/shared/lib/src/sdk/mod.rs
@@ -10,6 +10,7 @@ mod masp;
 mod tx;
 mod wallet;
 
+/// Represents the Sdk public API.
 #[wasm_bindgen]
 pub struct Sdk {
     client: HttpClient,
@@ -18,6 +19,8 @@ pub struct Sdk {
 }
 
 #[wasm_bindgen]
+/// Sdk mostly wraps the logic of the Sdk struct members, making it a part of public API.
+/// For more details, navigate to the corresponding modules.
 impl Sdk {
     #[wasm_bindgen(constructor)]
     pub fn new(url: String) -> Self {
--- a/specs/packages/sdk.md
+++ b/specs/packages/sdk.md
@@ -1,0 +1,46 @@
+# Sdk
+
+Sdk is a part of shared package. It's main purpose is to wrap shared logic from the namada crate, so it can be used in the context of the browser.
+
+Modules:
+
+- [masp](#masp)
+- [tx](#tx)
+- [wallet](#wallet)
+- [sdk](#sdk)
+
+## masp
+
+Contains implementation of the `ShieldedUtils` trait. It is responsible for:
+
+- loading and saving spend, convert and output params into and from the browser storage.
+- serializing and deserializing params for read/write
+- creating LocalTxProver needed for creating MASP proof
+
+## tx
+
+Encapsulates functionality for deserialization and mapping of byte arrays into Rust structures.
+Contains:
+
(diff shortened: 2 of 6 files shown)


packages/shared/lib/src/sdk/mod.rs
```diff
@@ -18,6 +19,8 @@ pub struct Sdk {
 }
 
 #[wasm_bindgen]
+/// Sdk mostly wraps the logic of the Sdk struct members, making it a part of public API.
+/// For specifics naviagate to the corresponding modules.
```
jurevans, line 23: Minor typo: `naviagate` -> `navigate`

Could maybe be a little more clear with a comma and adjustment like following:
```
/// For more details, navigate to the corresponding modules.
```

specs/packages/sdk.md
```diff
@@ -0,0 +1,45 @@
...
+## wallet
+
+Contains the logic for:
+
+- encoding/decoding wallet store
+- adding keys/spending kest
+
+### Why we are not using key generation functions from the Namada Shared?
+
+At the moment Anoma extension creates keys based on mnemonic, while Namada uses OsRng as a source of randomness. Because of that we have to introduce functions that allow to add keys generated(derived) inside JS codebase into wallet storage. This will most likely change in future after Namada introduces mnemonic support.
```
jurevans, line 37: I was going to suggest that we link here to our current "Key Managment" docs, but they are not particularly informative: https://github.com/anoma/namada-interface/blob/main/specs/browser-extension/key-management.md - so, I think this is fine. I would maybe slightly re-word it to something like the following (if I understand this part correctly):

```
Currently, the Anoma extension derives keys based on a mnemonic, while Namada derives keys based on a source of randomness (`OsRng`). As such, we have introduced functions that allow us to add keys derived within the extension into Sdk's wallet storage. This will most likely change in future after Namada introduces support for mnemonics into the Sdk.
```

Something like that?

specs/packages/sdk.md
```diff
@@ -0,0 +1,45 @@
...
+
+## sdk
+
+Sdk struct wraps logic from the other modules, creating public API using `wasm_bindgen`.
+
+## About wallet storage
+
+- At the moment client of the Sdk is responsible for keeping storage in sync. What does that mean? Anoma extension has its own key storage, which is modified by user actions. Every time extension storage changes(for example by deriving new address), we have to manually sync storages by using functions provided by `wallet` module. This is something that will most likely change in the future.
```
jurevans, line 45: Slight re-wording:

```
At the moment, the client of the Sdk is responsible for keeping storage in sync with the Sdk's wallet instance.
What does that mean? The Anoma extension has its own key storage, which is modified by user actions.
Every time  the extension storage changes (for example, when a new address is derived), we have to manually
sync the extension and Sdk wallet storage by using functions provided by `wallet` module. This is something
that will most likely change in the future.
```

jurevans (approved):
This looks good to me! I've added just a few minor comments/suggestions.

mateuszjasiuk:
> This looks good to me! I've added just a few minor comments/suggestions.

@jurevans  Thanks! Made all of the suggested changes :) I've re-requested the review.

jurevans (approved):
LGTM!

PR #277, opened 2023-05-17
feat: Allow user to change account password

### LINKED ISSUE #241: Figure out password resetting

Do some research on password resetting.
Check current solutions.

emccorson:
Fixes #241.

### DIFF
--- a/apps/extension/src/background/keyring/keyring.ts
+++ b/apps/extension/src/background/keyring/keyring.ts
@@ -25,12 +25,13 @@ import {
 } from "@anoma/types";
 import { chains } from "@anoma/chains";
 import { Crypto } from "./crypto";
-import { KeyRingStatus, KeyStore } from "./types";
+import { KeyRingStatus, KeyStore, ResetPasswordError } from "./types";
 import {
   readVecStringPointer,
   readStringPointer,
 } from "@anoma/crypto/src/utils";
 import { deserialize } from "borsh";
+import { Result } from "@anoma/utils";
 
 // Generated UUID namespace for uuid v5
 const UUID_NAMESPACE = "9bfceade-37fe-11ed-acc0-a3da3461b38c";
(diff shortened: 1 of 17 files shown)


apps/extension/src/background/keyring/keyring.ts
```diff
@@ -121,11 +122,15 @@ export class KeyRing {
     await this.activeAccountStore.set(PARENT_ACCOUNT_ID_KEY, parentId);
   }
 
-  public async checkPassword(password: string): Promise<boolean> {
-    const activeAccountId = await this.getActiveAccountId();
+  public async checkPassword(password: string, accountId?: string): Promise<boolean> {
+    // default to active account if no account provided
+    let idToCheck = accountId;
```
mateuszjasiuk, line 127: You could do this :)
`const idToCheck = accountId ?? (await this.getActiveAccountId());`

packages/utils/src/helpers/index.ts
```diff
@@ -126,3 +126,35 @@ export const truncateInMiddle = (
...
+export type Err<E> = { ok: false, error: E };
+
+/**
+ * Result type used to indicate success or failure.
+ * A value of type T is returned on success (Ok), and an error of type
+ * E is returned on failure (Err).
+ *
+ * Ok and Err types can be built with the helper functions ok and err.
+ */
+export type Result<T, E> = Ok<T> | Err<E>;
```
mateuszjasiuk, line 152: This is cool :D
What do you think about namespacing functions to give more context?.
Something like:
```
export type Result<T, E> = Ok<T> | Err<E>;

export const Result = {
  ok: <T>(value: T): Ok<T> => {
    return { ok: true as const, value };
  },

  err: <E>(error: E): Err<E> => {
    return { ok: false as const, error };
  },
};
```

mateuszjasiuk (approved):
Tested and LGTM! :)
Added two small suggestions.
There is a conflict in the keyring to resolve, but it's only the imports.

apps/extension/src/background/keyring/keyring.ts
```diff
@@ -138,6 +143,48 @@ export class KeyRing {
     return false;
   }
 
+  // Reset password by re-encrypting secrets with new password
```
jurevans, line 146: This works very well for me the first time I use it! I did notice something though: If I change my password, it doesn't always force me to re-enter it when I go to derive a new account (unless maybe the service worker stops)? So, if I go to reset password, and immediately go to derive a new address, I get an error if it doesn't prompt me for the new password. There's a couple things we might do: Either 1) explicitly lock the keyring after resetting password, or maybe better, 2)  Set the password in Keyring to the new password here. Otherwise, this all looks good!

apps/extension/src/background/keyring/keyring.ts
emccorson: I like that. It's much nicer. I'll add it in.

packages/utils/src/helpers/index.ts
emccorson: I like this. I was looking in to namespacing before but I didn't realise you could export a type and value with the same name in TS.

apps/extension/src/background/keyring/keyring.ts
emccorson: Ah! Thanks for spotting that. I've updated it now to change `this._password` on reset.

PR #279, opened 2023-05-18
fix: fix useSanitizedLocation hook

emccorson:
The hook broke when it was moved from namada-interface to packages/hooks. There were two versions of react-router-dom present, and the interface and the package were being built with different versions, which caused the Location context objects to not be equal, ultimately causing useLocation to fail.

This commit aligns the version of react-router-dom in the extension with the version in the interface, which ensures only one copy is installed by `yarn install`.


jurevans (commented):
LGTM!

PR #281, opened 2023-05-21
feat/280 - Implement events for new address & new mnemonics

### LINKED ISSUE #280: Extension: Add event broadcast on new accounts
When a new top-level account is added, or the Ledger hardware wallet is connected and account loaded into the background service, we should broadcast a message to connected tabs to reload account views.

- [ ] Creating a top-level account reloads interface accounts correctly
- [ ] Deriving a new address updates the interface accounts accordingly

jurevans:
Resolves #280 

__NOTE__ This still needs to be added to Ledger accounts once that service is implemented! This is simple enough to be handled in the PR that implements the Ledger service.

- [ ] Adding a new address reloads accounts in interface
- [ ] Adding a new mnemonic account reloads accounts in interface (sets it to the new top-level account)

### DIFF
--- a/apps/extension/src/App/Accounts/AccountListing.tsx
+++ b/apps/extension/src/App/Accounts/AccountListing.tsx
@@ -63,7 +63,8 @@ const AccountListing = ({ account, parentAlias }: Props): JSX.Element => {
           </Button>
         )}
         <Button
-          onClick={() => {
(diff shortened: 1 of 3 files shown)


mateuszjasiuk (approved):
LGTM!

PR #284, opened 2023-05-22
bug/282 - Fix all service worker-related errors in Create Account

### LINKED ISSUE #282: bug - Extension: Fix all service worker-related errors
The extension will throw an error and need to be reload if the service worker times out before an action can be performed. We can avoid this by using `useUntil` along with our initial queries to the service worker.

- [ ] Extension should not throw error if no user interaction occurs and the service worker has stopped

For any pop-up `App` and `Setup` views, we should make use of `useUntil` for any feature that requires the service worker to be active to function.

__NOTE__ this should probably be deferred until https://github.com/anoma/namada-interface/pull/274 has been merged, as there has been a bit of refactoring in that PR that affects this issue.

jurevans:
Partially resolves #282 

This PR adds the functionality from `useUntil` in several key places throughout the Account Creation setup process, where the app will break due to the service worker being inactive when the user request an action requiring the background scripts.

### DIFF
--- a/apps/extension/src/App/Accounts/AccountListing.tsx
+++ b/apps/extension/src/App/Accounts/AccountListing.tsx
@@ -1,6 +1,10 @@
 import { useNavigate } from "react-router-dom";
+
 import { Icon, IconName } from "@anoma/components";
 import { AccountType, Bip44Path, DerivedAccount } from "@anoma/types";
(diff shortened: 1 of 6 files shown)


mateuszjasiuk (approved):
LGTM!

PR #285, opened 2023-05-22
Feat: remove copy to clipboard button

emccorson:
This PR contains two commits:

- feat: remove copy to clipboard button and warn user
- feat: disallow highlighting seed phrase in production

Only the first commit is specified by the audit report. The second made sense to me but I am happy to remove the commit if it is not a desired change.


emccorson:
Fixed now to only remove copy to clipboard button in production.

jurevans (approved):
This looks good to me! We may get some feedback from product at some point to see if we want to block user selection, but I think this is good for now for the auditors!

Issue #200, opened 2023-01-24
Extension: Tx and Connection Approval


jurevans:
If the interface requests to connect to the extension, or calls a signing function from the extension's API, the extension should open the user and prompt for approval, and password if needed.

As to when the user is prompted for a password, the approach that I think is best would be:

1. User issues a Tx from the interface
2. The extension approval window opens, and if the user _approves_ the tx, then:
3. Prompt for password to unlock keyring (if it is locked), sign the Tx, then resolve the promise

This would result in the keyring only being unlocked when it is absolutely needed (e.g., for when an account's private key is needed to sign a tx).

PR #290, opened 2023-05-24
feat/200 - Tx approval screens

jurevans:
Partly addresses #200 

Enable user to decrypt keys and submit transparent transfer transactions

- [x] User is prompted to authenticate and approve of a transaction
- [x] Upon submitting approved transaction, remove from storage, dispatch event to the interface informing it to re-fetch balances
- [x] Upon rejecting approved transaction, remove from storage


apps/extension/src/Approvals/ApproveTransfer/ApproveTransfer.tsx
```diff
@@ -0,0 +1,89 @@
...
+    navigate(TopLevelRoute.ConfirmTx);
+  };
+
+  const handleReject = async (): Promise<void> => {
+    try {
+      // TODO: use executeUntil here!
+      await requester.sendMessage(Ports.Background, new RejectTransferMsg(id));
+
+      // Close tab
+      const tab = await browser.tabs.getCurrent();
```
mateuszjasiuk, line 57: Maybe worth making close tab a utility function?

apps/extension/src/background/approvals/messages.ts
```diff
@@ -0,0 +1,67 @@
...
+  constructor(
+    public readonly msgId: string,
+    public readonly address: string,
+    public readonly password: string
+  ) {
+    super();
+  }
+
+  validate(): void {
+    if (!this.msgId || this.msgId === "") {
```
mateuszjasiuk, line 48: I think `|| this.msgId === ""` will always be `true` if `!this.msgId` is `true` :)
So we can either remove right operand or make it more explicit by checking for null, undefined and ""

mateuszjasiuk (approved):
LGTM! Tested and works nice!
Not need to remember about password all the time is super convenient :)
Left two small comments

mateuszjasiuk (approved):
LGTM! Tested and works nice!
Not need to remember about password all the time is super convenient :)
Left two small comments

Issue #239, opened 2023-03-23
Return status info from the SDK transaction functions


mateuszjasiuk:
**Depends on SDK merge**

Right now we do not return any info about the status of the  transaction from the SDK.

Because of that if:
- transaction fails we can't display proper reason for the user
- transaction is successful we can't display any specifics about txid, hash, gas used etc.
- **if transaction does not "pass" the VP it is still considered successful**

PR #291, opened 2023-05-29
feat: transfers from shielded addresses

mateuszjasiuk:
- [x] Set TransferSource - either Address or xsk
- [x] Handle errors with notifications - right now failed transfers are treated as "Completed"
- [x] Fix transfers in FF
- [ ] ~~import "sdk-store"~~
- [x] Change Namada rev hash to the proper one
- [x] Replace loaded tx wasms with PathBuf
- [x] Fix notifications in Firefox - cloneInto

**Test against namada `1e94882d9feb6302b743ec3771fe2df17c168266` to be sure wasms are on chain**
---
In the next PRs:
- [x] Transfers reject by VP are still "valid" in a sense that future returned from `submit_transfer` function is resolved not rejected. Because of that we can't handle errors thrown by VPs, i.e. MASP epoch check. #239 
- [x] We need to figure out where to keep shielded params, so they are always accessible to download. #292
- [x] fetch_masp_params loader/progress bar #293
- [x] remove all of the tx wasms from repo #296


jurevans (approved):
Nice! Tested locally, and this LGTM!

PR #295, opened 2023-05-30
feat/253 Implement ledger service in Extension

### LINKED ISSUE #253: Support Ledger devices via Zondax app
We need to support connecting to the Ledger Nano S+ device via the Zondax integration.

See https://github.com/Zondax/ledger-namada

- [x] User can choose to connect their Ledger device
- [x] User is prompted to unlock their Ledger device if needed
- [x] User is directed to a connect to Ledger flow (See: https://support.ledger.com/hc/en-us/articles/4404366864657-Connect-your-Ledger-to-Metamask for an example of what this looks like on Metamask - similar to Keplr)
- [x] User can access accounts on their Ledger device
- [ ] Test transfer transaction signing using keys on their Ledger device
- [ ] Block PR merging until `murisi/separate-signing` is merged to Namada (See https://github.com/anoma/namada/pull/1498)

### LINKED ISSUE #335: Approve Bond Tx with Ledger signing
This relates to the work completed in https://github.com/anoma/namada-interface/issues/253

Currently, submitting a Bond Tx does not have a corresponding approval process prompting the user to unlock their keys for signing. This needs to be put in place, and we also need the ability to sign via the Ledger.

- [x] User is prompted to unlock keyring and approve a Bond Tx
- [x] If user has their Ledger device selected, it should make use of the Ledger for signing

jurevans:
Resolves #253 
Resolves #335 

This PR adds the Ledger service to the extension, and adds Bond Tx to the approval process, allowing user to sign a Bond Tx with their Ledger (along with Reveal Public Key tx)

- [x] User can select Ledger accounts from interface
- [x] Extension can track alias and paths to use with Ledger
- [x] Submitting a Bond for the first time requires and prompts user to approve a Reveal Public Key tx
- [x] Submitting a Bond Tx prompts user for approval
- [x] Approving a Bond Tx prompts for password (if using a private key)
- [x] Approving a Bond Tx provides Ledger prompts (if using Ledger account)
- [x] Rejecting a Bond Tx removes pending transaction
- [x] Bond is successfully submitted through SDK
- [x] Bond is successfully signed with the Ledger keys, and submits through the SDK

**Nice-to-have**

- [ ] Submitting a bond gives the user the appropriate notification in interface (e.g., "Bond is submitting" or something)
  - _TODO_ This is partially hooked up, but there still remains an issue with updating amount staked. Token balance _after_ bonding does currently work

## NOTES

- This PR includes a built copy of the `ledger-namada` package in `packages/`, which contains the necessary changes on the client-side to parse Ledger responses, so we are no longer using the `@zondax/ledger-namada` package.

# Testing

### Setup
2. Ping me for a copy of the installer script you need for your Ledger, then install with `./installer_s2.sh load` (e.g., if you're using a Ledger Nano S-Plus)

### Testing Bond tx
1. In the interface, select either your initial account, or ledger account, and click the `Staking` menu item (top menu)
2. You will see a link to a validator, so click that
3. From this view, click `Stake` (there is an account drop down, but you don't need to change this if the account selected has funds, but you can as an option)
4. In the modal window, enter an amount you would like to stake, then click `Confirm`
5. In the popup window, click to confirm the transaction:
  - If you selected a local wallet account, you will be prompted for a password to unlock your key and sign the transaction
  - If you selected Ledger, you will see the following:
     1. The first time you use your Ledger account, you will be prompted on the device to approve a reveal public key tx (go ahead and click through to approve!)
     2. Following that, you will need to also approve the `Bond` tx. Once this is approved, you should see a success message once it completes

### SCREENSHOTS

__Ledger account loaded in extension__

Ledger-Account-extension

__Ledger account loaded in interface__

Ledger Account - Interface


packages/utils/src/helpers/index.ts
```diff
@@ -160,3 +164,19 @@ export const Result = {
...
+ *
+ * @param {number} coinType - SLIP-044 Coin designation
+ * @param {Bip44Path} bip44Path - path object
+ * @returns {string}
+ */
+export const makeBip44Path = (
+  coinType: number,
+  bip44Path: Bip44Path
+): string => {
+  const { account, change, index } = bip44Path;
```
emccorson, line 179: `index` is optional on `Bip44Path` so if it is undefined this will return something like `"m/44'/877'/0'/0/undefined"`.

packages/types/src/tx/schema/transfer.ts
```diff
@@ -8,6 +8,7 @@ export class TransferMsgValue {
   source: string;
   target: string;
   token: string;
+  public_key?: string;
```
emccorson, line 11: Where does this get used? I couldn't see anywhere but I might just be missing it.

jurevans: https://github.com/anoma/namada-interface/blob/873cd3181a8ede971dd54f850102515fbdd0a172/packages/shared/lib/src/sdk/tx.rs#L276-L316

It gets used here - when this msg is deserialized, we can use the public key (if provided) to send as the `verification_key`

packages/utils/src/helpers/index.ts
jurevans: Good catch! Also, once fixed, I should probably update the other instances where it can be utilized.

jurevans: I pushed up a fix, and added it to Keyring where an address _without_ an index needs to be formatted (mnemonic account)

packages/types/src/tx/schema/transfer.ts
jurevans: At first I thought this was the Tx props - but you are correct, `public_key` got moved to Tx and out of transfer, and I just forgot to remove it!

apps/extension/src/background/approvals/messages.ts
```diff
@@ -4,6 +4,7 @@ import { ROUTE } from "./constants";
 enum MessageType {
   RejectTransfer = "reject-transfer",
   SubmitApprovedTransfer = "submit-approved-transfer",
+  SubmitApprovedLedgerTransfer = "submit-approved-ledger-transfer",
```
jurevans, line 7: Remove me

apps/extension/src/Setup/Ledger/LedgerConfirmation.tsx
```diff
@@ -0,0 +1,89 @@
...
+import { closeCurrentTab } from "utils";
+
+const LedgerConfirmation: React.FC = () => {
+  const navigate = useNavigate();
+  const requester = useRequester();
+  const themeContext = useContext(ThemeContext);
+  const { alias = "", address = "", publicKey = "" } = useSanitizedParams();
+  const [error, setError] = useState<string>();
+
+  const handleSubmitClick = async (): Promise<void> => {
```
mateuszjasiuk, line 38: I think we can use `useCallback` here as this does not use any state.

apps/extension/src/utils/index.ts
```diff
@@ -1,12 +1,58 @@
...
 };
+
+/**
+ * Return all unencrypted values from account stores
+ */
+export const getAccountValuesFromStore = (
+  accounts: AccountStore[]
+): DerivedAccount[] => {
+  return accounts.map(
+    ({
```
mateuszjasiuk, line 24: It would be great to add `pick` utility function or something :)

packages/shared/lib/src/sdk/mod.rs
```diff
@@ -90,9 +97,20 @@ impl Sdk {
     ) -> Result<(), JsError> {
         let args = tx::bond_tx_args(tx_msg, password)?;
 
-        namada::ledger::tx::submit_bond(&mut self.client, &mut self.wallet, args)
+        let (mut tx, _, pk) =
+            namada::ledger::tx::build_bond(&mut self.client, &mut self.wallet, args.clone())
+                .await
+                .map_err(|e| JsError::from(e))?;
```
mateuszjasiuk, line 103: I've learned recently that you can pass function as a rust closure so:
`.map_err(JsError::from)?;`
should be enough

packages/shared/lib/src/sdk/mod.rs
```diff
@@ -102,9 +120,87 @@ impl Sdk {
...
+        &self,
+        tx_bytes: &[u8],
+        data_key: String,
+        code_key: String,
+        header_key: String,
+    ) -> Result<JsValue, JsError> {
+        let mut tx: Tx = Tx::try_from(tx_bytes).map_err(|e| JsError::from(e))?;
+        let data_secret = SecretKey::from_str(&data_key).map_err(|e| JsError::from(e))?;
+        let code_secret = SecretKey::from_str(&code_key).map_err(|e| JsError::from(e))?;
+        let header_secret = SecretKey::from_str(&header_key).map_err(|e| JsError::from(e))?;
```
mateuszjasiuk, line 169: Also something to consider is moving those functions calls out of the sdk struct to another struct/function(that does not use wasm_bindgen). This way you do not have to map_err all the time, you can just map_err result of your new function.

packages/shared/lib/src/sdk/mod.rs
```diff
@@ -131,9 +238,20 @@ impl Sdk {
...
+        let (mut tx, _, pk) =
+            namada::ledger::tx::build_ibc_transfer(&self.client, &mut self.wallet, args.clone())
+                .await
+                .map_err(|e| JsError::from(e))?;
+
+        signing::sign_tx(&mut self.wallet, &mut tx, &args.tx, &pk)
+            .await
+            .map_err(|e| JsError::from(e))?;
+
+        namada::ledger::tx::process_tx(&self.client, &mut self.wallet, &args.tx, tx)
```
mateuszjasiuk, line 250: Usually I'm not the biggest fan of this, but maybe it's worth adding a function like `sign_and_process_tx` :stuck_out_tongue_closed_eyes:  as it seems that sign and process are often called together

mateuszjasiuk (commented):
LGTM left few minor comments!
I will approve after testing!

packages/shared/lib/src/sdk/mod.rs
jurevans: Nice! Somehow I forgot about that feature

apps/extension/src/utils/index.ts
jurevans: I added a pick utility. I originally wanted to do an `omit` utility as well, but I think this way it's more clear what's being included, and we can add that later if it seems useful enough!

apps/extension/src/Setup/Ledger/LedgerConfirmation.tsx
jurevans: Done!

packages/shared/lib/src/sdk/mod.rs
jurevans: Done!

packages/shared/lib/src/sdk/mod.rs
jurevans: This is now done!

apps/extension/src/background/approvals/service.ts
```diff
@@ -2,74 +2,162 @@ import browser from "webextension-polyfill";
...
 
     throw new Error("Pending transfer not found!");
   }
 
+  // Authenticate keyring and submit approved bond transaction from storage
+  async submitBond(msgId: string, password: string): Promise<void> {
+    try {
+      await this.keyRingService.unlock(password);
+    } catch (e) {
+      throw new Error(`${e}`);
```
mateuszjasiuk, line 118: I guess this does not do much? :D

mateuszjasiuk (approved):
I think this looks good. Left few comments/questions. I'm curious if it is possible to have one flow instead of two for submitting transactions. So for example we get rid off the submit_bond and we use only submit_signed_bond. The only difference is that we are asking sdk to sign instead of ledger.  This would potentially simplify few things :)

apps/extension/src/background/approvals/service.ts
```diff
@@ -2,74 +2,162 @@ import browser from "webextension-polyfill";
...
+    await this.txStore.set(id, txMsg);
 
     // Decode tx details and launch approval screen
     const txDetails = deserialize(txMsgBuffer, TransferMsgValue);
     const { source, target, token, amount: amountBN } = txDetails;
     const amount = new BigNumber(amountBN.toString());
     const url = `${browser.runtime.getURL(
       "approvals.html"
-    )}#/tx?id=${id}&source=${source}&target=${target}&token=${token}&amount=${amount}`;
+    )}#/approve-transfer?type=${type}&id=${id}&source=${source}&target=${target}&token=${token}&amount=${amount}`;
```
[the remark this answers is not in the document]
jurevans: yep, I agree! We used to have something like that, I can add it back in.

apps/extension/src/App/Settings/Settings.tsx
```diff
@@ -75,27 +67,31 @@ const Settings: React.FC<{
...
-        new SetActiveAccountMsg(id)
+        new SetActiveAccountMsg(id, type as ParentAccount)
       );
 
       // Lock current wallet keyring:
       await requester.sendMessage(Ports.Background, new LockKeyRingMsg());
 
       // Fetch accounts for selected parent account
-      await onSelectAccount(account);
+      onSelectAccount(account);
```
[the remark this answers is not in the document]
jurevans: I removed it because it's not an async function (throwing lint warning)

apps/extension/src/background/approvals/service.ts
```diff
@@ -2,74 +2,162 @@ import browser from "webextension-polyfill";
...
+    const txMsgBuffer = Buffer.from(fromBase64(txMsg));
+    const id = uuid();
+    await this.txStore.set(id, txMsg);
+
+    // Decode tx details and launch approval screen
+    const txDetails = deserialize(txMsgBuffer, SubmitBondMsgValue);
+
+    const { source, nativeToken, amount: amountBN } = txDetails;
+    const amount = new BigNumber(amountBN.toString());
+    // TODO: This query should include perhaps a "type" indicating whether it's a bond or unbond tx:
```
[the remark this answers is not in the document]
jurevans: I think for `type` I will make it a part of the routing, so we can just direct it by URL to catch that param,  but the `type` I've added here is account type, not tx type - it needs to be more explicit because it's confusing!

apps/extension/src/Approvals/ApproveTransfer/ConfirmLedgerTransfer.tsx
```diff
@@ -0,0 +1,103 @@
...
+
+      // Submit signatures for tx
+      await requester.sendMessage(
+        Ports.Background,
+        new SubmitSignedTransferMsg(msgId, toBase64(bytes), signatures)
+      );
+      setStatus(Status.Completed);
+    } catch (e) {
+      console.warn(e);
+      const ledgerErrors = await ledger.queryErrors();
```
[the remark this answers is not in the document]
jurevans: The only error that can happen is that the Ledger instance can't find `NamadaApp` - so either the app is not loaded, or the Ledger isn't plugged in an connected. The query for `status()` in `background/ledger/ledger.ts` catches that situation there, this is more of a wrapper function for `status()` that returns any non-error error messages so the client doesn't have to do that logic. However, if that error was thrown in the `Ledger` status call, it isn't getting caught here, which isn't good. Maybe it should be moved to the above `try` so that the status error, if thrown, will be caught.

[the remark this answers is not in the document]
jurevans: Also, I just realized that this is `ConfirmLedgerTransfer` - in my next PR I am getting rid of this file entirely - I took out the `queryErrors` call in `ConfirmLedgerBond`, so I'll leave this as is! Sorry I thought this was the other confirmation!

apps/extension/src/background/approvals/service.ts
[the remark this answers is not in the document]
mateuszjasiuk: I think we can do it later, to not change this PR too much

apps/extension/src/Approvals/ApproveBond/ApproveBond.tsx
```diff
@@ -0,0 +1,94 @@
...
+
+  const handleApproveClick = (): void => {
+    setMsgId(id);
+    if (type === AccountType.Ledger) {
+      return navigate(`${TopLevelRoute.ConfirmLedgerBond}`);
+    }
+    navigate(TopLevelRoute.ConfirmBond);
+  };
+
+  const handleReject = async (): Promise<void> => {
```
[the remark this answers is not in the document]
jurevans: Done!

apps/extension/src/background/approvals/service.ts
jurevans: I removed it for now!

apps/extension/src/Approvals/Approvals.tsx
```diff
@@ -36,15 +39,44 @@ export const Approvals: React.FC = () => {
...
+              }
+            />
+            <Route
+              path={TopLevelRoute.ConfirmBond}
+              element={<ConfirmBond msgId={msgId} address={address} />}
+            />
+            <Route
+              path={TopLevelRoute.ConfirmLedgerBond}
+              element={
+                <ConfirmLedgerBond
```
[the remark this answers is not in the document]
jurevans: I'm thinking I will start by ignoring differences in Tx (reducing the query string that opens the approval window) and keeping common values (`amount`, `type`, `source`) for now, then incorporate differences later as needed. That should make for much cleaner code

PR #297, opened 2023-05-31
feat: allow user to delete account

### LINKED ISSUE #242: No Option To Delete Wallets

Figure out if we want to allow users to delete wallet from the UI.

emccorson:
Resolves #242

### DIFF
--- a/apps/extension/src/App/Settings/ExtraSettings/DeleteAccount/DeleteAccount.components.ts
+++ b/apps/extension/src/App/Settings/ExtraSettings/DeleteAccount/DeleteAccount.components.ts
@@ -1,0 +1,44 @@
+import styled from "styled-components";
+
+export const Input = styled.input`
+  width: 100%;
+  box-sizing: border-box;
+  font-size: 16px;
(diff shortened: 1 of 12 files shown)


apps/extension/src/App/Settings/ExtraSettings/DeleteAccount/DeleteAccount.components.ts
```diff
@@ -0,0 +1,44 @@
+import styled from "styled-components";
+
+export const Input = styled.input`
```
jurevans, line 3: I think at some point (not in this PR) we could incorporate `Input`, `InputFeedback` and `InputContainer` into `@anoma/components` and standardize usage across the apps, but I think it would be good to first take a good look at our existing `Input` there and see if it suits our needs to be used everywhere. That's a future task that could be improved when we're incorporating designs I think

jurevans (approved):
LGTM!

mateuszjasiuk (approved):
LGTM!

PR #301, opened 2023-06-07
bug: amounts as strings

mateuszjasiuk:
For now we will just parse strings to int. In Future we would probably construct BigInt out of the string.


jurevans (approved):
LGTM!

PR #302, opened 2023-06-07
feat: initialize masp params load from the extension

### LINKED ISSUE #293: Add loader while downloading masp params.
We need to notify users that masp params are being downloaded as it may take some time.
We also need to figure out when to invalidate cached params, when they are not longer valid.

mateuszjasiuk:
Resolves #293
Moved initialization of the masp params download into the extension app.
It should fix service worker initialization errors.
- [x] add HasMaspParamsMsg to check if params are already loaded
- [x] add LoadMaspParamsMsg to load the params
- [x] add spinner and status messages

In the next PRs:
- [x] Make loading async - we need to refactor sdk first as it might lead to "multiple mutable references" error #304
- [x] Figure out how to invalidate stored masp.params #303

Preview:
fetch masp

### DIFF
--- a/apps/extension/src/App/App.tsx
+++ b/apps/extension/src/App/App.tsx
@@ -8,7 +8,11 @@ import { Icon, IconName } from "@anoma/components";
 import { useUntil } from "@anoma/hooks";
 
 import { Ports } from "router";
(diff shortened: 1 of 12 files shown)


jurevans (approved):
LGTM!

PR #305, opened 2023-06-09
feat: async masp params loading

### LINKED ISSUE #304: Make loading MASP params async
Load async params asynchronously. Avoid multiple mutable references error.
- It might require to do some refactoring to the Sdk struct

mateuszjasiuk:
Resolves #304
We are loading params in two cases:
- when opening extension- non blocking fetch
fetch-masp1

- when submitting a transfer - blocking fetch
fetch-masp2

### DIFF
--- a/apps/namada-interface/src/App/Token/TokenSend/TokenSendForm.tsx
+++ b/apps/namada-interface/src/App/Token/TokenSend/TokenSendForm.tsx
@@ -420,7 +420,7 @@ const TokenSendForm = ({
         <Button
           loading={isTransferSubmitting}
           variant={ButtonVariant.Contained}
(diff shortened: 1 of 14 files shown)


apps/namada-interface/src/App/Token/TokenSend/TokenSendForm.tsx
```diff
@@ -420,7 +420,7 @@ const TokenSendForm = ({
         <Button
           loading={isTransferSubmitting}
           variant={ButtonVariant.Contained}
-          // disabled={isFormInvalid}
+          disabled={isFormInvalid}
```
mateuszjasiuk, line 423: Regression :)

PR #308, opened 2023-06-14
fix: keep sdk wallet storage entry by parent account id

mateuszjasiuk:
- Changed "sdk-store" to Record<string, string> where key is a parent account id and value is serialized encrypted sdk wallet storage :)
- Added functionality to remove account from the sdk storage

**After this one is merged remove and add extension again to clear old storages**


jurevans (approved):
LGTM!

PR #312, opened 2023-06-19
feat: update namada to 0.17.3

mateuszjasiuk:
Make sure to:
- Install [CometBFT](https://github.com/cometbft/cometbft/releases/tag/v0.37.1)
- Recreate whole chain as usual - do not forget to change chain id in the envs like me :laughing:. Otherwise you might be able to query data, but submitting tx will fail.
- It looks like CORS issue is not an issue anymore :crossed_fingers:

### DIFF
--- a/packages/shared/lib/Cargo.lock
+++ b/packages/shared/lib/Cargo.lock
@@ -1982,8 +1982,8 @@ checksum = "e5ce46fe64a9d73be07dcbe690a38ce1b293be448fd8ce1e6c1b8062c9f72c6a"
 
 [[package]]
 name = "namada"
-version = "0.16.0"
-source = "git+https://github.com/anoma/namada?rev=1e94882d9feb6302b743ec3771fe2df17c168266#1e94882d9feb6302b743ec3771fe2df17c168266"
+version = "0.17.3"
+source = "git+https://github.com/anoma/namada#85e7eff6c663641d0cdf42deca78d3db9a31e5db"
 dependencies = [
  "async-std",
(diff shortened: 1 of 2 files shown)


emccorson (approved):
Looks good!

One thing is that I still had to fix the CORS thing to get it to work, but once that was done I could see my balances and get a transparent -> transparent transaction to work.

By the way, it looks like this file:
```
~/.local/share/namada/${CHAIN_ID}/setup/validator-0/.namada/${CHAIN_ID}/cometbft/config/config.toml
```
...is overwritten with the contents of:
```
~/.local/share/namada/${CHAIN_ID}/setup/validator-0/.namada/${CHAIN_ID}/config.toml
```
...when Namada is started, so the file that should be changed for CORS is the second one.

mateuszjasiuk:
> Looks good!
> 
> One thing is that I still had to fix the CORS thing to get it to work, but once that was done I could see my balances and get a transparent -> transparent transaction to work.
> 
> By the way, it looks like this file:
> 
> ```
> ~/.local/share/namada/${CHAIN_ID}/setup/validator-0/.namada/${CHAIN_ID}/cometbft/config/config.toml
> ```
> 
> ...is overwritten with the contents of:
> 
> ```
> ~/.local/share/namada/${CHAIN_ID}/setup/validator-0/.namada/${CHAIN_ID}/config.toml
> ```
> 
> ...when Namada is started, so the file that should be changed for CORS is the second one.

Not sure how it worked for me though :D Thanks!

PR #314, opened 2023-06-22
feat: scan shielded addresses

mateuszjasiuk:
Added functionality to scan for shielded addresses when importing accounts

### DIFF
--- a/apps/extension/src/background/keyring/keyring.ts
+++ b/apps/extension/src/background/keyring/keyring.ts
@@ -351,7 +351,8 @@ export class KeyRing {
 
   private async *getAddressWithBalance(
     seed: VecU8Pointer,
-    parentId: string
+    parentId: string,
+    type: AccountType
   ): AsyncGenerator<
     {
       path: Bip44Path;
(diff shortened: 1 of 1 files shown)


apps/extension/src/background/keyring/keyring.ts
```diff
@@ -373,13 +379,9 @@ export class KeyRing {
...
       const path = { account: 0, change: 0, index };
-      const accountInfo = this.deriveTransparentAccount(
-        seedClone,
-        path,
-        parentId
-      );
+      const accountInfo = deriveFn(seedClone, path, parentId);
       const balances: [string, string][] = await this.query.query_balance(
-        accountInfo.address
+        accountInfo.owner
```
emccorson, line 384: What does this change do?

emccorson (approved):
Added one question, but code looks good to me and the fix worked :+1:

apps/extension/src/background/keyring/keyring.ts
mateuszjasiuk: So this was the source of a bug that I mentioned. Basically for querying balances what we need to pass is an owner, which is either address for transparent accounts or extended viewing key for shielded.

PR #315, opened 2023-06-23
Fix up staking

emccorson:
- refactor(StakingOverview): Split into separate components

- fix: Display some unfetched amounts as "-" instead of "NAM 0"

- fix: Use more redux selectors instead of passing props

- feat(StakingBalancesList): Display total unbonded and withdrawable amounts

- feat(MyValidatorsTable): Display unbonded and withdrawable amounts

- feat(AllValidatorsTable): Add filtering by search and sorting by column

- feat(ValidatorDetails): Add UI to withdraw unbonded tokens

- feat(ValidatorDetails): Display each bond and unbond individually

- fix(ValidatorDetails): Display http URLs as links

- fix(UnbondPosition): Make stakedAmount BigNumber instead of number

- fix(StakingAndGovernance/types): Make ChangeInStakingPosition amount always positive (negative does not represent unbond)

- fix(StakingAndGovernance/types): Make more types optional; for when data is unfetched

- fix(StakingAndGovernance/types): Make some amounts BigNumbers; for use in calculations

- fix(StakingAndGovernance/actions): Make voting power 1:1 with staked amount in namnam

- fix(Modal): Apply app theme to modals

- feat(Table): Add subheading slot; for elements between table and heading

- feat(Table): Add onClick prop; called when a column heading is clicked

- fix(shared): Send bond amount in NAM

### DIFF
--- a/apps/namada-interface/src/App/Staking/StakingOverview/AllValidatorsTable/AllValidatorsTable.tsx
+++ b/apps/namada-interface/src/App/Staking/StakingOverview/AllValidatorsTable/AllValidatorsTable.tsx
@@ -1,0 +1,181 @@
+import { useState, useCallback } from "react";
+import { Table, TableLink, TableConfigurations } from "@namada/components";
+import { Validator } from "slices/StakingAndGovernance";
+import {
+  AllValidatorsSearchBar
+} from "./AllValidatorsTable.components";
+import { formatPercentage, assertNever } from "@namada/utils";
+import { useAppSelector, RootState } from "store";
+import { ValidatorsCallbacks } from "../StakingOverview";
+
+// AllValidators table row renderer and configuration
+// it contains callbacks defined in AllValidatorsCallbacks
+const AllValidatorsRowRenderer = (
+  validator: Validator,
+  callbacks?: ValidatorsCallbacks
+): JSX.Element => {
+  return (
+    <>
+      <td>
+        <TableLink
+          onClick={() => {
+            const formattedValidatorName = validator.name
+              .replace(" ", "-")
+              .toLowerCase();
+
+            callbacks && callbacks.onClickValidator(formattedValidatorName);
+          }}
+        >
+          {validator.name}
+        </TableLink>
+      </td>
+      <td>{validator.votingPower?.toString() ?? ""}</td>
+      <td>{formatPercentage(validator.commission)}</td>
+    </>
+  );
+};
+
+const getAllValidatorsConfiguration = (
+  navigateToValidatorDetails: (validatorId: string) => void,
+  onColumnClick: (column: AllValidatorsColumn) => void,
+  sort: Sort,
+): TableConfigurations<Validator, ValidatorsCallbacks> => {
+  const getLabelWithTriangle = (column: AllValidatorsColumn): string => {
+    let triangle = "";
+    if (sort.column === column) {
+      if (sort.ascending) {
+        triangle = " \u25b5"; // white up-pointing small triangle
+      } else {
+        triangle = " \u25bf"; // white down-pointing small triangle
+      }
+    }
+
+    return `${column}${triangle}`;
+  }
+
+  return {
+    rowRenderer: AllValidatorsRowRenderer,
+    callbacks: {
+      onClickValidator: navigateToValidatorDetails,
+    },
+    columns: [
+      {
+        uuid: "1",
+        columnLabel: getLabelWithTriangle(AllValidatorsColumn.Validator),
+        width: "45%",
+        onClick: () => onColumnClick(AllValidatorsColumn.Validator),
+      },
+      {
+        uuid: "2",
+        columnLabel: getLabelWithTriangle(AllValidatorsColumn.VotingPower),
+        width: "25%",
+        onClick: () => onColumnClick(AllValidatorsColumn.VotingPower),
+      },
+      {
+        uuid: "3",
+        columnLabel: getLabelWithTriangle(AllValidatorsColumn.Commission),
+        width: "30%",
+        onClick: () => onColumnClick(AllValidatorsColumn.Commission),
+      },
+    ],
+  };
+};
+
+const sortValidators = (sort: Sort, validators: Validator[]): Validator[] => {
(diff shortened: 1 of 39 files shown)


apps/namada-interface/src/App/Staking/StakingOverview/AllValidatorsTable/AllValidatorsTable.tsx
```diff
@@ -0,0 +1,178 @@
...
+
+const sortValidators = (sort: Sort, validators: Validator[]): Validator[] => {
+  const direction = sort.ascending ? 1 : -1;
+
+  const ascendingSortFn: (a: Validator, b: Validator) => number =
+    sort.column === AllValidatorsColumn.Validator ?
+      (a, b) => a.name.localeCompare(b.name) :
+    sort.column === AllValidatorsColumn.VotingPower ?
+      ((a, b) =>
+        a.votingPower === undefined || b.votingPower === undefined ? 0 :
```
mateuszjasiuk, line 92: I think we could use `!a.votingPower || !b.votingPower` if we want a bit less code

apps/namada-interface/src/App/Staking/StakingOverview/AllValidatorsTable/AllValidatorsTable.tsx
```diff
@@ -0,0 +1,178 @@
...
+        a.votingPower.isLessThan(b.votingPower) ? -1 : 1) :
```
mateuszjasiuk, line 93: Also I think you can just do `a.minus(b)` :)

apps/namada-interface/src/App/Staking/StakingOverview/AllValidatorsTable/AllValidatorsTable.tsx
```diff
@@ -0,0 +1,178 @@
...
+    ],
+  };
+};
+
+const sortValidators = (sort: Sort, validators: Validator[]): Validator[] => {
+  const direction = sort.ascending ? 1 : -1;
+
+  const ascendingSortFn: (a: Validator, b: Validator) => number =
+    sort.column === AllValidatorsColumn.Validator ?
+      (a, b) => a.name.localeCompare(b.name) :
```
mateuszjasiuk, line 89: Had no idea `localCompare` exists :D :+1:

apps/namada-interface/src/App/Staking/StakingOverview/AllValidatorsTable/AllValidatorsTable.tsx
```diff
@@ -0,0 +1,178 @@
...
+  const [sort, setSort] = useState<Sort>({
+    column: AllValidatorsColumn.Validator,
+    ascending: true
+  });
+
+  const sortedFilteredValidators = useAppSelector(
+    selectSortedFilteredValidators(sort, search)
+  );
+
+  const handleColumnClick = (column: AllValidatorsColumn): void =>
```
mateuszjasiuk, line 148: I think this will be redefined every time that search state changes? So maybe it's worth wrapping in useCallback?

mateuszjasiuk (approved):
Tested and works nice! Left couple of minor comments!

jurevans (approved):
Nice! LGTM

apps/namada-interface/src/App/Staking/StakingOverview/AllValidatorsTable/AllValidatorsTable.tsx
```diff
@@ -0,0 +1,178 @@
...
+      {
+        uuid: "2",
+        columnLabel: getLabelWithTriangle(AllValidatorsColumn.VotingPower),
+        width: "25%",
+        onClick: () => onColumnClick(AllValidatorsColumn.VotingPower),
+      },
+      {
+        uuid: "3",
+        columnLabel: getLabelWithTriangle(AllValidatorsColumn.Commission),
+        width: "30%",
```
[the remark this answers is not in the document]
emccorson: I think I like the sound of that. It's a little tricky to work with as it currently is.

apps/namada-interface/src/App/Staking/StakingOverview/AllValidatorsTable/AllValidatorsTable.tsx
emccorson: This one I need to return a `number` whereas `a.minus(b)` is a `BigNumber`, so I would need to convert it anyway.

apps/namada-interface/src/App/Staking/StakingOverview/AllValidatorsTable/AllValidatorsTable.tsx
emccorson: I was a bit unsure when writing this code originally. With `useCallback` I would define the dependency list as `[sort, search]`, or something like that? But then I guess it would still get redefined. I think there is something in redux about memoizing selectors, but I had planned to look into it sometime in the future (and probably there is a lot we could do to optimize the UI rendering). What do you think?

apps/namada-interface/src/App/Staking/StakingOverview/AllValidatorsTable/AllValidatorsTable.tsx
mateuszjasiuk: Ah right!

apps/namada-interface/src/App/Staking/StakingOverview/AllValidatorsTable/AllValidatorsTable.tsx
mateuszjasiuk: But do you need `search` as a dependency? You are only using sort here.  Or am I missing something? :)

emccorson: The selector filters the validators using the `search` value by calling the `filterValidators` function. So it's doing both filtering by the search term, and sorting by the selected column.

mateuszjasiuk (approved):

PR #317, opened 2023-06-29
feat: cleanup memory on rust panic

mateuszjasiuk:
Context:

> So I found a pretty shitty workaround. So it seems like the only way to gc the memory is to free all the references to the WebAssembly instance exports/properties. We can't do it manually by accessing memory or sth.
> The problem tho is that because rust panics the js stops executing(just hangs there). Because of that extension service worker(which we use to run rust sdk code) can't get inactive(and drop the references to WebAssembly exports) :confused:
> So the shitty fix is to timeout calls to rust sdk, after certain amount of time. This way we can catch reject on timeout and handle it as a normal error. So we still have garbage in memory, but it will be freed after service worker goes inactive which is 30 seconds of no received messages. Unfortunately we can't stop the service worker manually, we can only deregister it, but then we can't register it again as this happens automatically on first extension load :neutral_face:

What've changed:
- query functions are wrapped in timeout
- added wasm build script for dev: `wasm:build:dev`
- with `wasm:build:dev` we define panic_hook - added `dev` feature
- changed build targets to `es2015` to be able to extend generated Query class

We can do something similar for the Sdk class, it might be a bit tricky though as for example submit shielded transfer can take a while

### DIFF
--- a/packages/shared/src/index.ts
+++ b/packages/shared/src/index.ts
@@ -1,1 +1,26 @@
+import { Query as RustQuery } from "./shared/shared";
 export * from "./shared/shared";
+
+/**
+ *  CallWithTimeout - calls a function with specified timeout
+ */
+const cwt =
+  <U extends unknown[], T>(fn: (...args: U) => Promise<T>, timeout = 5) =>
+  (...args: U): Promise<T> => {
+    return new Promise(async (resolve, reject) => {
+      const t = setTimeout(() => {
+        reject(`Balance query timed out after ${timeout}s.`);
+      }, timeout * 1000);
+
+      const res = await fn(...args);
+      clearTimeout(t);
+      resolve(res);
+    });
+  };
+
+export class Query extends RustQuery {
+  query_balance = cwt(super.query_balance.bind(this), 10);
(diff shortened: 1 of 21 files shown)


packages/shared/src/index.ts
```diff
@@ -1 +1,26 @@
...
+
+/**
+ *  CallWithTimeout - calls a function with specified timeout
+ */
+const cwt =
+  <U extends unknown[], T>(fn: (...args: U) => Promise<T>, timeout = 5) =>
+  (...args: U): Promise<T> => {
+    return new Promise(async (resolve, reject) => {
+      const t = setTimeout(() => {
+        reject(`Balance query timed out after ${timeout}s.`);
```
jurevans, line 12: Maybe we could do something like [promiseWithTimeout](https://github.com/anoma/namada-interface/blob/48c0595e588c517c233dd4cb0615a08160aecbea/packages/utils/src/async/index.ts#L1-L22) where we pass in the error message as an arg to display, as these are not all balance queries?

mateuszjasiuk: Ah yeah :) Focuesd to much on balances :+1:

PR #321, opened 2023-07-03
fix: use es2015 in interface

emccorson:
This fixes a bug introduced in #317 where `new Query` cannot be called because Query was transpiled to ES5 and relies on the class it extends to also be transpiled, but the parent class was generated from Rust and not transpiled.

PR #317 already fixed this for the extension; this commit just sets the target JS version for namada-interface too.

### DIFF
--- a/apps/namada-interface/tsconfig.json
+++ b/apps/namada-interface/tsconfig.json
@@ -1,7 +1,7 @@
 {
   "compilerOptions": {
     "baseUrl": "./src",
-    "target": "es5",
+    "target": "es2015",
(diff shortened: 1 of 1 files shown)


mateuszjasiuk (approved):
Approved! I'm just curious how did it pass the build/tests :X

emccorson:
I think it's because of the behaviour of `createAsyncThunk`. Any error thrown in its callback is swallowed by redux and it dispatches a rejected action, which means things silently fail :vomiting_face:

PR #322, opened 2023-07-03
feat: do not decrypt mnemonic for auth

### LINKED ISSUE #199: Extension: Authentication should not decrypt mnemonic
Currently, when validating a user's password in the extension, we attempt to decrypt the mnemonic. 

This could be improved (and made more secure) to instead create a random piece of data (perhaps a randomly generated value stored during set-up) and encrypt this with the user's password, in order to later validate the password. The goal of this approach is to avoid side-channel attacks. This value is not otherwise needed, we only need to determine if decryption is successful.

mateuszjasiuk:
Resolves #199


jurevans (approved):
This looks good!

PR #325, opened 2023-07-06
refactor: amounts improvements

mateuszjasiuk:
- Replaced borsh-js with borsh.ts #316
- Replaced rust schemas to accept amount as string


packages/types/src/tx/schema/bond.ts
```diff
@@ -1,42 +1,28 @@
...
-export const SubmitBondMsgSchema = new Map([
-  TxMsgSchema as [unknown, unknown],
-  BondMsgSchema as [unknown, unknown],
-]) as Schema;
+  @field({ type: TxMsgValue })
+  tx!: TxMsgValue;
+
+  constructor(data: SubmitBondProps) {
+    Object.assign(this, data);
+    this.tx = new TxMsgValue(data.tx);
```
emccorson, line 26: What does this line do? I thought `tx` would already be set by the `Object.assign` call above.

apps/namada-interface/tsconfig.json
```diff
@@ -1,22 +1,23 @@
...
-    "target": "es2015",
-    "lib": ["dom", "dom.iterable", "esnext"],
     "allowJs": true,
-    "skipLibCheck": true,
-    "esModuleInterop": true,
     "allowSyntheticDefaultImports": true,
-    "strict": true,
+    "baseUrl": "./src",
+    "esModuleInterop": true,
+    "esModuleInterop": true,
```
emccorson, line 7: Not added by your PR but the sorting shows up a duplicate `esModuleInterop` key we should probably remove at some point. I'm wondering too if it would be a good idea to have a shared tsconfig in the future that each package can extend, but not sure.

emccorson (approved):
Looks good, and the schemas are much cleaner! Left a couple of minor comments.

apps/namada-interface/tsconfig.json
mateuszjasiuk: Ha did not notice that :D Thanks! Yeah one config makes sense

packages/types/src/tx/types.ts
```diff
@@ -20,15 +20,15 @@ export type TxProps = {
   feeAmount: BigNumber;
   gasLimit: BigNumber;
   chainId: string;
-  publicKey?: string;
+  publicKey: string | undefined;
```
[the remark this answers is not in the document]
mateuszjasiuk: Nice! Good to know! I will replace string | undefined with ? as discussed :)

packages/types/src/tx/schema/bond.ts
emccorson: I think we could actually use `TxMsgValue` to construct plain objects since TypeScript just sees classes as another type:

```
const props: TxMsgValue = {
  token: "token",
  feeAmount: new BigNumber(0),
  gasLimit: new BigNumber(0),
  chainId: "chainId",
};
```

This will typecheck just fine, but I don't know if that's intuitive behaviour :smile:

packages/types/src/tx/schema/tx.ts
```diff
@@ -17,9 +18,9 @@ export class TxMsgValue {
   chainId!: string;
 
   @field({ type: option("string") })
-  publicKey!: string | undefined;
+  publicKey?: string;
 
-  constructor(data: TxMsgValue) {
+  constructor(data: TxProps) {
```
emccorson, line 23: As mentioned in other comment, I think these are actually the same type.

```typescript
type IsSame =
  TxMsgValue extends TxProps
    ? TxProps extends TxMsgValue
      ? true
      : false
    : false;

const isSame: IsSame = true;
```

But it might make sense to keep both to distinguish between plain object and class, I'm not sure.

emccorson (approved):

packages/types/src/tx/schema/tx.ts
mateuszjasiuk: Ah yes sorry I forgot to answer that one :D So with typescript it is fine, but then what happens is that borsh.ts reads `data.constructor` which has to be `TxMsgValue`. Otherwise it "can't find" the schema to parse data.
So we could use only TxMsgValue, but we need to remember about setting prototype again after deserialization.

jurevans (approved):
LGTM!

packages/types/src/tx/schema/tx.ts
emccorson: Ahh right I see. It seems `InstanceType<typeof TxMsgValue>` will just resolve to `TxMsgValue` though, and doesn't check the constructor. The `InstanceType` way still allowed me to pass in a plain object that is not an instance of `TxMsgValue`.

``` typescript
class Child {}

class Parent {
  child: InstanceType<typeof Child>;

  constructor(child: Child) {
    if (!(child instanceof Child)) {
      throw new Error("uh oh");
    }
    console.log("all good");
    this.child = child;
  }
}

new Parent(new Child()); // works fine
new Parent({}); // throws error at runtime
```

There doesn't look like a way to check the constructor at compile-time in TypeScript, so maybe there is nothing we can do here.

mateuszjasiuk: Yeah I guess what we could do is remove props and change fields in schemas to private? :D This way we can force object construction using constructor xd I think what we have is good for now, Justin mentioned that he wants to remove nested "tx" in schemas anyway :P

PR #326, opened 2023-07-07
refactor: use wasm-pack for builds

mateuszjasiuk:
`wasm-pack` has built in support for `wasm-opt`, so it makes sense to use it for builds
I've also changed:
- opt levels a bit, not sure if it helps with anything though :D 
- test build profile to `dev` as it does skip optimizations


jurevans (approved):
Nice!

PR #327, opened 2023-07-07
chore - Update to SDK changes in namada v0.18.1

jurevans:
This PR updates us to support `namada` `v0.18.1`, which contains the refactoring of the SDK calls.

__NOTE__ We can now pass an optional `publicKey` along with our transactions, so this has been added to the `tx` schema and `TxProps` type, but isn't yet being passed in. It is required to be there so tx can be deserialized in Rust.


mateuszjasiuk (approved):
Tested and LGTM!

PR #331, opened 2023-07-10
fix: calculate toast props from toast type

emccorson:
This fixes a problem where TypeScript was not throwing an error when getToast was called with the wrong arguments. The code calling getToast has been commented out for now.

~This PR is based on #325 since they both touch the same file, so this should be rebased to main after #325 is merged.~ Done.

### DIFF
--- a/apps/namada-interface/src/slices/transfers.ts
+++ b/apps/namada-interface/src/slices/transfers.ts
@@ -2,10 +2,10 @@ import { createAsyncThunk, createSlice } from "@reduxjs/toolkit";
 import BigNumber from "bignumber.js";
 
 import { Account, Tokens, TokenType, Signer } from "@namada/types";
+import { assertNever } from "@namada/utils";
 import { getIntegration } from "@namada/hooks";
 
 import {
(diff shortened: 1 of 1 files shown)


apps/namada-interface/src/slices/transfers.ts
```diff
@@ -2,10 +2,11 @@ import { createAsyncThunk, createSlice } from "@reduxjs/toolkit";
 import BigNumber from "bignumber.js";
 
 import { Account, Tokens, TokenType, Signer } from "@anoma/types";
+import { assertNever } from "@anoma/utils";
 import { getIntegration } from "@anoma/hooks";
 
 import {
-  actions as notificationsActions,
+  //actions as notificationsActions,
```
mateuszjasiuk, line 9: Comment!

mateuszjasiuk (approved):
Looks good to me. Left a comment about a comment!

apps/namada-interface/src/slices/transfers.ts
emccorson: Fixed!

PR #336, opened 2023-07-11
feat/335 - Approve Bond Tx, Sign bond with Ledger

### LINKED ISSUE #335: Approve Bond Tx with Ledger signing

jurevans:
Resolves #335 

This PR adds Bond Tx to the approval process, and will allow user to sign a Bond Tx with their Ledger.

- [x] Submitting a Bond for the first time requires and prompts user to approve a Reveal Public Key tx
- [x] Submitting a Bond Tx prompts user for approval
- [x] Approving a Bond Tx prompts for password (if using a private key)
- [x] Approving a Bond Tx provides Ledger prompts (if using Ledger account)
- [x] Rejecting a Bond Tx removes pending transaction
- [x] Bond is successfully submitted through SDK
- [x] Bond is successfully signed with the Ledger keys, and submits through the SDK

**Nice-to-have**

- [ ] Submitting a bond gives the user the appropriate notification in interface (e.g., "Bond is submitting" or something)
  - _TODO_ This is partially hooked up, but there still remains an issue with updating amount staked. Token balance _after_ bonding does currently work

## NOTES

- This PR includes a built copy of the `ledger-namada` package in `packages/`, which contains the necessary changes on the client-side to parse Ledger responses, so we are no longer using the `@zondax/ledger-namada` package.

# Testing

### Setup
2. Ping me for a copy of the installer script you need for your Ledger, then install with `./installer_s2.sh load` (e.g., if you're using a Ledger Nano S-Plus)

### Testing Bond tx
1. In the interface, select either your initial account, or ledger account, and click the `Staking` menu item (top menu)
2. You will see a link to a validator, so click that
3. From this view, click `Stake` (there is an account drop down, but you don't need to change this if the account selected has funds, but you can as an option)
4. In the modal window, enter an amount you would like to stake, then click `Confirm`
5. In the popup window, click to confirm the transaction:
  - If you selected a local wallet account, you will be prompted for a password to unlock your key and sign the transaction
  - If you selected Ledger, you will see the following:
     1. The first time you use your Ledger account, you will be prompted on the device to approve a reveal public key tx (go ahead and click through to approve!)
     2. Following that, you will need to also approve the `Bond` tx. Once this is approved, you should see a success message once it completes


apps/extension/src/Approvals/ApproveBond/ConfirmLedgerBond.tsx
```diff
@@ -0,0 +1,181 @@
...
+  address,
+  msgId,
+  publicKey,
+}) => {
+  const requester = useRequester();
+  const [error, setError] = useState<string>();
+  const [status, setStatus] = useState<Status>();
+  const [statusInfo, setStatusInfo] = useState("");
+
+  const revealPk = async (ledger: Ledger, publicKey: string): Promise<void> => {
```
jurevans, line 45: NOTE: In my next PR, I will make this a top-level function for _any_ Ledger tx, as it applies to all tx signed with the Ledger (will refactor `Approvals` a bit to make this possible, and handle the different tx types more elegantly)

apps/extension/src/Approvals/ApproveBond/ConfirmLedgerBond.tsx
```diff
@@ -0,0 +1,181 @@
...
+        Ports.Background,
+        new SubmitSignedRevealPKMsg(msgId, toBase64(bytes), signatures)
+      );
+    } catch (e) {
+      console.warn(e);
+      throw new Error(`${e}`);
+    }
+  };
+
+  const queryPublicKey = async (
```
jurevans, line 82: Same with this function - this should apply to all Ledger tx and not need to be duplicated.

PR #339, opened 2023-07-20
feat/311 - Implement ledger integration improvements

### LINKED ISSUE #311: Improvements to Ledger functionality
Relates to #253, and addresses tasks not covered by #295 

We can improve and extend the functionality of the Ledger integration by implementing the following:

### Refactoring
- [x] Refactor Approvals so that we don't have to duplicate the "reveal public-key" functionality - this applies to any transaction signed with the Ledger, but _only_ with the Ledger (not needed for private-key signed tx)
- [x] Look into: Would it make sense to split `Tx` (with `TxProps`) arguments from, say, `SubmitTransferTx`, `SubmitBondTx`, etc.? This way, any Rust method that requires generic `Tx` args would not need a specific method for each Tx type, which is currently needed to deserialize those arguments.
  - _Update_ This is a massive update. For now, I'm deserializing this message to retrieve the original `TxProps`, then serializing that for submission to the single `submit_signed_tx` method

### Ledger transaction types

We should be able to sign all of the following transaction types at this point:

- [x] `reveal_pk`
- [x] `transfer`
- [x] `bond`
- [x] `ubond`
- [x] `withdraw`

### Etc

- [ ] Ensure bond/unbond amounts are fixed
  - Are they even broken, or do we just need to indicate that they will be submitted in Namnam versus Nam units?

jurevans:
Resolves #311 

- [x] Submitting a `transfer` from Ledger account works
- [x] Submitting an `unbond` from Ledger account works
- [x] Submitting a `withdraw` from Ledger account works
- [x] Consolidate approvals into single flow (`ApproveTx`)
- [x] Existing transactions work in new approval flow
- [x] Implement improvements to Rust lib (consolidate different Tx submissions into `submit_signed_tx`)
- [x] Clean up unused code (unused messages, arguments, etc.)

_NOTE_ I did a _little_ bit of clean up outside the scope of this PR (discovered some out-of-date and unused code)

# Testing

### Setup
2. Install the Ledger app with `./installer_s2.sh load` (e.g., if you're using a Ledger Nano S-Plus) (Ping me for a copy of the installer script you need for your Ledger)


apps/extension/src/background/approvals/service.ts
```diff
@@ -21,89 +24,131 @@ export class ApprovalsService {
...
-      id,
+      msgId,
       source,
-      token,
+      tokenAddress,
       amount: amount.toString(),
-      publicKey: publicKey || "",
-      type: type as string,
+      publicKey,
+      accountType: type as string,
```
emccorson, line 88: I think `type as string` can be changed to `type` and the behaviour will be the same, since `type` is a non-optional parameter.

apps/extension/src/background/approvals/service.ts
```diff
@@ -21,89 +24,131 @@ export class ApprovalsService {
...
     const url = paramsToUrl(baseUrl, {
-      id,
+      msgId,
       source,
       target,
-      token,
+      tokenAddress,
       amount: amount.toString(),
-      type: type as string,
+      accountType: type as string,
```
emccorson, line 55: If `type` is undefined, this will be read in ApproveTx as the string `"undefined"`. I'm not sure if that matters though. By the way, what are the cases of transfers where `type` would be undefined here?

apps/extension/src/hooks/useQuery.tsx
```diff
@@ -4,15 +4,26 @@ import { sanitize } from "dompurify";
...
-      sanitize(unsanitized);
-  }
+    return unsanitized === null ? unsanitized : sanitize(unsanitized);
+  },
+  getAll: (): Record<string, string | null> => {
+    const values: Record<string, string | null> = {};
+
+    urlSearchParams.forEach((val, key) => {
+      const unsanitized = val;
+      values[key] = unsanitized === null ? unsanitized : sanitize(unsanitized);
```
emccorson, line 22: I'm wondering if we should also be sanitizing the key here, since it could be set by the user.

apps/namada-interface/.prettierrc
```diff
@@ -2,5 +2,6 @@
   "trailingComma": "es5",
   "tabWidth": 2,
   "semi": true,
-  "singleQuote": false
+  "singleQuote": false,
+  "bracketSpacing": false
```
emccorson, line 6: I noticed `"bracketSpacing": false` also got set in `apps/extension/.prettierc` in one commit but removed again in another commit. Is it supposed to be added here?

emccorson (approved):
Left a couple of comments, and was not able to test the Ledger functionality due to not having one, but other than that looks good!

apps/extension/src/Approvals/ApproveTx/ConfirmTx.tsx
```diff
@@ -0,0 +1,170 @@
...
+    setStatus(Status.Pending);
+    setStatusInfo(
+      `Decrypting keys and submitting ${TxTypeLabel[txType as TxType]}...`
+    );
+
+    try {
+      if (!msgId) {
+        throw new Error("msgId was not provided!");
+      }
+      switch (txType) {
```
mateuszjasiuk, line 56: This switch will probably grow right?
I was thinking that maybe instead of the switch we can have a map of TxTypes into Msgs, as those messages have the same constructor parameters. So something like this:
```ts
type SupportedTx = Extract<
  TxType,
  TxType.Bond | TxType.Unbond | TxType.Transfer | TxType.Withdraw
>;

type ApproveMsg = new (msgId: string, password: string) => unknown &
  Message<void>;

const txMap: Map<SupportedTx, ApproveMsg> = new Map([
  [TxType.Bond, SubmitApprovedBondMsg],
  [TxType.Unbond, SubmitApprovedUnbondMsg],
  [TxType.Transfer, SubmitApprovedTransferMsg],
  [TxType.Withdraw, SubmitApprovedWithdrawMsg],
]);
```
We could use them like this:
```ts
const Msg: ApproveMsg | undefined = txMap.get(txType as SupportedTx);

if (Msg) {
    await requester.sendMessage(Ports.Background, new Msg(msgId, password));
}
```
And `SupportedTx` can be removed once we support all of those. 
Not sure if it's more readable, just an idea :)

mateuszjasiuk (approved):
Looks good to me. Left couple of comments. I will test a bit later!

packages/shared/lib/src/sdk/mod.rs
```diff
@@ -188,57 +176,91 @@ impl Sdk {
...
+                    &mut self.wallet,
+                    &mut self.shielded_ctx,
+                    args.clone(),
+                )
+                .await
+                .map_err(JsError::from)?;
+                transfer.0
+            }
+            TxType::Unbond => {
+                let args = tx::unbond_tx_args(tx_msg, None)?;
```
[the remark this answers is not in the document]
jurevans: Thanks! I noticed that as well, but this is helpful!

apps/namada-interface/.prettierrc
jurevans: I should probably remove this as well. It fixed annoying formatting issues in one case, but introduced another.

apps/extension/src/Approvals/Approvals.tsx
```diff
@@ -11,23 +12,29 @@ import {
...
+import { ConfirmTx } from "./ApproveTx/ConfirmTx";
+import { ApproveTx } from "./ApproveTx/ApproveTx";
 
 export enum Status {
   Completed,
   Pending,
   Failed,
 }
 
+export type ApprovalDetails = {
```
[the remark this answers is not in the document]
emccorson: I missed this in my review, but looking at the definition of `SanitizedURLSearchParams.prototype.getAll`, the call to `forEach` won't ever return `null` (I think), so you could actually remove the `null`s or change them to `undefined` and use the built-in `Partial` (which is the same idea as Mateusz's `Nullable`).

apps/extension/src/Approvals/ApproveTx/ConfirmTx.tsx
jurevans: I like this approach better, and this may also be used to clean up one of my utilities.

apps/extension/src/background/approvals/service.ts
jurevans: Thanks! This was an error from an earlier version of that handler I hadn't fixed yet. `type` is required

packages/shared/lib/src/sdk/mod.rs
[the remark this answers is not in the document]
jurevans: This fixed the issue, thanks!

apps/extension/src/background/approvals/service.ts
jurevans: I removed these, thanks!

apps/extension/src/hooks/useQuery.tsx
jurevans: I updated this to only return properties that have associated values, and am sanitizing the key as well. Thanks!

apps/extension/src/Approvals/ApproveTx/ConfirmTx.tsx
jurevans: This has been added, replacing the `switch`

apps/extension/src/Approvals/ApproveTx/ApproveTx.tsx
```diff
@@ -0,0 +1,116 @@
...
+    setDetails({
+      source,
+      txType,
+      msgId,
+      publicKey,
+      target,
+    });
+  }, [source, publicKey, txType, target, msgId]);
+
+  const handleApproveClick = (): void => {
```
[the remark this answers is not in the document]
jurevans: Added, thanks!

apps/extension/src/Approvals/Approvals.tsx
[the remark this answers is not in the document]
jurevans: I updated this to ignore any property that doesn't have an associated value

PR #346, opened 2023-07-27
feat/342 - Ledger - support for additional addresses

### LINKED ISSUE #342: Add support for multiple Ledger addresses
### Handling Ledger accounts

- [x] User can specify any number of BIP-044 addresses (similar to what normal mnemonic-based accounts can do)
- [x] User can remove the connection to Ledger, removing all related addresses

jurevans:
Resolves #342 

This PR enables address derivation from a user's Ledger hardware wallet. The initial account when connecting a Ledger uses path `/0'/0/0`, so the address form should start the user at `/0'/0/1`, otherwise behavior is the same as deriving from a mnemonic account. _Note_ that paths such as `m/44'/877'/0'/0` are not supported on Ledger, i.e., the path should always contain the `index` of the path as well.

- [x] User can derive and track additional aliases/addresses/public keys from Ledger wallet
- [x] User can successfully sign any supported Tx with these new accounts
- [x] User can remove Ledger connection from wallet
- [x] Disable "Reset password" for Ledger accounts (not needed)
- [x] Update to `v0.20.1`
- [x] Handle Ledger error states (e.g., Locked device, etc.) - these should be presented to user

### DIFF
--- a/apps/extension/src/App/Accounts/AccountListing.tsx
+++ b/apps/extension/src/App/Accounts/AccountListing.tsx
@@ -62,6 +62,7 @@ const AccountListing = ({ account, parentAlias }: Props): JSX.Element => {
   const { address, alias, path, type } = account;
   const navigate = useNavigate();
   const isChildAccount = isChild(type, path);
+  const isLedgerParent = type === AccountType.Ledger && !isChildAccount;
 
   return (
     <AccountListingContainer>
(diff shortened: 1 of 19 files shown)


mateuszjasiuk (approved):
Tested, works nicely! Looks good to me!

PR #350, opened 2023-08-02
Improve extension errors

### LINKED ISSUE #139: Bug: Extension error messages not formatted correctly
When a promise fails to resolve using the `browser.sendMessage` API, the returned error messages are improperly formatted, and missing relevant data.

- Currently there are situations where errors returned via a `Message` are logged as `Error[Object, object]`
  - Take a look at: https://github.com/anoma/namada-interface/blob/8b455e546b24cab4920599b0bbb58067b68a0928/apps/extension/src/extension/ExtensionRequester.ts#L35-L40 where `result.error` (which is supposed to be a string) seems to be converted to an object (not 100% sure that is what is happening, but probably a good place to start)
  - See the `Result` type in https://github.com/anoma/namada-interface/blob/8b455e546b24cab4920599b0bbb58067b68a0928/apps/extension/src/extension/ExtensionMessenger.ts#L14-L17C3
- See how we can return errors from WebWorkers (not sure exactly how this will be done, but we might have to return a message indicating an error has occurred, or broadcast an event we need to listen for)

Relates to issue: https://github.com/anoma/namada-interface/issues/123

emccorson:
This PR fixes a bug where background errors were getting logged as `[Object, object]` in Chrome, and adds source maps for the extension on Firefox in development mode. These changes should make error reporting in the extension better.

I wanted to add source maps for Chrome, but the type of source map I chose uses `eval`, which is not supported in manifest v3. There are non-eval source maps, but these crashed webpack when I tried them.

---

### Changed

- Print background errors properly on Chrome (83792a18)

### Added

- Add extension source maps on Firefox (8890336d)

### DIFF
--- a/apps/extension/src/extension/ExtensionMessenger.ts
+++ b/apps/extension/src/extension/ExtensionMessenger.ts
@@ -12,7 +12,7 @@ export type Callback = (
 ) => Promise<RouterResult | void>;
 
 export type Result<M> = {
-  error?: string;
+  error?: { message: string, stack: string };
   return: Promise<M extends Message<infer R> ? R : never>;
 };
 
--- a/apps/extension/src/extension/ExtensionRequester.ts
+++ b/apps/extension/src/extension/ExtensionRequester.ts
@@ -33,7 +33,10 @@ export class ExtensionRequester {
     }
 
     if (result.error) {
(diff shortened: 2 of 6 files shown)


jurevans (approved):
Nice, this looks good!

PR #359, opened 2023-08-08
fix: show web worker errors in extension console

emccorson:
This PR make sure errors thrown in web workers running the submit-transfer-web-worker.js script are logged properly in the extension console. The errors are passed from the web worker to the offscreen document and then to the background script where they are logged to the console. This fix should only affect Chrome.


jurevans (approved):
Nice!

PR #360, opened 2023-08-08
chore: update namada to draft/0.22

mateuszjasiuk:


packages/shared/lib/src/sdk/mod.rs
```diff
@@ -106,53 +110,47 @@ impl Sdk {
...
+        let pk = &signing_data
+            .public_keys
+            .clone()
+            .into_iter()
+            .nth(0)
+            .expect("No public key provided");
+
+        let address = Address::from(pk);
+
+        //TODO: add cehck for faucet
```
jurevans, line 132: I think we can remove this comment

jurevans (approved):
Looks good to me!

PR #363, opened 2023-08-10
bug/353 - Ledger error-handling improvments

### LINKED ISSUE #353: [Bug] Ledger Reveal PK TX Rejection/Error does not break flow
When the user submits a transaction using a Ledger account for the first time, they are required to approve and submit a "reveal PK" transaction. Currently, it is not handled if the user _rejects_ that transaction, and following rejection will immediately try to execute the original transaction, which shouldn't happen.

- [x] If user rejects a Reveal PK, it should break out of the flow, and reset, displaying the rejection response to the user
- [x] If an error occurs (such as if the device is locked), this should also break that flow correctly (it will currently break the flow, but clicking "Try Again" does not hit the reveal PK step again successfully, meaning if the following tx is attempted, it will fail on chain)
- [x] Look into handling errors such as when the App itself is not loaded and thus the JS client is unable to initialize. Take a look at the `LedgerErrors` enum, and compare the results of the status query in each of these cases so we can handle them appropriately

This may take a little bit of time to determine the best way to handle all potential Ledger states, and could potentially be improved across the board.

jurevans:
Resolves #353 

Any time an error is encountered with the Ledger device, the extensions should always allow the user to correct the issue and proceed. A locked device or unloaded app should not force the user to refresh! Also, the appropriate error should be reported to user in each case!

- Connect Ledger
  - [x] Recovers from device locked
  - [x] Recovers from app not loaded
  - [x] Recovers from device not plugged in, or device not initially unlocked (same error: "Access Denied")
- Add Account
  - [x] Recovers from device locked
  - [x] Recovers from app not loaded
  - [x] Recovers from device not plugged in, or device not initially unlocked (same error: "Access Denied")
- Reveal PK
  - [x] Does not proceed to next transaction if any errors are encountered (including tx rejection)
- Submit Tx (Reveal PK, Bond, Unbond, Withdraw, Transfer)
  - [x] Recovers from device locked
  - [x] Recovers from app not loaded
  - [x] Recovers from transaction rejected
  - [x] Recovers from device not plugged in, or device not initially unlocked (same error: "Access Denied")

### DIFF
--- a/apps/extension/src/App/Accounts/AddAccount.tsx
+++ b/apps/extension/src/App/Accounts/AddAccount.tsx
@@ -219,6 +219,8 @@ const AddAccount: React.FC<Props> = ({
   };
 
   const addLedgerAccount = async (): Promise<DerivedAccount | void> => {
+    setFormStatus(Status.Pending);
+
     const bip44Path = {
       account: parentAccount.path.account,
       change,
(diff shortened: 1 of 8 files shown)


mateuszjasiuk (approved):
Tested! Looks good to me :)

PR #372, opened 2023-08-25
feat/347 - Faucet transfers in Interface

### LINKED ISSUE #347: Faucet transfers from extension
Opening this issue to discuss how we might bring back faucet transfers the interface (or now, extension), as that would simplify testing for users when connected to a devnet.

We originally had a check in the interface when running `yarn dev:local` that would automatically fund accounts with 1000 tokens (NAM/ETH/BTC/etc). It might be better to simply have an additional UI button and Amount form available when connected to a devnet. As transactions are executed in the extension, it would make sense that they are initiated there, and could dispatch an event to update balances in the interface (if it is currently connected).

Also, as we need to add balances to the Extension accounts overview, it would be more clear that an account has been funded if it was immediately apparent in the extension itself.

__NOTE__ We could just use the presence of a env variable for faucet address to determine if we display a fund via faucet form to users - would be the simplest approach.

jurevans:
Resolves #347 

- [x] If `REACT_APP_NAMADA_FAUCET_ADDRESS` is defined in `.env` _AND_ a Namada chain is selected, display a `Get testnet tokens` button (for transparent accounts) that links to a faucet token form
- [x] If no `REACT_APP_NAMADA_FAUCET_ADDRESS` is defined in `.env`, user should _not_ see `Get testnet tokens`
- [x] __Ledger accounts__: Submitting faucet transfer should use approval process for accessing the keys of the _target_, which should sign the transaction
- [x] __Transparent accounts__: Submitting faucet transfer should use approval process for accessing the keys of the _target_, which should sign the transaction (_in progress_ - we need to update the signing for this in our SDK wrapper in order to sign with the correct account. Currently the `keypair` look up fails because it's looking up `source`, which in this case is the faucet address!)
- [x] Faucet transfer amount input should  be validated against `REACT_APP_NAMADA_FAUCET_LIMIT`, which defaults to `1000`, and should match the faucet transfer limit defined in the testnet genesis toml

### Testing

- Set up a local node (`v0.20.1`)
- Run `namadaw address list` to get the address for `faucet`
- Add `faucet` address to `apps/namada-interface/.env` like so (note that this is an established address, and as such will be different for each chain you set up):
  ```
  REACT_APP_NAMADA_FAUCET_ADDRESS=atest1v4ehgw368qurzd69xfpnysenggenqwpeg9rr23jr8q6ygvzxgvenx3jpx4rrsv2pgyervsejnxeng7
  # OPTIONAL - DEFAULTS TO 1000 (this is set in the genesis file)
  REACT_APP_NAMADA_FAUCET_LIMIT=1000
   ```
- Start the extension and interface as usual, and for both Ledger and Transparent Accounts (extension private key-based) you should see `Get testnet tokens` - clicking that should launch a modal window where you can enter an amount, then you will need to approve/authenticate as per usual to submit the transaction

### Notes

- This works for all tokens that we have definitions for in the interface (`NAM`, `ATOM` (which is really `Apfel`), and `ETH`, however, if you load `ETH` from the faucet, you will notice an impressively large number of zeros after your expected balance.
  - e.g.: `123` becomes `123000000000000` - this only happens for ETH, so perhaps there is some logic we need to incorporate to present this correctly - that can be fixed in a future PR


mateuszjasiuk (approved):
Tested both ledger and transparent and looks good!

PR #373, opened 2023-08-28
chore: remove borales/actions-yarn from CI

emccorson:
This fixes an error where the CI would not run due to having uses and run keys in the same job.


jurevans (approved):
I think this looks good! Like you said, we can revert if it doesn't work correctly

mateuszjasiuk (approved):
LGTM! :)

PR #375, opened 2023-08-28
feat/374 - Hook up IBC to Approval and Ledger flows

### LINKED ISSUE #374: IBC - Hook up to Ledger & Approvals
We need to add IBC Transfer tx to our current approval process, and enable for Ledger (currently unsupported, but it should be in place)

- [x] User is able to submit an IBC Transfer and be prompted to approve it and decrypt their keys
- [x] User is able to submit an IBC Transfer to be approved and signed on Ledger

jurevans:
Resolves #374 

_Note 1_: As a part of this PR to update the Bridge page for IBC transfers, I did a bit of refactoring to remove the `transfers` slice, as that part of the state was going completely unused.

_Note 2_: The submission of IBC is currently disabled at the end of the submission process, so we can follow through the approval flow, but `v0.20.1` has some issues (will crash for some reason), so this will be best enabled and tested in `0.22.x`

_Note 3_: I left Bridge transfer unchanged (it is currently not properly hooked up) to focus on IBC

- [x] IBC Transfers follow the Approvals flow and can be submitted (_Note_ Without a proper IBC set up, i.e., with multiple chains and a relayer running, these tx will not be applied)
- [x] IBC transfers work with Ledger approvals
  - Tx submission is disabled for Ledger, but the signing process should work as a _Custom_ tx (though the Ledger is unable to parse information from it currently)

### Testing

This assumes you have the Keplr extension installed!

- Create a mnemonic account in extension, or connect to Ledger
- Load funds into account via faucet
- Navigate to the `Bridge` tab
- Select the account+token that has a balance
- Click to add a channel ID (e.g., `channel-0`)
- Enter recipient, or you can also click `Load Accounts from Keplr extension` and it will load the recipient address
- Specify a valid amount, then click `Submit` - You should be prompted to approve (and authenticate or unlock Ledger). Note that currently approving will not actually submit the tx! You should instead get a "Success" notification. Ledger approvals will need to also approve on the device (it is currently parsed as a `Custom` tx, so you won't see any details on the device, as it cannot parse any details)


packages/shared/lib/src/sdk/mod.rs
```diff
@@ -235,6 +236,17 @@ impl Sdk {
...
             }
+            TxType::IBCTransfer => {
+                let args = tx::ibc_transfer_tx_args(tx_msg, None)?;
+                let transfer = namada::ledger::tx::build_ibc_transfer(
+                    &self.client,
+                    &mut self.wallet,
+                    args.clone(),
+                )
+                .await
+                .map_err(JsError::from)?;
```
mateuszjasiuk, line 247: In many cases you can just do `await?` without map_err. The reason is that JsError implements From\<std::error::Error\>
```rs
    impl<E> From<E> for JsError
    where
        E: std::error::Error,
    {
        fn from(error: E) -> Self {
            JsError::new(&error.to_string())
        }
    }
```

mateuszjasiuk (approved):
Tested and it looks good to me. Left minor comment about map_err :)

PR #377, opened 2023-08-29
test: derive transparent and derive shielded address e2e tests

mateuszjasiuk:
Added tests for deriving addresses


jurevans (approved):
LGTM!

PR #383, opened 2023-09-07
WIP - feat/380 - IBC Updates

### LINKED ISSUE #380: IBC - Updates, testing
This issue it to track updates to IBC functionality in interface & extension

- [ ] Balances can be queried for Keplr accounts
- [ ] IBC is enabled and can be submitted to Namada (Transparent accounts & Ledger)
- [ ] _TBD_

jurevans:
resolves #380


apps/namada-interface/src/App/Token/types.ts
```diff
@@ -47,8 +47,9 @@ export type TxTransferArgs = TxArgs & {
   faucet?: string;
 };
 
-export type TxIbcTransferArgs = TxArgs & {
+export type TxIbcTransferArgs = Omit<TxArgs, "token"> & {
   chainId: string;
   channelId: string;
   portId: string;
+  token: TokenInfo
 };
```
jurevans, line 55: Note for the above: For IBC, I'm intending to pass `TokenInfo` instead of `token` (which is a `string` representing an address), as for Keplr, we don't want the address, but rather the denom ("uatom"), so I omit the `string` version in favor of the `TokenInfo` version - this is a little weird, but as we are likely to refactor token handling in general, this will probably not be around for too long.

packages/types/src/tx/tokens/Cosmos.ts
```diff
@@ -1,8 +1,22 @@
...
+
+export const tokenByMinDenom = (minDenom: CosmosMinDenom): string => {
+  const tokenDenom = tokenDenomLookup(minDenom)
+  return tokenDenom ? tokenDenom[1] : ""
+}
+
+export const minDenomByToken = (token: CosmosTokenType): string => {
+  const tokenDenom = tokenDenomLookup(token)
+  return tokenDenom ? tokenDenom[0] : "";
+}
```
jurevans, line 22: This way of handling Cosmos tokens may change significantly when token handling is refactored - This is just a way to map between the denom and min-denom (ATOM & uatom) for now

mateuszjasiuk (approved):
Tested and works nice! :)

PR #384, opened 2023-09-07
Update for ledger-namada multisig changes

jurevans:
With the multisig changes accounted for in the `ledger-namada` app, we need to update how we are retrieving and constructing signatures from the device.

- [x] Ledger can sign tx from `v0.22.0`
- [x] Tx is valid when submitted to chain


mateuszjasiuk (approved):
LGTM! Tested and works with no probs!

PR #387, opened 2023-09-13
docs: sdk usage

mateuszjasiuk:


specs/packages/sdk.md
```diff
@@ -44,3 +44,36 @@ Sdk struct wraps logic from the other modules, creating public API using `wasm_b
...
+shared.d.ts
+shared.js
+shared_bg.js
+shared_bg.wasm
+shared_bg.wasm.d.ts
+```
+You can take a look at `shared.d.ts` to see what is exported.
+
+Next step would be to import SDK itself.
+For now we wrap all of the SDK logic in the [Sdk struct](https://github.com/anoma/namada-interface/blob/main/packages/shared/lib/src/sdk/mod.rs) and [Query struct](https://github.com/anoma/namada-interface/blob/main/packages/shared/lib/src/query.rs). Main reason we do that is to not pass things like `HttpClient`, `Wallet` and `ShieldedContext` every time we want to submit new transaction or query data. This will most likely change in future as we are sometimes getting "multiple mutable references" Rust error.
```
jurevans, line 65: > Main reason we do that is to not pass things like `HttpClient`, `Wallet` and `ShieldedContext` every time we want to submit new transaction or query data. This will most likely change in future as we are sometimes getting "multiple mutable references" Rust error.

I would maybe reword this to something like the following (but you may want to check my accuracy ;) ):

(_In regards to Sdk & Query structs_)
```
Creating instances of the `Sdk` and `Query` structs allows us to persist instances of `HttpClient`, `Wallet`, and `ShieldedContext` for the duration of the service-worker lifetime, making them immediately available to any calls invoked by the extension. Currently, the `namada` SDK requires that these references be _mutable_, which presents an issue when attempting to execute SDK functionality in parallel. In the SDK's current implementation, calls should be made sequentially. 
```

specs/packages/sdk.md
```diff
@@ -44,3 +44,36 @@ Sdk struct wraps logic from the other modules, creating public API using `wasm_b
 
 At the moment, the client of the Sdk is responsible for keeping storage in sync with the Sdk's wallet instance. What does that mean? The Namada extension has its own key storage, which is modified by user actions. Every time the extension storage changes (for example, when a new address is derived), we must manually sync the extension and Sdk wallet storage
 using functions provided by `wallet` module. This is something that will most likely change in the future.
+
+## How can you import and use the SDK?
+First you need to build the shared package. We do that using `wasm-pack`.
```
jurevans, line 49: I would maybe update this as follows (very minor differences):

First, we need to compile the [shared](https://github.com/anoma/namada-interface/tree/main/packages/shared/lib) package to a _wasm_ target, which is done using [wasm-pack](https://rustwasm.github.io/wasm-pack/).

specs/packages/sdk.md
```diff
@@ -44,3 +44,36 @@ Sdk struct wraps logic from the other modules, creating public API using `wasm_b
...
+LICENSE_MIT
+package.json
+README.md
+shared.d.ts
+shared.js
+shared_bg.js
+shared_bg.wasm
+shared_bg.wasm.d.ts
+```
+You can take a look at `shared.d.ts` to see what is exported.
```
jurevans, line 62: > You can take a look at `shared.d.ts` to see what is exported.

The file `shared.d.ts` contains the exported TypeScript type definitions.

specs/packages/sdk.md
```diff
@@ -44,3 +44,36 @@ Sdk struct wraps logic from the other modules, creating public API using `wasm_b
...
+
+Next step would be to import SDK itself.
+For now we wrap all of the SDK logic in the [Sdk struct](https://github.com/anoma/namada-interface/blob/main/packages/shared/lib/src/sdk/mod.rs) and [Query struct](https://github.com/anoma/namada-interface/blob/main/packages/shared/lib/src/query.rs). Main reason we do that is to not pass things like `HttpClient`, `Wallet` and `ShieldedContext` every time we want to submit new transaction or query data. This will most likely change in future as we are sometimes getting "multiple mutable references" Rust error.
+
+To use structs (almost)all you have to do is to create new instance:
+```ts
+import { Query, Sdk } from "@namada/shared";
+
+const sdk = new Sdk(URL_OF_NAM_NODE);
```
jurevans, line 71: Maybe this is a little more specific:

```
const sdk = new Sdk(RPC_URL_OF_NAM_NODE);
const query = new Query(RPC_URL_OF_NAM_NODE);
```

specs/packages/sdk.md
```diff
@@ -44,3 +44,36 @@ Sdk struct wraps logic from the other modules, creating public API using `wasm_b
...
+shared_bg.js
+shared_bg.wasm
+shared_bg.wasm.d.ts
+```
+You can take a look at `shared.d.ts` to see what is exported.
+
+Next step would be to import SDK itself.
+For now we wrap all of the SDK logic in the [Sdk struct](https://github.com/anoma/namada-interface/blob/main/packages/shared/lib/src/sdk/mod.rs) and [Query struct](https://github.com/anoma/namada-interface/blob/main/packages/shared/lib/src/query.rs). Main reason we do that is to not pass things like `HttpClient`, `Wallet` and `ShieldedContext` every time we want to submit new transaction or query data. This will most likely change in future as we are sometimes getting "multiple mutable references" Rust error.
+
+To use structs (almost)all you have to do is to create new instance:
```
jurevans, line 67: > To use structs (almost)all you have to do is to create new instance:

There's a missing space, but I also think we could reword this to get rid of it entirely:

> Within the extension, we create instances of `Sdk` and `Query` as follows:

specs/packages/sdk.md
```diff
@@ -44,3 +44,36 @@ Sdk struct wraps logic from the other modules, creating public API using `wasm_b
...
+```ts
+import { Query, Sdk } from "@namada/shared";
+
+const sdk = new Sdk(URL_OF_NAM_NODE);
+const query = new Query(URL_OF_NAM_NODE);
+```
+Why almost?
+Query is ready to use. When it comes to the Sdk though there is one more resposibility for the client. This might also change in future. Basically client is responsible for:
+- persisting, encoding and decoding of [wallet](https://github.com/anoma/namada-interface/blob/main/packages/shared/lib/src/sdk/wallet.rs) state
+- adding keys to the [wallet](https://github.com/anoma/namada-interface/blob/main/packages/shared/lib/src/sdk/wallet.rs) struct
```
jurevans, line 77: I might changes this (Starting at `Why almost?`) to (also, check whether this is correct!):

**NOTE**: Regarding this instance of `Sdk`, it is the client's responsibility to persist the wallet state in storage, as well as to populate the wallet state with keys pulled from that storage. The `Sdk` class has methods for exporting and restoring state, and adding keys:

- `sdk.encode()` - Encode the SDK state for storage within the extension
- `sdk.decode()` - Decode the SDK state for loading into current app
- `sdk.add_key()` - Add key to SDK state
- `sdk.add_spending_key()` - Add MASP spending key to SDK state

The overall public API of the `Sdk` struct is as follows:

```ts
/**
* Represents the Sdk public API.
*/
export class Sdk {
  free(): void;
  constructor(url: string);
  static has_masp_params(): Promise<any>;
  static fetch_and_store_masp_params(): Promise<void>;
  load_masp_params(): Promise<void>;
  encode(): Uint8Array;
  decode(data: Uint8Array): void;
  clear_storage(): void;
  add_key(private_key: string, password?: string, alias?: string): void;
  add_spending_key(xsk: string, password: string | undefined, alias: string): void;
  submit_signed_reveal_pk(tx_msg: Uint8Array, tx_bytes: Uint8Array, raw_sig_bytes: Uint8Array, wrapper_sig_bytes: Uint8Array): Promise<void>;
  build_tx(tx_type: number, tx_msg: Uint8Array, gas_payer: string): Promise<any>;
  submit_signed_tx(tx_msg: Uint8Array, tx_bytes: Uint8Array, raw_sig_bytes: Uint8Array, wrapper_sig_bytes: Uint8Array): Promise<void>;
  submit_transfer(tx_msg: Uint8Array, password?: string, xsk?: string): Promise<void>;
  submit_ibc_transfer(tx_msg: Uint8Array, password?: string): Promise<void>;
  submit_bond(tx_msg: Uint8Array, password?: string): Promise<void>;
  submit_unbond(tx_msg: Uint8Array, password?: string): Promise<void>;
  submit_withdraw(tx_msg: Uint8Array, password?: string): Promise<void>;
}
```

It is important to note that these many of these methods expect a Borsh-serialized message representing the transaction (`Uint8Array`). For an overview of the schemas used to define these messages, see [schema.ts](https://github.com/anoma/namada-interface/tree/main/packages/types/src/tx/schema). The extension uses [borsh-ts](https://github.com/dao-xyz/borsh-ts) to encode these messages. To use these schema in your TypeScript app, they must be imported from the `@namada/types` package.

specs/packages/sdk.md
```diff
@@ -44,3 +44,36 @@ Sdk struct wraps logic from the other modules, creating public API using `wasm_b
...
+
+To get a better view look for `sdk.decode`, `sdk.encode`, `sdk.add_key`, `sdk.add_spending_key` calls in the `@namada/extension` app.
```
jurevans, line 79: We could get rid of this line if we include the ^^^ above overview

jurevans (approved):
LGTM!

PR #394, opened 2023-09-18
feat/393 - Add scripts to start local CORS proxies

### LINKED ISSUE #393: Interface: Add CORS proxy to local dev scripts
This adds the ability to launch local CORS proxies for connecting to chain RPCs.

1. Add a global dependency for `local-cors-proxy`: `npm i -g local-cors-proxy`
2. Run either:
  - `yarn start:proxy` to launch the proxies in a terminal, or
  - `yarn dev:proxy` to launch both the proxies _and_ the interface in development mode

jurevans:
Resolves #393 

This PR introduces scripts to proxy requests to chains in order to circumvent CORS errors when developing locally.

This script looks at the `apps/namada-interface/.env` file, collects the variables, and for each chain specified launches a proxy with a unique URL/PORT, which maps to the chain configs and is used when `REACT_APP_PROXY` is defined.

Running `yarn dev:proxy` is equivalent to:

```bash
yarn start:proxy
REACT_APP_PROXY=true; yarn dev:local
```

## Testing

- In `namada-interface`, run either:
  - `yarn start:proxy`, or:
  - `yarn dev:proxy`
  (The former only starts proxies based on your `.env` config. The latter runs the same script, but also launches the interface in development mode. In practice, we'll generally use `yarn dev:proxy`)
- In `extension`, if `yarn dev:proxy` is running, run either:
  - `yarn start:chrome:proxy`, or:
  - `yarn start:firefox:proxy`

When testing against `namada`, make sure _not_ to update the allowed origins for CORS in the config! It will work without that setting, and for any other chain (i.e., Cosmos)


emccorson (commented):
Looks good! I just have a couple of comments:

- Should the extension also use the proxy URL? It looks like it is just the interface at the moment.
- You could add `local-cors-proxy` as a dev dependency to skip the manual global install. This actually wouldn't require changing the node script because running the script through yarn/npm adds dev dependencies to the path (which I was surprised to find out).

jurevans:
> 
>     * Should the extension also use the proxy URL? It looks like it is just the interface at the moment.
> 

Yeah, you're right, the extension should also have scripts with the `REACT_APP_PROXY` env set, for avoiding Namada cors issues. I was thinking mostly of Cosmos, etc., but should cover all of the Namada requests too, since allowing all CORS origins is a pain every time!

jurevans:
> Yeah, you're right, the extension should also have scripts with the `REACT_APP_PROXY` env set, for avoiding Namada cors issues. I was thinking mostly of Cosmos, etc., but should cover all of the Namada requests too, since allowing all CORS origins is a pain every time!

That said, we can only invoke the start proxy scripts from one or the other, so for the sake of simplicity, I'll leave that part on the interface. I should also update the `README.md`!

emccorson:
It looks like the extension still uses `127.0.0.1:27657` even with the second commit.
I think it's because of this line (but not 100% sure):
https://github.com/anoma/namada-interface/blob/a26bfabc5f5bd83bb6a46036fdc2259b2a7fa218/apps/extension/src/background/index.ts#L72

Other than that, changes look good!

jurevans:
> It looks like the extension still uses `127.0.0.1:27657` even with the second commit. I think it's because of this line (but not 100% sure):
> 
Yep, you're right! I'll double check all URL instances this time.

emccorson (approved):

PR #398, opened 2023-09-26
chore: update namada to 0.23.0

mateuszjasiuk:
For now I've used  DefaultIo struct  which does nothing for us.
We can replace it with our IO  to print some logs to the console.
FIxed e2e tests


jurevans (approved):
Looks good to me! Tested against 0.23.0 - looks like we need a Ledger update, but not relevant to this PR

PR #399, opened 2023-09-28
feat: strict storage durablity

### LINKED ISSUE #349: Extension storage should be guaranteed durable
The extension storage that stores keys should be ensured to be durable (IndexedDB).

- [ ] In Chrome, we need to force the "strict" durability option for IndexedDB (we could confirm that it is "strict" by default, but perhaps we still _enforce_ it should defaults ever change)
- [ ] In Firefox, set "strict" durability option during transaction

Firefox caveat (see https://developer.mozilla.org/en-US/docs/Web/API/IDBTransaction/durability):

> Note: In Firefox, if you wish to ensure durability for some reason (e.g. you're storing critical data that cannot be recomputed later) you can force a transaction to flush to disk before delivering the complete event by creating a transaction using the experimental (non-standard) readwriteflush mode (see [IDBDatabase.transaction](https://developer.mozilla.org/en-US/docs/Web/API/IDBDatabase/transaction).) This is currently experimental, and can only be used if the dom.indexedDB.experimental pref is set to true in about:config.

### Resources
- <https://developer.mozilla.org/en-US/docs/Web/API/IDBTransaction#firefox_durability_guarantees>
- <https://developer.mozilla.org/en-US/docs/Web/API/IDBTransaction/durability>

mateuszjasiuk:
Resolves #349

### In this PR:
- we set transaction durability option to "strict"
- for firefox we use "readwriteflush" mode if possible. if not we display info icon with warning
### Testing:
In chrome storage should be durable by default,  so there should be no visible info icon

In firefox you should see yellow info icon with information about storage durability. Follow the steps from the icon title. Open extension popup again - info should be gone.

### Caveats
- this solution checks durability every time user opens the popup
- it sets durability options globally(for every storage we use),  we might want to change this in the future if it turns out to be slow

### DIFF
--- a/apps/extension/src/App/App.components.ts
+++ b/apps/extension/src/App/App.components.ts
@@ -66,6 +66,7 @@ export const Heading = styled.h1`
 `;
 
 export const HeadingLoader = styled.div`
+  display: flex;
   position: relative;
   width: 20px;
   height: 20px;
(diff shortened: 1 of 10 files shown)


jurevans (approved):
Looks good to me!

PR #401, opened 2023-09-28
feat/400 - Set up basic Faucet form App

### LINKED ISSUE #400: Set up Faucet API front-end
We need a basic front-end to connect to the Faucet API. It should be configurable & easy to build+deploy.

Refer to https://github.com/heliaxdev/namada-faucet/blob/master/scripts/faucet.py for an example of using this API.

In addition to a form UI, we also need to perform some mining logic to find a solution to a tagged challenge returned by API.

Form should have sanitation, as well as bech32m validation for the target field to prevent invalid (or potentially malicious) data from reaching the server. Validation should provide meaningful feedback to the user.

jurevans:
resolves #400 

- [x] User can enter a target, select a token, and enter an amount to request a faucet transfer
- [x] Form is validated for valid `amount`, and `target` is both sanitized and bech32m-validated
- [x] Error will display if transfer failed
- [x] If transfer is successful, form resets, and user is shown the JSON output of the transfer
- [x] Banner displays if app is accessed before testnet launch: 5 Oct 2023 17:00 UTC - this disables submissions if accessed before then

### Notes

The logic in `utils` surrounding the compute POW, etc., is based on [faucet.py](https://github.com/heliaxdev/namada-faucet/blob/master/scripts/faucet.py).

### Screenshots

Successful transfer:
Screen Shot 2023-10-03 at 10 38 13 AM

Pending testnet launch banner:
Screen Shot 2023-10-04 at 10 29 58 AM

Valid form `Submit` is disabled if banner is displayed (`isTestnetLive === false`):
Screen Shot 2023-10-05 at 5 49 05 AM


mateuszjasiuk (approved):
Code looks good! Also tested with testnet and works nice!

PR #404, opened 2023-10-04
test(e2e): bond, unbond, withdraw, add to bridge pool

mateuszjasiuk:
How to run tests:
- go to e2e
- just in case go to `e2e/.namada` and change `.version` file to v0.22.0 - this will retrigger all the downloads during first run
- `./setup-frontend.sh` and wait for both extension and interface  to start
- `yarn test` or` yarn test:watch` - first run will be slow as it will download bunch of stuff like: namada binaries, wasms, masp params
- wait for tests to finish(hopefully successfully :))

I've tested it on WSL/ubuntu. Pretty sure it should work on Mac too.


jurevans (approved):
Looks good to me!

PR #407, opened 2023-10-09
Un-nest common tx args and refactor tx code

emccorson:
This PR refactors the transfer and SDK code to un-nest the common TX args and separate TX submission into separate build, sign, and process steps, as well as remove duplicate code and fix a bug in the query timeout message.

Notes:
- I've tested transparent->transparent transfers, bond, unbond and withdraw. I couldn't test the other transaction types since I don't have the local setup/hardware.
- I don't particularly like the name `specificMsg` to refer to transaction type-specific args (e.g. transfer args, bond args), so suggestions welcome for anything better.

---

### Changed
- Separate SDK submit functions into build functions, sign_tx function and process_tx function
- Make Ledger build_tx function reuse non-Ledger build_ functions
- Rename Ledger sign_tx function to append_signature
- Remove Tx arg from schemas and pass Tx as separate argument
- Replace Props types with type alias to equivalent MsgValue types
- Replace separate submit and approve functions/messages with single Tx function/message covering all supported Tx types
- Move SupportedTx type to @namada/types

### Fixed
- Fix wrong timeout value being printed in query error message

### DIFF
--- a/apps/extension/src/Approvals/ApproveTx/ConfirmTx.tsx
+++ b/apps/extension/src/Approvals/ApproveTx/ConfirmTx.tsx
@@ -9,6 +9,7 @@ import {
 } from "@namada/components";
 import { shortenAddress } from "@namada/utils";
 import { TxType, TxTypeLabel } from "@namada/shared";
+import { SupportedTx } from "@namada/types";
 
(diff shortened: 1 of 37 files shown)


jurevans (approved):
Code looks good, tested this locally against the current testnet, and tested with ledger, works great!

PR #409, opened 2023-10-10
Fix web-ext failing to install on CI

emccorson:
Adds global install of node-gyp before trying to install web-ext, which should fix the failure when trying to release the extension.

 I don't know why this works but it seems to.

I couldn't run this in this repo since it needs to be merged to main first, but it succeeded in my fork:
https://github.com/emccorson/namada-interface/actions/runs/6463924139


jurevans (approved):
Nice!

PR #415, opened 2023-10-12
bug/414 - Fix staking for testnet

### LINKED ISSUE #414: Bug: Staking page is not usable on testnet
The staking page is currently set up to issue a call for all validator addresses, and iterates with a nested query to fetch total bonded for that validator. This is causing the interface to hang as it processes a large number of validators.

One potential solution is to paginate the listing of all validators, and then only query total bonded for the currently visible page. We may also want to cache these values in storage, and only update once the query completes the next time.

jurevans:
resolves #414 

- [x] All validator addresses display on Staking Overview
- [x] Total bond can be fetched for these addresses and displayed
- [x] Implement pagination into this view, as we can have 1000+ results, and that is quite slow performance just to display total bonded - filter should work with paginated results
- [x] Debounce user input (search filter) as otherwise, the performance is sluggish and issues far too many requests

__NOTE__ There is still some temp data, like Commission, that isn't being utilized. I left it in there, but also added into the `validatorAssets` state, as I didn't want to have to refactor everything in this PR. This is an area that could be cleaned up further and simplified. Keeping the state split like this makes handing updates in the UI more smooth IMHO, as there are many queries happening async.

### Testing

Update `namada-interface/apps/namada-interface/.env` to include this configuration:

```bash
REACT_APP_NAMADA_CHAIN_ID=public-testnet-14.5d79b6958580
REACT_APP_NAMADA_URL=https://proxy.heliax.click/public-testnet-14.5d79b6958580/
```

Run interface with:
```bash
yarn dev:proxy
```

Run extension with:
```bash
yarn start:chrome:proxy
```

- Existing staking functionality should work
- Validators from testnet should load, with the results paginated

_NOTE_ I'm suddenly having issues with timeouts on the testnet, regardless of what branch I test on. If you want to see the staking with `My Validators` populated, you can restore with this mnemonic, which has bonds posted:
```
ribbon carpet winner accuse describe lamp pluck noble supply chaos patch inherit
```
### Screenshots

Screen Shot 2023-10-16 at 11 33 20 AM


apps/namada-interface/src/App/Staking/ValidatorDetails/ValidatorDetails.tsx
```diff
@@ -117,19 +129,23 @@ const validatorToDataRows = (
...
     {
-      uuid: "4",
-      key: "Description",
-      value: validator.description,
+      uuid: "3",
+      key: "Voting Power",
+      value: validator.votingPower?.toString() ?? "",
     },
-    { uuid: "5", key: "Website", value: validator.homepageUrl },
+    // {
```
mateuszjasiuk, line 143: Comments!

mateuszjasiuk (approved):
Tested and code looks good. There is some unused code commented. Also it would be nice to change 127.0.0.1 to localhost in proxy configuration :)

mateuszjasiuk (approved):

PR #417, opened 2023-10-12
Updating README with a few missing steps

pedrorezende:
I've just ran the installation process on my end and noticed a few missing bits in the README file. Updating it for future reference.

### DIFF
--- a/README.md
+++ b/README.md
@@ -46,7 +46,7 @@ sudo apt-get install -y libssl-dev
 sudo apt-get install -y protobuf-compiler
 
 # Install wasm-pack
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -201,6 +201,9 @@ yarn build:firefox
 Within `apps/namada-interface/`, we may issue the following commands:
 
 ```bash
+# Make sure to build wasm dependencies before starting development server
+yarn wasm:build
+
```
jurevans, line 206: I think originally it was assumed that the user already ran `yarn wasm:build` in the previous step (under extension - compiles the same lib that's also needed on interface), in which case they wouldn't need to here, but I think it's perfectly fine to also include it here, as it is required!

PR #419, opened 2023-10-16
Fix extension CI build

emccorson:
This PR fixes the release builds of the extension for Firefox and Chrome in the CI.

The Firefox build was broken because eval source maps were being added in the production build, but eval is not allowed in the production CSP.

The Chrome build was broken on the CI only due to an old version of wasm-pack. This caused some problem with the code webpack was generating to load `crypto_bg.wasm` and `shared_bg.wasm`, but I'm not sure exactly what.

I tested the release on my fork:
https://github.com/emccorson/namada-interface/releases/tag/v0.1.0-2d30252

---

### Fixed
- Stop eval source maps being added in Firefox production build
- Use newer version of wasm-pack and add to dev dependencies


jurevans (approved):
This looks good!

PR #428, opened 2023-10-31
Gas fees pt. 1

emccorson:
This PR introduces some improvements for gas fee handling and the token send form.

The main changes are that gas price is now fixed in the token send form, and that shielded transactions will unshield the gas fee from the source account. The amount input of the form will take into account the gas fee when determining what amount is valid.

It is helpful to have long epochs when testing shielded transfers because they take a very long time (5 minutes+ for me).

I am planning to add further gas fee improvements later, such as the max button to use all available funds.

---

### Changed
- Use disposable gas payer to pay for shielded transactions
- Replace gas price buttons with fixed minimum gas price
- Rename `feeAmount` to `gasPayer` and make optional

### Added
- Query minimum gas price
- Add new input component for BigNumbers

### Removed
- Remove unused faucet transfer form
- Remove unused QR code parts of token send form

### Fixed
- Take gas fee into account when determining invalid transaction amount


emccorson:
Closing as replaced by #576.

PR #434, opened 2023-11-06
feat/182 - Private key derivation updates

### LINKED ISSUE #182: Implement the usage of Namada SDK for Account derivation
Subtask of https://github.com/anoma/namada-interface/issues/177

Perform the required changes in the client (Extension/Wallet) and the Rust wrapper for using Accounts from the new SDK.
* Creating account address from a seed and derivation path

jurevans:
resolves #182 

This PR is to make use of Namada types & functions for key gen to align with process used in SDK.

- [x] Update `Mnemonic` to use `tiny-bip39`
- [x] Update `HDWallet` to use `slip10_ed25519`
- [x] Update tests
- [x] Verify that derived keys in extension from a given mnemonic are equivalent to those created in CLI

__NOTES__

- `publicKey` is now being used from the `Address` type, and removed from the Key returned in `HDWallet`
- `HDWallet.derive()` now returns a `Key` type, and not `Keypair` (only private key)
- Updated to Namada v0.25.0
- BIP44 Path now requires `index` - type updated. To align with CLI, we derive using all path components including `index`. As `slip10_ed25519` derivation requires an array of path indices, there is a new helper to create this array from a `coinType` and `path`.

### DIFF
--- a/apps/extension/src/background/keyring/keyring.ts
+++ b/apps/extension/src/background/keyring/keyring.ts
@@ -37,7 +37,7 @@ import {
   UtilityStore,
   AccountStore,
 } from "./types";
-import { Result } from "@namada/utils";
+import { Result, makeBip44PathArray } from "@namada/utils";
 
 import { Crypto } from "./crypto";
 import { getAccountValuesFromStore, generateId } from "utils";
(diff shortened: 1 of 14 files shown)


mateuszjasiuk (approved):
Looks good! Tested manually, both extension and CLI produce the same addresses!

PR #436, opened 2023-11-08
Single password for the extension

pedrorezende:
According to the latest discussions, we've decided to move towards the direction of having one master password to unlock the extension. In order to make it in a more consistent way, this PR aims to separate some of the methods inside the KeyRing into a new abstraction called `VaultService`.

The Vault will be responsible for everything related to lock / unlock the extension, and sensitive data management. Some background messages were also changed, being part of the `vault` module now.


jurevans (approved):
I've tested the extension locally, everything works great! Once CI passes, this should be good to go!

PR #440, opened 2023-11-10
Simplifying code with Context API

pedrorezende:
Implementing a tiny context for `accounts` management, reusing calls and state along the extension code. Also reimplements the way we reuse route names, to avoid many concatenations and string replacements in routes like `/account/view/:accountId` . 

In general, this PR simplifies a bit the code and also improves a the performance of some areas, like `ViewAccount`


apps/extension/src/App/Accounts/ViewAccount/ViewAccount.tsx
```diff
@@ -1,99 +1,58 @@
...
-      setParentAccount(parentAccount);
 
-      accounts.forEach((account) => {
-        if (
-          account.type === AccountType.Mnemonic ||
-          account.type === AccountType.Ledger
-        ) {
-          setTransparentAddress(account.address);
-        }
+  const searchAccounts = (accountId: string): void => {
```
mateuszjasiuk, line 26: Nitpick: I think it might be a bit clearer  to return a value from this and maybe set state in the  effect :)

apps/extension/src/App/AppContent.tsx
```diff
@@ -110,6 +89,7 @@ export const AppContent = ({
   }, []);
 
   useEffect(() => {
+    console.log(">>>", getStartPage(accounts));
```
mateuszjasiuk, line 92: loglog

mateuszjasiuk (approved):
Tested and everything seems to be working ok!
There is this error thrown when opening extension:
```
No routes matched location "/"
```
Also added two small comments. But in general LGTM! :)

apps/extension/src/App/AppContent.tsx
pedrorezende: 💩 Ops. Fixed

PR #443, opened 2023-11-14
Setup: Removing return button from last step

pedrorezende:
When users reached the end of the Setup flow (confirmation), they could return to the previous step. This PR removes this possibility and also ensures that the seedPhrase gets deleted before moving forward to the next step.


jurevans (approved):
Looks good!

PR #444, opened 2023-11-14
Properly displaying validation messages on Password form

pedrorezende:
Password fields were displaying `zxcvbnFeedback.suggestions` only as a hint, but they're being caught by validation. This PR shows these suggestions as errors instead.


mateuszjasiuk (approved):
LGTM! :)

PR #446, opened 2023-11-15
feat: Add ability to import an account from a private key

emccorson:
Notes:
- Private key validation currently checks that the key is not more than 64 characters. I'm not sure if that should be *exactly* 64 characters.
- I'm not 100% sure I'm doing the right thing in the replacement of `selectedSeedPhrase` which is supposed to be used to allow deleting the seed phrase from `seedPhrase ` in `Setup.tsx`.

### DIFF
--- a/apps/extension/src/background/keyring/keyring.ts
+++ b/apps/extension/src/background/keyring/keyring.ts
@@ -34,9 +34,10 @@ import {
   KeyRingStatus,
   SensitiveAccountStoreData,
   UtilityStore,
+  AccountSecret,
 } from "./types";
 
-import { Result, makeBip44PathArray } from "@namada/utils";
+import { Result, makeBip44PathArray, assertNever } from "@namada/utils";
 
 import { VaultService } from "background/vault";
 import { generateId } from "utils";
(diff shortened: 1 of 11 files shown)


apps/extension/src/background/keyring/messages.ts
```diff
@@ -141,23 +146,27 @@ export class RenameAccountMsg extends Message<DerivedAccount> {
...
   }
 
   validate(): void {
-    if (!this.words) {
-      throw new Error("A wordlist is required to save a mnemonic!");
+    if (!this.accountSecret) {
+      throw new Error("A wordlist or private key is required!");
     }
 
     if (
```
pedrorezende, line 166: Do you think we should validate here for the privateKey as well?

pedrorezende (approved):
LGTM! Just a small comment on message validation

apps/extension/src/background/keyring/messages.ts
jurevans: We could at least check that it's 64 bytes, which is what our private keys are, though I'm not sure if it could be less than that...

apps/extension/src/background/keyring/keyring.ts
```diff
@@ -169,23 +170,54 @@ export class KeyRing {
...
+
+            return { sk, text: phrase, accountType: AccountType.Mnemonic };
+
+          case "PrivateKey":
+            const { privateKey } = accountSecret;
+
+            return {
+              sk: privateKey,
+              text: privateKey,
+              accountType: AccountType.PrivateKey
```
jurevans, line 213: Yeah this makes sense to me! That tells us what we need to know if we are to sign with this account.

jurevans (approved):
LGTM!

PR #451, opened 2023-11-15
Another set of fixes for Extension

pedrorezende:
This PR contains a few fixes spotted during another round of tests: 
- Attempting to fix `Error: Cannot read properties of undefined (reading 'password')` error
- Changing height of textarea that shows mnemonic according to its size
- Removing all user data after all the keys are removed
- Interaction on Clipboard button


jurevans (approved):
Changes look good! I tested several flows with extension and Ledger

PR #454, opened 2023-11-16
feat/453 - SDK/Ledger Signing Updates - Support v0.28.0

### LINKED ISSUE #453: Extension: Ledger Signing Updates
We need to test Ledger signing with the new Ledger NamadaApp based on `v0.27.0` with the changes involving tx type by tag.

See: https://github.com/anoma/namada/pull/2182

Use tag `v0.27.0` for testing on Namada and SDK.

- This involves a bit of refactoring of the SDK store, as it is currently not managed well with recent updates, making it impossible to query keys in other Tx, as well as the disabling of the `verification_key` to avoid encoding issues. That needs to be re-enabled for Ledger, and the encoding needs to be corrected
- It might be possible to remove the SDK store altogether, as we're already storing everything we need, if we can rework the signing/build-tx calls to work simply with a private key. Ledger Tx avoid the key lookup by specifying a `verification_key` - one re-enabled, we should take a look at doing that for all Tx to see if the SDK store can be avoided.

jurevans:
Resolves #453 

- [x] Test Ledger signing, update as necessary
- [x] SDK storage needs to be removed, and signing steps updated to provide the keys from extension storage

### Notes

The SDK refactor included the following:
- Remove the `sdkStore`, as we can pass in an array of keys to the `signing_keys` Tx args (currently only providing a single key, but it is possible to pass in more than one in the future)
- Remove references to `parentId` for Ledger functionality
- Remove `addKeys` from SDK wallet
- Remove `password` from all SDK calls
- As a transfer _must_ be submitted now with a private key, I've added a type `SigningKey` which allows us to handle both private key and spending key, so we can validate spending key in `build_transfer` (as before), and now sign with a specified private key in `sign_tx`. For now, I am appending the `xsk` to the SDK wallet instance for look up there, as I don't know how else it can be applied

### Testing

Checkout Namada tag `v0.28.0`

```bash
make build-release
make build-wasm-scripts
```

_NOTE_ You can use the following account data to set default balances:

```
# mnemonic
liar bird install win wool venue observe maid flock clap bullet myth illness trip bread fresh polar smart use lunar tired embody come deer

# derived account data
secret: 59f0d21f1c7375697275d8fe6ef58b06505f5a073aeca5c5b05e0dda24bd6324
address: tnam1qz4sdx5jlh909j44uz46pf29ty0ztftfzc98s8dx
publicKey: tpknam1qptrn64myunqr4847yq4cn0uwek5ecwc7eeexjfc5npmd5kmg6ex563n5as
publicKeyHash: AB069A92FDCAF2CAB5E0ABA0A545591E25A56916
```

In `genesis/localnet/src/pre-genesis/wallet.toml`:

```toml
[secret_keys]
tester-key = "unencrypted:0059f0d21f1c7375697275d8fe6ef58b06505f5a073aeca5c5b05e0dda24bd6324"

[public_keys]
tester-key = "ED25519_PK_PREFIXtpknam1qptrn64myunqr4847yq4cn0uwek5ecwc7eeexjfc5npmd5kmg6ex563n5as"

[addresses]
tester-key = "tnam1qz4sdx5jlh909j44uz46pf29ty0ztftfzc98s8dx"

[pkhs]
AB069A92FDCAF2CAB5E0ABA0A545591E25A56916 = "tester-key"
```

In `genesis/localnet/src/balances.toml`:

```toml
[token.NAM]
# tester-key
tpknam1qptrn64myunqr4847yq4cn0uwek5ecwc7eeexjfc5npmd5kmg6ex563n5as = "1000000"

[token.BTC]
# tester-key
tpknam1qptrn64myunqr4847yq4cn0uwek5ecwc7eeexjfc5npmd5kmg6ex563n5as = "1000000"

[token.ETH]
# tester-key
tpknam1qptrn64myunqr4847yq4cn0uwek5ecwc7eeexjfc5npmd5kmg6ex563n5as = "1000000"

[token.DOT]
# tester-key
tpknam1qptrn64myunqr4847yq4cn0uwek5ecwc7eeexjfc5npmd5kmg6ex563n5as = "1000000"
```

Check that these accounts appear:

```bash
cargo run --bin namadaw -- --base-dir "genesis/localnet/src" key list
```

Configure a local chain:

```bash
NAMADA_BASE_DIR=localnet namadac utils init-network --chain-prefix localnet --genesis-time "2023-01-01T00:00:00Z" --templates-path genesis/localnet --wasm-checksums-path wasm/checksums.json

# take note of Chain ID to use below:

NAMADA_NETWORK_CONFIGS_DIR=. namadac utils join-network --chain-id $CHAIN_ID --genesis-validator validator-0 --pre-genesis-path genesis/localnet/src/pre-genesis/validator-0 --dont-prefetch-wasm

cp -R wasm/*.wasm /path/to/Namada/$CHAIN_ID/wasm/
cp wasm/checksums.json /path/to/Namada/$CHAIN_ID/wasm/

namada ledger run
```

Make sure to run `yarn` in `namada-interface` base folder, as I have upgraded the `ledger-namada` package to use the published package from Zondax, and removed our local copy.

Successfully tested the following Tx types (including with Ledger device):
- Transfer (transparent, transparent->shielded) _NOTE_ We need to update `gas_payer` value for shielded->shielded, as that is currently broken
- Bond
- Unbond
- Withdraw
- Reveal PK

_NOTE_ Currently, submitting a Transparent->Shielded Tx succeeds, as far as the interface is concerned, no errors are thrown, but for some reason they don't seem to be applied. The signing in that case should be identical to a Transparent->Transparent (source account signs the Tx), so not sure what the issue could be (perhaps my local chain is not correctly configured for MASP?)


apps/extension/src/background/keyring/keyring.ts
```diff
@@ -630,16 +599,59 @@ export class KeyRing {
...
+      await this.vaultService.reveal<SensitiveAccountStoreData>(account);
+    if (!sensitiveProps) {
+      throw new Error(`Signing key for ${address} not found!`);
+    }
+    const { text: phrase } = sensitiveProps;
+    const mnemonic = Mnemonic.from_phrase(phrase);
+    const seed = mnemonic.to_seed();
+    const hdWallet = new HDWallet(seed);
+    const key = hdWallet.derive(new Uint32Array(bip44Path));
+    const privateKeyStringPtr = key.to_hex();
```
mateuszjasiuk, line 635: I think you should free the pointer at some point right? So private key zeroizes.

apps/extension/src/background/keyring/types.ts
```diff
@@ -60,3 +60,8 @@ export type MnemonicValidationResponse = {
   isValid: boolean;
   error?: string;
 };
+
+export type SigningKey = {
+  pk?: string;
```
mateuszjasiuk, line 65: it would be nice to rename to privateKey or secretKey, so it's not confused with publicKey :)

packages/shared/lib/Cargo.toml
```diff
@@ -16,15 +16,15 @@ dev = []
 [dependencies]
 async-trait = {version = "0.1.51"}
 tiny-bip39 = "0.8.2"
-borsh = {version = "1.0.0-alpha.4", features = ["schema", "derive"]}
-borsh-ext = {tag = "v1.0.0-alpha.4", git = "https://github.com/heliaxdev/borsh-ext"}
+borsh = {version = "1.2.0", features = ["unstable__schema", "derive"]}
+borsh-ext = { git = "https://github.com/heliaxdev/borsh-ext", tag = "v1.2.0" }
```
mateuszjasiuk, line 20: We should change borsh deps and masp_primitives in crypto package also, just to be sure :) I've noticed that when I was trying to reduce bundles size.

packages/shared/lib/src/sdk/wallet.rs
```diff
@@ -105,21 +37,15 @@ pub fn add_key(
...
 /// * `xsk` - String representing serialized ExtendedSpendingKey.
-/// * `password` - Password to encrypt the spending key.
 /// * `alias` - Spending key alias.
-pub fn add_spending_key(
-    wallet: &mut Wallet<BrowserWalletUtils>,
-    xsk: &str,
-    password: Option<String>,
-    alias: &str,
-) {
+pub fn add_spending_key(wallet: &mut Wallet<BrowserWalletUtils>, xsk: &str, alias: &str) {
```
mateuszjasiuk, line 41: Do we still need this?

mateuszjasiuk (approved):
Looks good! :)  Left few comments.

packages/shared/lib/src/sdk/wallet.rs
jurevans: Currently, yes, as I am still invoking it for transfers from a shielded account. I'm not sure to remove this, but I'm taking a look along with the shielded tx error in another issue

packages/shared/lib/Cargo.toml
jurevans: Done!

apps/extension/src/background/keyring/keyring.ts
jurevans: Done!

PR #456, opened 2023-11-17
CI fixes and yarn bump to v4

emccorson:
This PR fixes the release CI workflow by bumping yarn to v4, as well as a few other minor CI changes/fixes.

---

### Changed
- Upgrade yarn to v4 and track yarn in the repo
- Add `web-ext` to `devDependencies`; there is no need to install it globally now
- Remove `export` from yarn scripts as it is not supported by yarn's shell
- Use `webpack-cli` instead of `webpack` because yarn can't find `webpack` for some reason

### Added
- PRs will cause the CI to build the extension and make the built extensions available as artifacts in the CI run summary

### Fixed
- Fix CI warnings about deprecated actions/checkout@v2


jurevans (approved):
This looks good!

PR #461, opened 2023-11-17
feat/460 - Add Ledger approval/rejection handling

### LINKED ISSUE #460: Extension: Ledger  Import - Update to prompt for approval to import address/public key
This issue is tentative - we may not want to add this, but in case we do, the import process for Ledger should look like this:

- [ ] When user selects to import keys from Ledger, they are notified to approve on their Ledger
- [ ] Upon approving on Ledger, they are directed to set an alias (and password if one isn't set)
- [ ] Upon rejection, display message to user, and allow them to try again if they choose

jurevans:
resolves #460 

Adds Ledger approval & rejection handling for connecting to wallet.

### Pending state
Screen Shot 2023-11-17 at 10 59 07 AM

### Error state (rejected)
Screen Shot 2023-11-17 at 10 59 27 AM

### DIFF
--- a/apps/extension/src/background/ledger/ledger.ts
+++ b/apps/extension/src/background/ledger/ledger.ts
@@ -90,6 +90,30 @@ export class Ledger {
     };
   }
 
+  public async showAddressAndPublicKey(
+    bip44Path?: string
+  ): Promise<{ address: string; publicKey: string }> {
+    if (!this.namadaApp) {
+      throw new Error("NamadaApp is not initialized!");
+    }
+
+    const path = bip44Path || DEFAULT_LEDGER_BIP44_PATH;
+
+    try {
+      const { address, publicKey } = await this.namadaApp.showAddressAndPubKey(
+        path
+      );
+
+      return {
+        // Return address as bech32-encoded string
+        address: address.toString(),
+        // Return public key as hex-encoded string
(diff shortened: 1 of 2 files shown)


apps/extension/src/background/ledger/ledger.ts
```diff
@@ -90,6 +90,30 @@ export class Ledger {
     };
   }
 
+  public async showAddressAndPublicKey(
```
pedrorezende, line 93: Nitpick: once we’re not triggering showing / displaying anything, wouldn’t be better to have this method named as retrieveAddressAndPublicKey? Or getAddressOrPublicKey?

pedrorezende (approved):
Small comment, but LGTM!

apps/extension/src/background/ledger/ledger.ts
jurevans: we already have `getAddressAndPublicKey`, which simply retrieves the key. I'm leaving that one in there as it might be used in the signing process still (trying not to break anything)

I can rename it to `retrieveAddressAndPublicKey`!

jurevans: I'm gonna make the call to update this in the future. I'll add this to https://github.com/anoma/namada-interface/pull/454, that way I can remove the other call as well

PR #462, opened 2023-11-17
feat: remove shared dep from types

mateuszjasiuk:


emccorson (approved):
Looks good to me!

I'm not sure what we can do in the future about the `any` in `TxMsgProps`. I can't see anything in the wasm-bindgen docs about importing an enum from TypeScript to Rust, so if the type has to be defined on the Rust side, then maybe the shared package should have some Rust code too. But I don't know :laughing:

PR #463, opened 2023-11-17
feat/455 - Extension: API to return default account

### LINKED ISSUE #455: Extension: Should have API to return selected account
The extension should provide an API to query the selected account in the extension. We should also provide an associated event with this information (currently, changing and account only broadcasts an event that the accounts have changed, but this should be update to simply return the selected account instead)

- [x] Extension API should support querying only the selected account (currently only supports returning all accounts)

jurevans:
resolves #455 

## Testing

1 - Install this version of the extension
2 - Launch `namada-interface`:
  ```bash
  # In namada-interface/apps/namada-interface
  yarn dev:local
  ```
3 - Create/Import at least 2 accounts in extension (to test that default account gets updated)
4 - In `namada-interface`, connect to extension, then open the console, and enter the following:
```
await namada.defaultAccount()
```

This should give you a result like:
```
{
    "id": "9f318429-3627-50e2-b5f5-26f861b00d13",
    "alias": "tester",
    "address": "tnam1qz4sdx5jlh909j44uz46pf29ty0ztftfzc98s8dx",
    "owner": "tnam1qz4sdx5jlh909j44uz46pf29ty0ztftfzc98s8dx",
    "chainId": "localnet.aa8b11f87ddde352c2-0",
    "path": {
        "account": 0,
        "change": 0,
        "index": 0
    },
    "publicKey": "tpknam1qptrn64myunqr4847yq4cn0uwek5ecwc7eeexjfc5npmd5kmg6ex563n5as",
    "type": "mnemonic"
}
```

You can switch the selected account in the extension, then re-run `await namada.defaultAccount()` in the console, and the account should be updated.

## Events

I had wanted to make a specific event that fires when the default accounts changed, but wanted to keep this simple for the release. For now, we can subscribe to `Events.AccountChanged`, and re-issue the `await namada.defaultAccount()` if we need event handling for any reason.

### DIFF
--- a/apps/extension/src/background/keyring/keyring.ts
+++ b/apps/extension/src/background/keyring/keyring.ts
@@ -75,7 +75,7 @@ export class KeyRing {
     protected readonly sdk: Sdk,
     protected readonly query: Query,
     protected readonly cryptoMemory: WebAssembly.Memory
-  ) { }
+  ) {}
 
   public get status(): KeyRingStatus {
     return this._status;
@@ -188,9 +188,9 @@ export class KeyRing {
 
     try {
       const { sk, text, accountType } = ((): {
-        sk: string,
-        text: string,
-        accountType: AccountType
+        sk: string;
+        text: string;
+        accountType: AccountType;
       } => {
         switch (accountSecret.t) {
           case "Mnemonic":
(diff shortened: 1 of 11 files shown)


apps/extension/src/background/keyring/keyring.ts
```diff
@@ -575,6 +578,15 @@ export class KeyRing {
     return [];
   }
 
+  public async queryDefaultAccount(): Promise<DerivedAccount | undefined> {
+    const accounts = await this.queryAllAccounts();
```
pedrorezende, line 582: Should we use the method queryAccountById here?

pedrorezende (approved):
LGTM! Only one comment

apps/extension/src/background/keyring/keyring.ts
jurevans: `queryAccountById`, despite its name, returns an array for some reason :/  I feel like it needs to be renamed

PR #464, opened 2023-11-20
chore: fix main CI workflow broken by #456 and rename

emccorson:
Fixes main CI workflow that I broke in #456.

CI run on my fork to show that the fix works (fails at Slack notification as expected since it's a fork):
https://github.com/emccorson/namada-interface/actions/runs/6924320288

---

### Changed
- Rename main CI workflow

### Fixed
- Fix main CI workflow broken by yarn file not being present


jurevans (approved):
Looks good!

PR #472, opened 2023-11-21
feat/470 - Prevent duplicate keys from being added

### LINKED ISSUE #470: Extension: Prevent importing of duplicate keys
- [ ] When user attempts to import keys from Ledger, check to see if address/public key have been added, otherwise, display message to user the key is already imported
- [ ] When user imports a seed phrase or private key, check to see if keys exist, and display message and prevent importing if already imported

jurevans:
resolves #470 

- [x] Prevent duplicate Ledger addreses
- [x] Prevent duplicate keys for mnemonic import
- [x] Prevent duplicate keys for private key import

__NOTE__ I needed to remove and add some `try`/`catch` to get the errors to propagate up to the Setup interface

#### Ledger - duplicate import
Screen Shot 2023-11-21 at 11 12 59 AM

### Mnemonic or Private Key - duplicate import
Screen Shot 2023-11-21 at 11 14 16 AM

### DIFF
--- a/apps/extension/src/background/keyring/service.ts
+++ b/apps/extension/src/background/keyring/service.ts
@@ -4,7 +4,7 @@ import { PhraseSize } from "@namada/crypto";
 import { public_key_to_bech32, Query, Sdk, TxType } from "@namada/shared";
 import { IndexedDBKVStore, KVStore } from "@namada/storage";
 import { AccountType, Bip44Path, DerivedAccount } from "@namada/types";
-import { Result } from "@namada/utils";
+import { Result, truncateInMiddle } from "@namada/utils";
 
 import {
   createOffscreenWithTxWorker,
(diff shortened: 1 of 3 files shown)


apps/extension/src/background/keyring/service.ts
```diff
@@ -114,6 +114,17 @@ export class KeyRingService {
   ): Promise<AccountStore | false> {
     const publicKeyBytes = fromHex(publicKey);
     const bech32PublicKey = public_key_to_bech32(publicKeyBytes);
+
+    const account = (await this._keyRing.queryAllAccounts()).find(
```
pedrorezende, line 118: Maybe we should separate this piece of code into a new function, once we're having the same behavior in keyring.ts

pedrorezende (approved):

PR #478, opened 2023-11-27
Generate manifest extension version from package.json

emccorson:
### Changed
- Generate the manifest.json version field from package.json
- Make the injected and content scripts read the version from package.json

### Added
- Mark Cargo.lock and yarn.lock files as binary in git


jurevans (approved):
Looks good!

PR #481, opened 2023-11-28
WIP: bug/480 - Fix extension connection issues

### LINKED ISSUE #480: bug: Extension - Fix issues with connections
This issue is to implement fixes for the following:

- [ ] Extension fails to connect to interface and allow approvals when no accounts have been created - this should fix that
- [ ] Don't require a chain ID in order to connect - the extension will currently fail to connect if there is a mismatch between the chain ID and the interface during build - this should not prevent interfaces from connecting

jurevans:
Resolves #480 

- [ ] Chain ID should not be a requirement for extension connection
- [ ] Extension connection should not be prevented if no accounts exist

### DIFF
--- a/apps/extension/src/background/approvals/service.ts
+++ b/apps/extension/src/background/approvals/service.ts
@@ -52,7 +52,7 @@ export class ApprovalsService {
     protected readonly keyRingService: KeyRingService,
     protected readonly ledgerService: LedgerService,
     protected readonly vaultService: VaultService
-  ) {}
+  ) { }
 
   async approveTx(
     txType: SupportedTx,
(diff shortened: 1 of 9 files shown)


apps/extension/src/background/approvals/service.ts
```diff
@@ -321,7 +318,7 @@ export class ApprovalsService {
     if (allowConnection) {
       await addApprovedOrigin(this.approvedOriginsStore, interfaceOrigin);
```
mateuszjasiuk, line 319: Not sure if `await this.keyRingService.connect(interfaceTabId);` cant throw an error anymore, but maybe it's worth to move this `await addApprovedOrigin(this.approvedOriginsStore, interfaceOrigin);` below it. Otherwise  we can add origin even if  connection potentially fails? I did this in the airdrop branch

mateuszjasiuk (approved):
LGTM! Added  one comment to check sth :)

apps/extension/src/background/approvals/service.ts
jurevans: Fixed! Thanks!

PR #483, opened 2023-11-29
fix: only add specified env vars to bundle

emccorson:


emccorson:
Closing in favour of #493.

PR #486, opened 2023-12-01
bug/480 - Make chainId optional to support released extension

jurevans:
Relates to #480 

- [x] Make `chainId` optional to support released extension (`Namada` interface, `Namada` provider and integration)


mateuszjasiuk (approved):
LGTM!

PR #487, opened 2023-12-01
bug/480 - Add optional chainId to accounts

jurevans:
Relates to #480

- [x] Optional chain ID for accounts & default account in extension integration


mateuszjasiuk (approved):
Perfect! :)

PR #488, opened 2023-12-01
feat: make pub key read only input a top one

mateuszjasiuk:
Context:
> When we were taking a look at website with Chris Holt
 we've noticed that maybe it would make sense to change order of readonly inputs in extension. If we move pubkey up it will be consistent with airdrop website also both transparent and shielded addresses will be next to each other. So it will look like this:
image


jurevans (approved):
LGTM!

mateuszjasiuk:
Closing this one as it turns out not such a big deal :)

PR #491, opened 2023-12-04
Feat: faucet redesign

RuslanGlaznyov:
## What does this PR do
Implement a new design for the Namada faucet

## Context
1. Community builder @cryptobtcbuyer (CB Discord member) created a new design for a faucet 
https://www.figma.com/file/MpTmPN6A0WLgj0Op8xvQAx/Namada-Faucet-by-cryptobtcbuyer?type=design&node-id=0%3A1&mode=design&t=lxFevWm0BSom96Ic-1
2. `Chris | Head of Art and UX` liked it and asked from community to implement it
 

More context can be found in this Discord CB post:
https://discord.com/channels/833618405537218590/1174062025664647300/1174062372634230846

## Externals 
Preview page: https://faucet-namada-fix.faucet-namada.pages.dev/

https://github.com/anoma/namada-interface/assets/26277721/250b42b8-55c8-49d5-9581-9ca2bd44c52a


apps/faucet/src/App/Faucet.tsx
```diff
@@ -159,6 +161,7 @@ export const FaucetForm: React.FC<Props> = ({ isTestnetLive }) => {
 
       <InputContainer>
         <Input
+          placeholder="From 1 to 1000"
```
emccorson, line 164: The faucet limit is set at build time and is not necessarily 1000, so this should use the variable `faucetLimit` instead of 1000.

apps/faucet/src/App/Faucet.tsx
```diff
@@ -104,7 +106,7 @@ export const FaucetForm: React.FC<Props> = ({ isTestnetLive }) => {
         transfer: {
           target: sanitizedTarget,
           token: tokenAddress,
-          amount: amount * 1_000_000,
+          amount: (amount || 0) * 1_000_000,
```
emccorson, line 109: You can probably just return early from this function if `amount` is undefined since we don't want to submit in that case, the same way `targetAddress` is handled on line 64.

emccorson (commented):
I left just a couple of comments, but I think this is really nice :+1:

apps/faucet/src/App/Faucet.tsx
RuslanGlaznyov: Fixed!

apps/faucet/src/App/FaqDropdown.tsx
```diff
@@ -0,0 +1,29 @@
...
+    <FaqDropdownContainer onClick={handleToggle}>
+      <DropDownTitle>
+        <DropDownTitleText>
+          {title}
+        </DropDownTitleText>
+        <PlusIcon src={plusIcon} isOpen={isOpen} />
+      </DropDownTitle>
+
+      <FaqDropdownContent isOpen={isOpen}>{children}</FaqDropdownContent>
+
```
emccorson, line 26: There is a trailing whitespace error in this line. (Actually, this should have been caught by the linter so I'm a bit surprised it wasn't, but I will fix the linting config in a separate PR.)

RuslanGlaznyov: Strange. Run lint no more issues were detected 
Also, it would be great to add eslint to dev dependency. Because if you haven't eslint  globally you have to run something like that 
`npx eslint src --ext .ts,.tsx`

RuslanGlaznyov: Created PR example with a solution: https://github.com/anoma/namada-interface/pull/505

pedrorezende (approved):
Sorry for the late reply :( I won't be too picky, maybe we could reuse some icons from @components/Icon , but it LGTM 🚀

emccorson:
I squashed this and reworded the commit message, then pushed here (since otherwise GitHub won't realise it's been merged), and merged to main.

Your changes should be live at some point this week (hopefully today).

Thanks for your contribution!

PR #493, opened 2023-12-05
Fix: only add specified env vars to bundle + yarn fixes

emccorson:
In order to whitelist the environment variables we want to add to the bundle, I've prefixed all (hopefully) of our existing environment variables with `NAMADA_INTERFACE_`.  But I'm happy to change the prefix to something different, and it should be an easy change since the prefixing is in its own commit.

---

### Changed
- **Breaking:** Prefix all environment variables with `NAMADA_INTERFACE_`
- Rename @namada/eslint-config to @namada/config
- Make eslint configs JS files to get around eslint config naming requirement

### Fixed
- Only add specified environment variables to the bundle
- Fix faucet yarn scripts not working after yarn v4 upgrade
- Fix yarn not finding eslint by adding it per package
- Fix some yarn missing peer dependency warnings


mateuszjasiuk (approved):
This is really nice and important :) Couldn't find any problems testing airdrops website!

jurevans (approved):
Nice!

PR #502, opened 2023-12-06
Airdrop Fixes

mateuszjasiuk:
- tos capitalization
- retrigger animation in non-eligible view
- change confirmation path
- display amount in correct format - two decimal places rounded down


jurevans (approved):
Looks good!

PR #506, opened 2023-12-06
Airdrop: Terms and Conditions

pedrorezende:
This PR updates the terms and conditions and also adds a new option for the AcceptTermsCheckbox component, which ca be rendered with the acceptance text or with a summary below

### DIFF
--- a/apps/airdrop/src/App/Claim/ClaimConfirmation/ClaimConfirmation.tsx
+++ b/apps/airdrop/src/App/Claim/ClaimConfirmation/ClaimConfirmation.tsx
@@ -9,7 +9,7 @@ import {
 import { DerivedAccount, WindowWithNamada } from "@namada/types";
 import { AirdropResponse, ToastMessage, toast } from "App/utils";
 import { useAtom } from "jotai";
-import { useCallback, useLayoutEffect, useState } from "react";
+import { useCallback, useEffect, useState } from "react";
 import { useNavigate } from "react-router-dom";
 
 import { ClaimsSectionSignature } from "../../App.components";
@@ -23,9 +23,18 @@ import {
   confirmationAtom,
 } from "../../state";
 
+import { bech32mValidation } from "@namada/utils";
 import { AcceptTermsCheckbox } from "App/Common/AcceptTermsCheckbox";
 import { BigNuclearClaimButton } from "App/Common/BigNuclearClaimButton";
 import { StepIndicator } from "App/Common/StepIndicator";
+import {
+  claimWithGitcoin,
+  claimWithGithub,
+  claimWithKeplr,
+  claimWithTrustedSetup,
+} from "App/claimService";
+import { ClaimResponse } from "App/types";
+import gsap from "gsap";
 import {
   ButtonContainer,
   ClaimBadge,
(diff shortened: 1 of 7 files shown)


apps/airdrop/src/App/Claim/ClaimConfirmation/ClaimConfirmation.tsx
```diff
@@ -114,6 +115,14 @@ export const ClaimConfirmation: React.FC = () => {
   const [tsSignature, setTsSignature] = useState<string>("");
   const [isToggleChecked, setIsToggleChecked] = useState(false);
 
+  const isTrustedSetup = claimState?.type === "ts";
+  const isClaimSectionActive = !isTrustedSetup || !!tsSignature;
+  const airdropAddressValid = bech32mValidation("tnam", airdropAddress);
+
+  const isValidForm = Boolean(
+    airdropAddressValid && airdropPubKey && isToggleChecked
```
emccorson, line 123: If the TS signature is removed after the form is completed, `isValidForm` will still be true and the big nuclear claim button will still be shown.

emccorson: Also the same behaviour in the case where after the form is completed, `airdropPubKey` is changed from something valid to something invalid, because this checks whether `airdropPubKey` exists instead of whether it is valid.

emccorson (commented):
Just left a few minor comments, but other than that looks good to me!

mateuszjasiuk (approved):
Can't comment on Terms and Condition text, but overall it looks good!

apps/airdrop/src/App/Common/AcceptTermsCheckbox/AcceptTermsCheckbox.tsx
```diff
@@ -1,27 +1,70 @@
...
 };
 
 export const AcceptTermsCheckbox = ({
   disabled = false,
+  fullVersion = false,
   checked,
   onChange,
 }: AcceptTermsCheckboxProps): JSX.Element => {
   return (
     <TOSToggle disabled={disabled}>
```
mateuszjasiuk, line 18: @pedrorezende 
One thing've noticed is that before when toggle was disabled it had different color:
image
now it's always bright yellow:
image
It might be by design though,  so the text is easier to read.

apps/airdrop/src/App/Claim/ClaimConfirmation/ClaimConfirmation.tsx
```diff
@@ -118,10 +118,13 @@ export const ClaimConfirmation: React.FC = () => {
   const isTrustedSetup = claimState?.type === "ts";
   const isClaimSectionActive = !isTrustedSetup || !!tsSignature;
   const airdropAddressValid = bech32mValidation("tnam", airdropAddress);
+  const airdropPubKeyValid = bech32mValidation("tpknam", airdropPubKey);
 
-  const isValidForm = Boolean(
-    airdropAddressValid && airdropPubKey && isToggleChecked
+  const isTopFormValid = Boolean(isTrustedSetup ? tsSignature : true);
```
emccorson, line 123: `isTopFormValid` looks like it's the same thing as `isClaimSectionActive` so you might be able to remove one of them.

emccorson (approved):
Looks good to me!
One note that I was able to break the big nuclear button animation my toggling a lot between valid and invalid public key input, but I would just ignore it as it's a very small edge case.

apps/airdrop/src/App/Claim/ClaimConfirmation/ClaimConfirmation.tsx
mateuszjasiuk: Thanks! Fixed :)

PR #512, opened 2023-12-08
fix: conditional hook

mateuszjasiuk:
Does not change much other than lint warning as:
```export const useMetamaskHandler = (
  chainId: string,
  metamask?: MetaMaskInpageProvider
): (() => Promise<void>) => {
  const [_cl, setClaim] = useAtom(claimAtom);
  const [_co, setConfirmation] = useAtom(confirmationAtom);

  const handleMetamask = useCallback(async () => {
    if (!metamask) {
      handleExtensionDownload("https://metamask.io/download/");
```
Does:
```
    if (!metamask) {
```


jurevans (approved):
LGTM!

PR #514, opened 2023-12-08
fix: blank screen on airdrop page when no metamask installed 

RuslanGlaznyov:
# What does this PR do? 
Fix the issue when you see a blank screen if Metamsk is not installed! 

# UI  before fix
Blank screen: 
Screenshot 2023-12-08 at 20 07 19
# UI after fix: 
Screenshot 2023-12-08 at 20 09 28


RuslanGlaznyov:
3 minute late

PR #517, opened 2023-12-12
fix: yellow borders on filled action button

RuslanGlaznyov:
# What does this PR do? 
Fix the yellow border when the action button is filled

# Context 
This issue may not exist on all screens, but I see yellow borders when the action button is filled on my retina screen. 
Also, this fix made animation more smooth (IMHO)

# Externals 
## Before fix: 
before

## After fix: 
after

### DIFF
--- a/packages/components/src/ActionButton/ActionButton.components.tsx
+++ b/packages/components/src/ActionButton/ActionButton.components.tsx
@@ -32,7 +32,7 @@ export const ButtonText = styled.span`
   z-index: 10;
 `;
 
-export const ButtonContainer = styled(Button)<{
+export const ButtonContainer = styled(Button) <{
   variant: ThemeColor;
   hoverColor?: ThemeColor;
   size: ButtonSize;
@@ -41,6 +41,13 @@ export const ButtonContainer = styled(Button)<{
 }>`
   all: unset;
   background-image: ${(props) => {
+    if (props.disabled && !props.outlined) {
+      return `linear-gradient(${color(
+        "utility1",
+        "main50"
+      )(props)} 50%, ${color("utility1", "main50")(props)} 50%)`;
+    }
+
     if (props.outlined) {
       // main color is transparent, filler color is yellow
       return `linear-gradient(transparent 50%, ${color(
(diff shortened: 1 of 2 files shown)


pedrorezende (changes requested):
Hey @RuslanGlaznyov ! That was a very good catch! I also noticed that the same rendering error is happening on ProgressIndicator component. I'll check that today. It's so weird that this kind of issue has been around for SO LONG ( https://bugs.chromium.org/p/chromium/issues/detail?id=491574 - it was opened in 2015!). 

I've tested and reviewed your PR (thank you!) and have some comments:

1 - I think it would be nicer if we keep using the former easing / duration, which are `var(--ease-out-circ)` and `0.45s`, in order to make the animation a bit smoother

2 - I've noticed that now the disabled buttons are not rendering with the gray background anymore (`&:disabled { background-color .... }` styles)

3 - Quick question, why have you used 220% instead of 200% for the background-size?

Also, just fyi, I'm working at the moment in the #518 PR, converting all the components to Tailwind. I can migrate your fix there, or if you feel like waiting a bit to send another PR with the new structure, I think the styles are more organized now

RuslanGlaznyov:
@pedrorezende Thanks for your review! 
Provided fixes for `1` and `2`.
> 3 - Quick question, why have you used 220% instead of 200% for the background-size?

If I use 200% instead 220% in zoom you can see a yellow border. In theory, we can use 200%, because the user won't zoom extension.  
b5379e9a-bd9c-4689-be98-f59de9804522


> Also, just fyi, I'm working at the moment in the https://github.com/anoma/namada-interface/pull/518 PR, converting all the components to Tailwind. I can migrate your fix there, or if you feel like waiting a bit to send another PR with the new structure, I think the styles are more organized now 

It's up to you. I'm comfortable with both

emccorson:
@RuslanGlaznyov Sorry for not following this up sooner. We have now completely moved the extension styles over to Tailwind. Would you still be interested in fixing up this PR? I noticed the bug this week so I am pretty sure it is still present.

Also, I think we cannot assume the user won't zoom the extension, so the fix would need to work at any zoom level.

RuslanGlaznyov:
@emccorson thanks for your response!  
I've created a new PR. Because a lot of changes were made since my last commit. I closed this PR and would like to 
move the conversation to the new one 
https://github.com/anoma/namada-interface/pull/660

PR #522, opened 2023-12-14
Feat/0.28.0 signing without wallet

### LINKED ISSUE #523: Make transactions work without using wallet in 0.28.0
So we do not have to insert keypairs to wallet struct

mateuszjasiuk:
Resolves #523


jurevans (approved):
Tested, works nicely, and code looks good!

PR #525, opened 2023-12-14
fix: Correct spelling of GitHub

Yaminyam:
It is correct to capitalize the H in GitHub's brand name.


mateuszjasiuk:
Hello! Thank you for your contribution. As of now we do not accept typo fixing PRs. Especially when they fix only one typo. We do not want repo to be flooded with one line PRs as this as impact on PRs readability.

PR #527, opened 2023-12-15
chore: make e2e tests work with 0.28.0

### LINKED ISSUE #528: Fix e2e tests not working with namada 0.28.0

mateuszjasiuk:
Resolves #528 
- fixes e2e setup
- makes basic tests work again
  - open popup
  - create new account

The rest will be added in new issue


e2e/setup-frontend.sh
```diff
@@ -9,7 +9,7 @@ cd ../apps/extension && \
     export NAMADA_INTERFACE_MASP_PARAMS_PATH=http://localhost:8080/assets/; \
     export NAMADA_INTERFACE_NAMADA_CHAIN_ID=${CHAIN_ID}; \
     export NAMADA_INTERFACE_NAMADA_URL=http://localhost:27657 && \
-    npx webpack --watch &
+    webpack-cli --watch &
```
emccorson, line 12: This might run into problems if you're running the version of webpack from the e2e package.json, because the interface uses webpack v5 and the extension uses webpack v4. I think if you run something like:

    NAMADA_INTERFACE_NAMADA_URL=http://localhost:27657 yarn workspace @namada/extension run start:chrome

instead of directly calling webpack it will use the interface/extension webpack, and you wouldn't need to change directory.

e2e/package.json
```diff
@@ -32,6 +32,7 @@
     "jest-puppeteer": "^9.0.0",
     "puppeteer": "^20.8.2",
     "start-server-and-test": "^2.0.0",
-    "terminate": "^2.6.1"
+    "terminate": "^2.6.1",
+    "webpack-cli": "^5.1.4"
```
emccorson, line 36: If going with the suggestion on setup-frontend.sh, this can be removed too.

e2e/setup-namada.sh
```diff
@@ -1,6 +1,6 @@
 #!/bin/bash -x
 
-VERSION="v0.24.0"
+VERSION="v0.28.0"
```
emccorson, line 3: Also worked locally as v0.28.1 :tada:

emccorson (approved):
Left some minor comments, but code looks good to me and locally the tests are working again! :raised_hands:

apps/extension/src/Setup/AccountCreation/Steps/SeedPhraseConfirmation/SeedPhraseConfirmation.tsx
```diff
@@ -105,6 +105,7 @@ const SeedPhraseConfirmation = (
         <Stack gap={GapPatterns.FormFields}>
           <VerifyPanelContainer>
             <Input
+              data-testid="setup-seed-phrase-verification-1-input"
```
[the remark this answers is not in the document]
mateuszjasiuk: Yeah I agree. I'm not exactly sure yet what are best practices for reusability here. Maybe it's worth having ids unique? I will think about it more when writing next tests.

e2e/setup-frontend.sh
mateuszjasiuk: Cool! I will check it out :)

PR #529, opened 2023-12-15
feat/520 - Hook up additional testnet tokens

### LINKED ISSUE #510: Interface/SDK: Update testnet token address configuration
Currently, we have testnet token addresses defined in two places:

```bash
packages/types/src/tx/tokens/types/index.ts
packages/shared/lib/src/query.rs
```

These should be defined in a single location. This may involve an update to `Query->query_transparent_balance()` where we pass in a vector of address strings to query. Also, we only currently define 3 of the generated addresses, so we should include each potential testnet token address in a easily configurable location, as these will change with each testnet.

Note that to display in the interface, we should perhaps set a generic asset to use for tokens like `Schnitzel`, `Apfel`, etc., (anything that can be transferred via the faucet app should appear in the interface). The interface also expects additional data, currently defined in `packages/types/src/tx/tokens/types/index.ts`, so our config may included additional slip-044 params for these fictional tokens.

- [ ] Define token configuration in single location
- [ ] Update Query `query_transparent_balance` and `query_balance` to accept an array of token addresses to query
- [ ] Provide configuration for _all_ testnet tokens - this needs to remain in sync with current testnet

jurevans:
Resolves #510 

- [x] Display supported tokens that have assets: `NAM`, `BTC`, `ETH`, and `DOT`, add placeholders for `Schnitzel`, `Apfel` and `Kartoffel` (these reuse image assets for Namada)
- [x] Remove `ATOM`, as we do not have an address for this
- [x] Update `query.rs` to accept array of token addresses

_NOTE_ I also updated `AccountOverview` to only display tokens that have a balance, as there are many now.

### Screenshots
Balances (every testnet token enabled)
Screen Shot 2023-12-15 at 1 10 44 PM

### DIFF
--- a/apps/extension/src/background/keyring/keyring.ts
+++ b/apps/extension/src/background/keyring/keyring.ts
@@ -30,6 +30,7 @@ import {
   SubmitUnbondMsgValue,
   SubmitVoteProposalMsgValue,
   SubmitWithdrawMsgValue,
+  Tokens,
   TransferMsgValue,
 } from "@namada/types";
 
@@ -82,7 +83,7 @@ export class KeyRing {
     protected readonly sdk: Sdk,
     protected readonly query: Query,
     protected readonly cryptoMemory: WebAssembly.Memory
-  ) {}
+  ) { }
 
   public get status(): KeyRingStatus {
     return this._status;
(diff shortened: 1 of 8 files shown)


mateuszjasiuk (approved):
LGTM! :)

PR #530, opened 2023-12-18
feat: reduce bundle size

mateuszjasiuk:
- Removed Metamask SDK(it's quit haevy)
- Moved `useIntegration` hook to integrations package(otherwise importing hooks would also import all of the cosmos stuff)
- ~~Changed wasm optimizations for `-z` or `-Oz` which means optimization  for size aggressively(not exactly sure how much this will affect performance :( )~~
- now we copy wasm files to extension assets and use fetch to load them in background service worker 👍 
- I've changed rustopt options to defaults for profile

You can now also run: `BUNDLE_ANALYZE=true yarn build:firefox` (or chrome) to see bundle size analyzed under: `localhost:8888`


jurevans (approved):
Aside from removing:
```
entry: {
    shared: "../../packages/shared/src/init.ts",
```
everything looks good!

mateuszjasiuk:
> Aside from removing:
> 
> ```
> entry: {
>     shared: "../../packages/shared/src/init.ts",
> ```
> 
> everything looks good!

Again 😅 👍

jurevans (approved):
Looks good!

Issue #516, opened 2023-12-12
Faucet - Implement API changes


jurevans:
We will need to implement any changes to the Faucet API into the `faucet` app.

More details _TBD_

PR #532, opened 2023-12-18
feat/516: Hardcode timestamp for faucet app

jurevans:
Relates to #516 

- [x] Hardcode timestamp, only use `/setting` endpoint for `difficulty` and `tokens`

Once merged, we can publish using:
```
NAMADA_INTERFACE_FAUCET_API_URL=https://faucet.api.heliax.click/public-testnet-15.0dacadb8d663
```


emccorson (approved):
Looks good to me!
Since the timestamp is now in the past, does that mean the faucet is live now?

jurevans:
> Looks good to me! Since the timestamp is now in the past, does that mean the faucet is live now?

Yes :D I just deployed from this branch as the launch was approaching. So I think this can still go into `main` should be need to deploy any fixes based on this

PR #533, opened 2023-12-18
feat/531: Namada Interface - Public Testnet updates

### LINKED ISSUE #531: Interface: Public Testnet updates
This issue it to implement any updates to better support the public testnet.

- [x] Hook up to public testnet: `public-testnet-15.0dacadb8d663` and test, using tokens from `https://faucet.heliax.click`
- [x] Disable the Ethereum Bridge form in the interface
- [x] Disable ETH chain config for Goerli network
- [ ] Add Osmosis chain config and assets?
- [ ] Support native Cosmos & Osmosis tokens (i.e., display balance on IBC form)?
- [ ] Support IBC tokens for Cosmos & Osmosis _TODO_
- [x] Perhaps we should not display accounts with 0 token balance in the Send forms? We could link to testnet faucet (for now). Also, there is a bug that on submit transfer, the drop-down defaults to the first entry
- [x] Staking addresses should have aliases! This is very difficult to use unless you know which address has a balance
- [x] Staking - we should probably not display shielded addresses in this drop-down

jurevans:
Resolves #531 

### Notes

- Removed `selectedChainId` from Redux
- Moved network dropdown to IBC form, this now has local component state
- Added integration connect button for _Source_ - this will allow us to load source accounts from Cosmos (and other chains should we add more configs)
- Disabled ETH Bridge form
- Simplified overview heading to just `Wallet` so I didn't have to mess with styles (this display was broken with very large token amounts)
- Remove Staking sub-menu (specifically, the `Staking ... Public Goods Funding` sub-menu
- Don't show tokens with zero balance (Transparent/Shielded Send form, & IBC)

### Screenshots

IBC Form: `Cosmos` selected:
Screen Shot 2023-12-19 at 11 26 01 AM

IBC Form: Keplr extension attached, no balances warning:
Screen Shot 2023-12-20 at 7 59 28 AM

Staking - remove submenu, with _Public Goods Funding_
Screen Shot 2023-12-20 at 7 27 49 AM


apps/namada-interface/src/services/extensionEvents/handlers/namada.ts
```diff
@@ -45,6 +45,9 @@ export const NamadaTxStartedHandler =
 export const NamadaTxCompletedHandler =
   (dispatch: Dispatch<unknown>) => async (event: CustomEventInit) => {
     const { msgId, txType, success, payload } = event.detail;
+    if (payload) {
```
emccorson, line 48: It might be more future-proof to check `success` instead of `payload` for a failed transaction in case we want to send a payload for a successful transaction at some point.

apps/namada-interface/src/App/AccountOverview/DerivedAccounts/DerivedAccounts.tsx
```diff
@@ -38,8 +37,6 @@ import { CoinsState, fetchConversionRates } from "slices/coins";
 import { SettingsState } from "slices/settings";
 import Config from "config";
 
-const { NAMADA_INTERFACE_NAMADA_FAUCET_ADDRESS: faucetAddress } = process.env;
```
emccorson, line 41: This could also be dropped from `.env.sample` now.

packages/types/src/tx/tokens/Cosmos.ts
```diff
@@ -1,22 +1,71 @@
...
+};
 
 export const minDenomByToken = (token: CosmosTokenType): string => {
-  const tokenDenom = tokenDenomLookup(token)
+  const tokenDenom = tokenDenomLookup(token);
   return tokenDenom ? tokenDenom[0] : "";
-}
+};
+
+const supportedCoinTypes: RegisteredCoinType[] = [
```
emccorson, line 37: I think `CosmosSymbols.includes` could be used instead of `indexOf`, and the array spread can also be removed.

emccorson (approved):
Looks good to me! Just left some minor comments.

mateuszjasiuk (approved):
LGTM!

apps/namada-interface/src/services/extensionEvents/handlers/namada.ts
jurevans: fixed!

packages/types/src/tx/tokens/Cosmos.ts
jurevans: Fixed!

apps/namada-interface/src/App/AccountOverview/DerivedAccounts/DerivedAccounts.tsx
jurevans: Done!

PR #538, opened 2023-12-19
feat: masp with disposable key

### LINKED ISSUE #524: Bug: Transfers from shielded account are broken
When doing transfers from shielded account, we were originally getting an error that gas-payer cannot be MASP, but recently this is throwing the error: `cannot read properties of undefined` when doing a transfer from a shielded account.

mateuszjasiuk:
Resolves #524
We have to use disposable key for the privacy reasons.

### DIFF
--- a/apps/extension/src/background/keyring/keyring.ts
+++ b/apps/extension/src/background/keyring/keyring.ts
@@ -83,7 +83,7 @@ export class KeyRing {
     protected readonly sdk: Sdk,
     protected readonly query: Query,
     protected readonly cryptoMemory: WebAssembly.Memory
-  ) { }
+  ) {}
 
   public get status(): KeyRingStatus {
     return this._status;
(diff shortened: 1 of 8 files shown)


apps/extension/src/background/keyring/keyring.ts
```diff
@@ -761,9 +761,13 @@ export class KeyRing {
 
     if (account.public.type === AccountType.ShieldedKeys) {
       const xsk = JSON.parse(sensitiveProps.text).spendingKey;
+      //TODO: verify if this is still needed
       // Append xsk to SDK wallet instance
       this.sdk.add_spending_key(xsk, account.public.id);
```
jurevans, line 766: We can get rid of this line. I just confirmed that it works without this with your other updates!

jurevans (approved):
This looks good, and works great for me! Just left a couple minor comments

PR #547, opened 2023-12-26
Remove noise content script log

grod220:
Given this happens on every successful connection, it's a bit noisy. Especially given it logs no matter what website your are viewing. Think it's worth removing or putting behind a "debug mode" flag.

### DIFF
--- a/apps/extension/src/extension/ExtensionRouter.ts
+++ b/apps/extension/src/extension/ExtensionRouter.ts
@@ -23,7 +23,6 @@ export class ExtensionRouter extends Router {
       throw new Error("Empty port");
     }
 
-    console.info(`Listening on port ${port}`);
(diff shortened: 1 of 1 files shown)


mateuszjasiuk:
Hello, thank you for contribution. We do not consider this a noise at this point, so I will close this one.

grod220:
I can definitely see the use for debugging, but in the case of extensions, this log console is shared between all websites. Prod best practices are that these should be kept clean as possible so folks can identify only relevant issues.

Have you considered implementing different log levels? For example, having a `debug` log level fits this best. It would only display when the user enables a flag or is in local dev mode.

mateuszjasiuk:
Yes! We just have a lot of other things to do at the moment :)

PR #554, opened 2024-01-08
feat/552 - Updates to integrations, better support chain ID updates

### LINKED ISSUE #552: Interface - Public Testnet release updates
With the extension update in https://github.com/anoma/namada-interface/pull/551, there are a few small tasks for the interface before we re-publish it:

- Bump interface & packages to `0.2.0` (to match extension)
- Update components to use newest components where the display is quite affected (especially where it prevents text from being readable)
- Make sure Settings button shows up so users can select currency (not appearing for me in desktop mode, but appears in mobile menu)
- There may be a potential issue with integrations as they are indexed by chain ID - we may want to move this instead to extension-related keys, like `namada | keplr | metamask` and avoid indexing things with chain ID (not sure if we want to do this for accounts to, as that would affect IBC, where we are loading Keplr accounts by chain ID, but that may be a future task) - I believe this is why the integration fails to work correctly when the extension & interface are built with different chain IDs - this should fix that
- Make _Update balance_ fetch more efficient - this is getting called inadvertently, which can cause issues with slow connections (like the testnet), where if you switch from testnet to a local node, the new balances display, then the old balances come back when the extra testnet balance call finishes. Also, it appears that these requests are being issues more often than needed (NOTE We are querying for a lot of token types now, so if these are inadvertantly re-issued, that's a lot of extra requests)

jurevans:
resolves #552 

- Bump versions to `0.2.1`
- Updates integrations (`useIntegration`) to index instances by a `ChainKey` which is `"namada" | "cosmos" | "ethereum"` for now, instead of chain Id, as the chain ID can be changed on the Namada side in the extension
- Similarly, account state in Interface was also keyed this way, updated to use `ChainKey`
- With this change, got rid of the `defaultChainId` export (and related ones from Cosmos, etc.)
- Changing `chain ID` in Extension does not break association with integration now
- Fixed IBC form to work with integration updates
- Update Ledger Tx to use correct `chainId` for Reveal PK (`ConfirmLedgerTx`)
- All requests in Interface now use `rpc` & `chainId` from the extension settings (which user may override)

### Testing

If you have balances on the current public testnet, then testing is pretty easy. This is the default value in the chain config for Namada, so the extension & interface can be built without any `.env` variables - if they are defined, they only affect the initial state of the chain config in the extension, and a user can override those with the `Network` form

1. With a local chain connected, use the `Network` form in the interface, and update the Chain ID & URL
2. Issue a transaction on the interface - this should submit to your local chain, with the correct chain ID (it will fail and complain otherwise)
3. In the extension, switch the network to the following:
  ```
  public-testnet-15.0dacadb8d663
  https://proxy.heliax.click/public-testnet-15.0dacadb8d663/
  ```
  - Balances should update in the interface (if you have testnet balances for those same accounts)
4. Issue a transaction - this should use balances & chain ID for the testnet

Also, we can test integrations on the IBC page - when switching Network around, and without refreshing the interface, you should see that Keplr & Namada integrations always work


mateuszjasiuk (approved):
Code looks good! Couldn't find problems during testing!

apps/namada-interface/src/App/Token/index.ts
```diff
@@ -1,5 +1,5 @@
-export * from "./TokenSend";
-export * from "./TokenReceive";
 export * from "./IBCTransfer";
-export * from "./EthereumBridge";
+export * from "./TokenReceive";
+export * from "./TokenSend";
+// export * from "./EthereumBridge";
```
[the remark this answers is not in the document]
jurevans: Thanks! Removed

apps/namada-interface/src/App/Token/EthereumBridge/index.ts
```diff
@@ -1,2 +1,2 @@
-export * from "./EthereumBridge";
-export * from "./EthereumBridge.components";
+// export * from "./EthereumBridge";
```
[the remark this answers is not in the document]
jurevans: Thanks! I added it back in here, as we will eventually bring this back, I think

apps/namada-interface/src/slices/proposals.ts
```diff
@@ -35,7 +35,7 @@ export const fetchProposals = createAsyncThunk<
   void,
   { state: RootState }
 >(`${PROPOSALS_ACTIONS_BASE}/${ProposalsActions.FetchProposals}`, async () => {
-  const { rpc } = chains[defaultChainId];
+  const { rpc } = chains.namada;
```
emccorson, line 38: In some cases, we are getting the Namada chain config from Redux state, and in others using `chains.namada`. What is the difference?

emccorson (approved):
Code looks good to me! I just left some questions where I'm not familiar with the code.

mateuszjasiuk (approved):

apps/namada-interface/src/slices/proposals.ts
jurevans: oh, this is wrong! That should be pulled from Redux state. I can add it here, or in a follow up PR. I mainly focused on getting everything confirmed to work at least for transfers (to confirm extension is correct), but yeah, this needs to use Redux state here.

jurevans: I just pushed up a fix for this, and also the Staking actions were referring to default config instead of the Redux state

PR #557, opened 2024-01-09
Fix extension manifest permissions, fix approve Tx form

jurevans:
This PR simply updates the following:

- Chrome `manifest.json` permissions removed for `identity` and `notifications`, as they are not used
- Approve Tx form was not properly authenticating users, this is fixed:
  - Removed `password` as a parameter to `SubmitApprovedTxMsg`, and instead am using `UnlockVaultMsg` to validate user credentials - User will now see `Invalid password!` if authentication fails, and will be allowed to try again


mateuszjasiuk (approved):
LGTM!

PR #571, opened 2024-01-15
feat: yarn publish

mateuszjasiuk:
- added release-it package

For now, to simplify things, I've added two separate dependencies and publish scripts to two separate packages.

You can test it with:
`yarn release:dry-run` inside `shared` and `types` packages

### DIFF
--- a/packages/shared/src/borsh-schemas.ts
+++ b/packages/shared/src/borsh-schemas.ts
@@ -1,6 +1,15 @@
-import { field, vec } from "@dao-xyz/borsh";
+import { BinaryReader, BinaryWriter, field, vec } from "@dao-xyz/borsh";
 import BigNumber from "bignumber.js";
-import { BigNumberSerializer } from "@namada/types";
+
+export const BigNumberSerializer = {
+  serialize: (value: BigNumber, writer: BinaryWriter) => {
+    writer.string(value.toString());
+  },
+  deserialize: (reader: BinaryReader): BigNumber => {
+    const valueString = reader.string();
+    return new BigNumber(valueString);
+  },
+};
 
 export class Proposal {
   @field({ type: "string" })
--- a/package.json
+++ b/package.json
@@ -21,8 +21,6 @@
     "@cosmjs/encoding": "^0.27.1",
     "@cosmjs/json-rpc": "^0.27.1",
     "@cosmjs/tendermint-rpc": "^0.27.1",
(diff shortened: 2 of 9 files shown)


emccorson (commented):
Code looks good to me! I have a few comments about how we will use this:

- I don't think we can use the default tagging of `release-it` because it will give name the tag something like `0.1.0`, but we would need to prefix it to something like `shared-0.1.0`.
- The same with the changelog, since it will include commits that don't affect the shared/types packages (and maybe changelogs should be written manually anyway if we are going to use them for these packages).
- The shared package has some missing peer dependency warnings that will need to be fixed before we publish to npm.
- I'm not sure how to handle multicore, since we effectively have two builds of the shared package now. I guess we could bundle both builds in the same release, or make separate releases for each build.

mateuszjasiuk:
>     * I don't think we can use the default tagging of `release-it` because it will give name the tag something like `0.1.0`, but we would need to prefix it to something like `shared-0.1.0`.
Right! Taking a look
 
>     * The same with the changelog, since it will include commits that don't affect the shared/types packages (and maybe changelogs should be written manually anyway if we are going to use them for these packages).
I will look into disabling this for time being

>     * The shared package has some missing peer dependency warnings that will need to be fixed before we publish to npm.
👍 

>     * I'm not sure how to handle multicore, since we effectively have two builds of the shared package now. I guess we could bundle both builds in the same release, or make separate releases for each build.
We will just publish non multicore version for now :) as it has the same  "API"

packages/shared/src/borsh-schemas.ts
```diff
@@ -1,6 +1,15 @@
-import { field, vec } from "@dao-xyz/borsh";
+import { BinaryReader, BinaryWriter, field, vec } from "@dao-xyz/borsh";
 import BigNumber from "bignumber.js";
-import { BigNumberSerializer } from "@namada/types";
+
+export const BigNumberSerializer = {
```
emccorson, line 4: What does it do to copy BigNumberSerializer to shared?

emccorson (approved):
Looks good to me! Just left one question :+1:

packages/shared/src/borsh-schemas.ts
mateuszjasiuk: I wanted to remove dependency from types  :) In fturue it  wouldbe  nice to plan a  bit what is dependant on what.

PR #573, opened 2024-01-17
fix: publish

mateuszjasiuk:
Otherwise packages won't build.
To test:
- `yarn prepublish`  in the types or shared package
- `npm pack`
- add some repo with `npm init`
- inside repo `npm install --save-dev webpack`
- inside repo `npm install {PATH_TO_PACKAGE_TARBALL}`
- add index.js with something like this:
```js
import { Query } from "@namada/shared";
import { Events } from "@namada/types";

console.log(Query, Events);
```
```sh
npx webpack ./index.js
node ./dist/main.js
```


jurevans (approved):
Reviewed and tested, worked perfectly for me!

emccorson (approved):
I tried the command ` npx tsx src/index.ts` but for some reason it didn't do anything for me.

It should be possible to use something like the `prepare` life cycle script to do the building for us before publish/packing. That should give us `.js` files, a `.d.ts` file for the types, and the WASM files, unless we're going to publish the TS source files. I think in that case we'd also need to change the package.json so `main` points to the JS file, and `types` points to the `.d.ts` file. That would also mean we don't need to add `tsconfig.json` to `files`, and probably can still use `@namada/config` as a devDependency since devDepdendencies shouldn't be installed by default.

But that might be too much of a hassle to set up, so if this appears to be working for getting the packages published, then this is good!

mateuszjasiuk:
> I tried the command ` npx tsx src/index.ts` but for some reason it didn't do anything for me.
> 
> It should be possible to use something like the `prepare` life cycle script to do the building for us before publish/packing. That should give us `.js` files, a `.d.ts` file for the types, and the WASM files, unless we're going to publish the TS source files. I think in that case we'd also need to change the package.json so `main` points to the JS file, and `types` points to the `.d.ts` file. That would also mean we don't need to add `tsconfig.json` to `files`, and probably can still use `@namada/config` as a devDependency since devDepdendencies shouldn't be installed by default.
> 
> But that might be too much of a hassle to set up, so if this appears to be working for getting the packages published, then this is good!

Yeah we should do that :D

jurevans (approved):
Code looks good! Works great for me!

PR #574, opened 2024-01-18
Extension: Sign Arbitrary Data

jurevans:
Adding functionality to extension to sign arbitrary data with the extension.

- Added a `.sign(address, data)` function to to `namada.getSigner()` to produce signatures for arbitrary data with our keys. This promise resolves when the user has both approved the signature and successfully authenticated, and is rejected if the user clicks `Reject`
- As the `data` can be of arbitrary size, similar to the Tx approvals, this gets added to a temporary in-memory key store, referenced by `msgId` passed to the approval window. There is a limit to the number of characters we can provide to the Approval popup via query-strings

## Testing

- Install the extension, setup accounts, and copy one of the addresses (make sure it is unlocked before next step)
- In the browser console, do the following (replacing the following address with a real address from your wallet):
  ```
  const signer = namada.getSigner();
  const signature = await signer.sign("tnam1qry3lnk03j965y92np6e25jvadk3kw9u7cvwjclp", "Arbitrary Data!");
  ```
  The resulting output should look something like:
  ```
  {
    "hash": "c5ebe9588ae5b0966977043d94f033d5d729787fdc4df686cf54d0388968df6b",
    "signature": "5f62366e652209586208f5e8b40014b35e1361e1d82dd67bd0609e894fd60a7cdb8ad7c43b7e778bd1e38b5ef259c9e8f3338e5929d4468333a1d46b4706660e"
  }
  ```
- Also, you can test `Reject` behavior - this will invoke `Promise.reject()` so it can be handled in a client, you should just see an error in this case

_NOTE_ You can also invoke it directly on the injected `namada` like this:
```
await namada.sign({ signer: "tnam1qry3lnk03j965y92np6e25jvadk3kw9u7cvwjclp", data: "Arbitrary Data!" });
```

### Screeenshots
Sign-Arbitrary approval:
Screen Shot 2024-01-23 at 8 20 43 AM

Screen Shot 2024-01-23 at 8 21 08 AM

Screen Shot 2024-01-23 at 8 23 12 AM

Screen Shot 2024-01-23 at 8 24 03 AM

### DIFF
--- a/apps/extension/src/Approvals/ConfirmSignature.tsx
+++ b/apps/extension/src/Approvals/ConfirmSignature.tsx
@@ -47,7 +47,7 @@ export const ConfirmSignature: React.FC<Props> = ({ details }) => {
       await requester
         .sendMessage(
           Ports.Background,
-          new SubmitApprovedSignatureMsg(signer, msgId)
+          new SubmitApprovedSignatureMsg(msgId, signer)
         )
         .catch((e) => {
           throw new Error(e);
--- a/apps/extension/src/Approvals/ConfirmSignature.tsx
+++ b/apps/extension/src/Approvals/ConfirmSignature.tsx
@@ -1,0 +1,100 @@
+import { useCallback, useEffect, useState } from "react";
+import { useNavigate } from "react-router-dom";
+
+import { ActionButton, Alert, Input, Stack } from "@namada/components";
+import { shortenAddress } from "@namada/utils";
+import { SignatureDetails, Status } from "Approvals/Approvals";
+import { SubmitApprovedSignatureMsg } from "background/approvals";
+import { UnlockVaultMsg } from "background/vault";
+import { useRequester } from "hooks/useRequester";
+import { Ports } from "router";
+import { closeCurrentTab } from "utils";
+
+type Props = {
+  details?: SignatureDetails;
+};
+
+export const ConfirmSignature: React.FC<Props> = ({ details }) => {
+  const { msgId, signer } = details || {};
+
+  const navigate = useNavigate();
+  const requester = useRequester();
+  const [password, setPassword] = useState("");
+  const [error, setError] = useState<string>();
+  const [status, setStatus] = useState<Status>();
+  const [statusInfo, setStatusInfo] = useState("");
+
+  const handleApproveSignature = useCallback(async (): Promise<void> => {
+    if (!msgId || !signer) {
+      throw new Error("Not all required arguments were provided!");
+    }
+
+    setStatus(Status.Pending);
+    setStatusInfo(
+      `Decrypting keys and signing with ${shortenAddress(signer, 24)}`
+    );
+
+    try {
+      const isAuthenticated = await requester.sendMessage(
+        Ports.Background,
+        new UnlockVaultMsg(password)
+      );
+
+      if (!isAuthenticated) {
+        throw new Error("Invalid password!");
+      }
+
+      await requester
+        .sendMessage(
+          Ports.Background,
+          new SubmitApprovedSignatureMsg(signer, msgId)
+        )
+        .catch((e) => {
+          throw new Error(e);
+        });
+      setStatus(Status.Completed);
+    } catch (e) {
+      console.info(e);
+      setError(`${e}`);
+      setStatus(Status.Failed);
+    }
+  }, [password]);
+
+  useEffect(() => {
+    if (status === Status.Completed) {
+      closeCurrentTab();
+    }
+  }, [status]);
+
+  return (
+    <Stack gap={4}>
+      {status === Status.Pending && <Alert type="info">{statusInfo}</Alert>}
+      {status === Status.Failed && (
+        <Alert type="error">
+          {error}
+          <br />
+          Try again
+        </Alert>
+      )}
+      {status !== (Status.Pending || Status.Completed) && signer && (
+        <>
+          <Alert type="warning">
+            Decrypt keys for{" "}
+            <strong className="text-xs">{shortenAddress(signer)}</strong>
+          </Alert>
+          <Input
+            variant="Password"
+            label={"Password"}
+            onChange={(e) => setPassword(e.target.value)}
+          />
+          <Stack gap={3} direction="horizontal">
+            <ActionButton onClick={handleApproveSignature} disabled={!password}>
+              Authenticate
+            </ActionButton>
+            <ActionButton onClick={() => navigate(-1)}>Back</ActionButton>
+          </Stack>
+        </>
+      )}
+    </Stack>
+  );
+};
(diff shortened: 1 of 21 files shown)


apps/extension/src/provider/Signer.ts
```diff
@@ -58,6 +59,14 @@ export class Signer implements ISigner {
       };
     }
   }
+
+  public async sign(
+    signer: string,
+    data: string
+  ): Promise<SignatureResponse | undefined> {
+    return await this._namada.sign({ signer, data });
```
emccorson, line 67: This is just a question, but I get a bit confused about the difference between InjectedNamada and Signer. Is it correct that to call `sign`, it goes:

1. Call `getSigner` on InjectedNamada
2. Call `sign` on the returned Signer
3. ...which then calls `sign` on InjectedNamada
4. ...which calls the proxy to contact the extension?

packages/shared/lib/src/sdk/mod.rs
```diff
@@ -380,6 +382,27 @@ impl Sdk {
...
+    ) -> Result<JsValue, JsError> {
+        let hash = Hash::sha256(&data);
+        let secret = ed25519::SecretKey::from_str(&signing_key)?;
+        let signature: ed25519::Signature = ed25519::SigScheme::sign(&secret, &hash);
+
+        let sig_bytes: &[u8] = &signature.0.to_bytes();
+
+        to_js_result((
+            // Hash
+            &hash.to_string().to_lowercase(),
```
emccorson, line 401: Just another question, but why does the hash get returned here?

emccorson (approved):
Works nicely for me and the code looks good! I Just left a couple of questions.

I'm also wondering whether the extension needs to prompt the user for approval here. Otherwise a website could use the user's key to silently sign any random data.

mateuszjasiuk (approved):
LGTM!

packages/shared/lib/src/sdk/mod.rs
jurevans: This is useful for the backend - the signature is actually for this hash, so the backend can validate that it was both hashed and signed correctly. And Gianmarco requested that it be in hex-lowercase

apps/extension/src/background/approvals/service.ts
```diff
@@ -46,12 +47,84 @@ export class ApprovalsService {
...
+    }
+
+    if (!data) {
+      throw new Error(`Signing data for ${msgId} not found!`);
+    }
+
+    try {
+      const signature = await this.keyRingService.signArbitrary(
+        signer,
+        data || "test"
```
mateuszjasiuk, line 107: I think this `||` is not needed, right?

jurevans:
> Works nicely for me and the code looks good! I Just left a couple of questions.
> 
> I'm also wondering whether the extension needs to prompt the user for approval here. Otherwise a website could use the user's key to silently sign any random data.

This is now implemented!

mateuszjasiuk (approved):
LGTM!

apps/extension/src/background/approvals/service.ts
jurevans: you are correct, that was just used when debugging an issue :D Fixed!

apps/extension/src/provider/Signer.ts
jurevans: I think this is something that should be refactored in the near future, such that either all signing functions are only present in the `Signer` class, or they are moved to `Namada` and `Signer` is removed. That would apply to other functionality that is there. Originally we wanted to match the interface of Keplr's injected instance, but that's probably not as useful as we once thought.

apps/extension/src/background/approvals/service.ts
```diff
@@ -46,12 +47,81 @@ export class ApprovalsService {
...
+      throw new Error(`Signing data for ${msgId} not found!`);
+    }
+
+    try {
+      const signature = await this.keyRingService.signArbitrary(signer, data);
+      resolvers.resolve(signature);
+    } catch (e) {
+      resolvers.reject(e);
+    }
+    resolvers.resolve();
```
emccorson, line 110: I think this won't do anything since the promise will already be resolved/rejected before this line.

apps/extension/src/background/approvals/handler.ts
```diff
@@ -96,3 +118,27 @@ const handleRevokeConnectionMsg: (
...
+  return async ({ senderTabId: popupTabId }, { msgId }) => {
+    return await service.rejectSignature(popupTabId, msgId);
+  };
+};
+
+const handleSubmitApprovedSignatureMsg: (
+  service: ApprovalsService
+) => InternalHandler<SubmitApprovedSignatureMsg> = (service) => {
+  return async ({ senderTabId: popupTabId }, { msgId, signer }) => {
+    return await service.submitSignature(popupTabId, signer, msgId);
```
emccorson, line 142: It looks like the `signer` and `msgId` arguments are the wrong way round.

apps/extension/src/Approvals/ConfirmSignature.tsx
```diff
@@ -0,0 +1,100 @@
...
+      );
+
+      if (!isAuthenticated) {
+        throw new Error("Invalid password!");
+      }
+
+      await requester
+        .sendMessage(
+          Ports.Background,
+          new SubmitApprovedSignatureMsg(signer, msgId)
```
emccorson, line 50: These are also the wrong way round, so they cancel each other out and the signing still works :smile:

emccorson (approved):
Works nicely and code looks good to me! I just left a few minor comments.

apps/extension/src/Approvals/ConfirmSignature.tsx
jurevans: Thanks, I thought I had fixed that, but you're right! I'll fix this up now

apps/extension/src/background/keyring/keyring.ts
```diff
@@ -880,4 +882,17 @@ export class KeyRing {
     const query = await this.sdkService.getQuery();
     return await query.query_public_key(address);
   }
+
+  async signArbitrary(
+    signer: string,
+    data: string
+  ): Promise<SignatureResponse | undefined> {
```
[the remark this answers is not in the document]
jurevans: Thanks, I forgot to clean that up!

apps/extension/src/provider/Signer.ts
jurevans: I opened this to track this issue: https://github.com/anoma/namada-interface/issues/583 - this is something that can be addressed before the next release

apps/extension/src/background/approvals/handler.ts
jurevans: Fixed, thanks!

PR #576, opened 2024-01-22
feat: calculate gas fees in token send form

emccorson:
### Changed
- Replace gas fee buttons with minimum gas price

### Added
- Calculate gas fee based on minimum gas price and whether reveal PK transaction is needed
- Add max button to use all available funds in account

### Removed
- Remove unused QR reader code


mateuszjasiuk (approved):
LGTM! One thing that I've noticed is that if you have only one address with balance it is impossible to send transfer? I think it's the same problem for shared source, but I think we can solve it in another PR

jurevans (approved):
LGTM!

PR #579, opened 2024-01-22
fix: make signing with imported secret key work

### LINKED ISSUE #561: BUG: Signing issue with imported private key
We need to add logic to enable signing with a `AccountType.PrivateKey` - when these are imported, the signing assumes that it is `AccountType.Mnemonic` and will fail to sign for those keys.

mateuszjasiuk:
Resolves #561


jurevans (approved):
This looks good to me!

PR #581, opened 2024-01-23
fix: fix balances not displaying on initial connection

### LINKED ISSUE #575: Bug: Balances don't display when initially connecting extension
We need to fix the issue where balances don't display when initially connecting extension. This is also breaking end-to-end tests. I believe this is only an issue with the interface.

- [ ] When a user initially connects the extension to the interface, the balance requests should be fetched, and balances updated accordingly

emccorson:
It was failing because there was no chain config present in the store.

Fixes #575.


jurevans (approved):
LGTM! This fixes another issue we encountered today :)

PR #582, opened 2024-01-23
fix: query account balances individually

### LINKED ISSUE #558: Split up transparent & shielded balance queries
It would be nice if the balance queries for transparent & shielded addresses could be fetched independently to avoid making the user wait for all to complete before getting a result (as shielded balance queries take substantially longer). This will take a little bit of refactoring:

- Balances are invoked via the extension integration - we could keep this function for transparent only, and then (for the Namada integration only) add a `queryShieldedBalances`, _or_, we could just issue these queries directly in the interface. Ultimately, all RPC calls will happen on the interface side
- In `shared/lib/src/query.rs`, we should update the `query_transparent_balances` and `query_shielded_balances` to both use the data formatting from `query_balances`, then possibly remove `query_balances` altogether if we have no further use for it
- When the interface dispatches these requests, it should be updated to do so separately so that the completion of one is not dependent on the other

emccorson:
Fixes #558.

### DIFF
--- a/apps/namada-interface/src/slices/accounts.ts
+++ b/apps/namada-interface/src/slices/accounts.ts
@@ -1,9 +1,9 @@
 import { createAsyncThunk, createSlice, PayloadAction } from "@reduxjs/toolkit";
 import BigNumber from "bignumber.js";
 
-import { Account as AccountDetails, TokenType } from "@namada/types";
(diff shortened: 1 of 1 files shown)


jurevans (approved):
Code looks good, and works very well!

PR #588, opened 2024-01-26
fix: await connect message send before closing tab

emccorson:
This should fix connections to the extension from intermittently failing when the approval window closes before the connection message has sent.


mateuszjasiuk (approved):
This makes sense to me :) I've tested by setting timeout on the sendMessage and it did not  resolve the promise!

jurevans (approved):
LGTM!

PR #603, opened 2024-01-30
Update imports, fix signing

jurevans:
Updating Rust imports in `@namada/shared`

- Moved `BorshSerialize` `BorshDeserialize` and `BorshSerializeExt` to `namada::core::borsh`
- Fix spots where `SecretKey` was importing from `pos::common`. Switched to `namada::types::key::common`
- Also, import `SigScheme` & `Signature` from `namada::types::key::common`
- Hooked up `verify_arbitrary` to Extension API

### Testing

```
// Change `signer` to address from your wallet:
await namada.sign({ signer: "tnam1qry3lnk03j965y92np6e25jvadk3kw9u7cvwjclp", data: "Arbitrary Data!" });

// Change these to match public key in your wallet for that account, and the hash and signature from the previous response
await namada.verify({
  publicKey: "tpknam1qrxzmeyka3v43jnrhep9stnj3jhtgzq96ku3u6lk8hy8xqmjqgjtw8qu274",
  hash: "c5ebe9588ae5b0966977043d94f033d5d729787fdc4df686cf54d0388968df6b",
  signature: "005f62366e652209586208f5e8b40014b35e1361e1d82dd67bd0609e894fd60a7cdb8ad7c43b7e778bd1e38b5ef259c9e8f3338e5929d4468333a1d46b4706660e",
});

// Expected: Promise resolves with `null`

```

Alternatively, for `verify` you can change some of the characters in the `hash` to view the error state, promise should error out with `Invalid signature`

### DIFF
--- a/apps/extension/src/provider/Namada.ts
+++ b/apps/extension/src/provider/Namada.ts
@@ -5,6 +5,7 @@ import {
   SignArbitraryProps,
   SignatureResponse,
   TxMsgProps,
+  VerifyArbitraryProps,
 } from "@namada/types";
 import { MessageRequester, Ports } from "router";
 
@@ -19,6 +20,7 @@ import {
   QueryAccountsMsg,
   QueryBalancesMsg,
   QueryDefaultAccountMsg,
+  VerifyArbitraryMsg,
 } from "./messages";
 
 export class Namada implements INamada {
@@ -61,6 +63,14 @@ export class Namada implements INamada {
     return await this.requester?.sendMessage(
       Ports.Background,
       new ApproveSignArbitraryMsg(signer, data)
+    );
+  }
+
+  public async verify(props: VerifyArbitraryProps): Promise<void> {
+    const { publicKey, hash, signature } = props;
+    return await this.requester?.sendMessage(
+      Ports.Background,
+      new VerifyArbitraryMsg(publicKey, hash, signature)
     );
   }
 
--- a/apps/extension/src/background/keyring/handler.ts
+++ b/apps/extension/src/background/keyring/handler.ts
@@ -5,6 +5,7 @@ import {
   QueryAccountsMsg,
   QueryBalancesMsg,
   QueryDefaultAccountMsg,
+  VerifyArbitraryMsg,
 } from "provider/messages";
 import { Env, Handler, InternalHandler, Message } from "router";
 import {
@@ -110,6 +111,8 @@ export const getHandler: (service: KeyRingService) => Handler = (service) => {
           env,
           msg as AddLedgerAccountMsg
         );
+      case VerifyArbitraryMsg:
+        return handleVerifyAbitraryMsg(service)(env, msg as VerifyArbitraryMsg);
       default:
         throw new Error("Unknown msg type");
     }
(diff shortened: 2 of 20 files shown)


gnosed:
Investigating some error of deserialization. I get the following error in the console:
```
namada.ts:47 1 failed: Error: Could not submit bond tx: Error: Encountered error while broadcasting transaction: server error: {"code":2,"data":"","log":"Mempool validation failed: Error decoding a transaction from bytes: Error deserializing transaction field bytes: Not all bytes read","hash":"7357A4155B30423A5F942B0DB60050E44A9F56BD9DB1EAE6E7018CBAE6D8607C"}

Location:
    /home/runner/.cargo/registry/src/index.crates.io-6f17d22bba15001f/flex-error-0.4.4/src/tracer_impl/eyre.rs:10:9
```

Could it be related to the latest push to `main` and this draft?

apps/extension/src/provider/Namada.ts
```diff
@@ -64,6 +66,15 @@ export class Namada implements INamada {
     );
   }
 
+  public async verify(props: VerifyArbitraryProps): Promise<void> {
+    const { publicKey, hash, signature } = props;
+    return await this.requester?.sendMessage(
+      Ports.Background,
+      // TODO:
```
emccorson, line 73: Is this TODO done?

emccorson (approved):
Looks good to me and works very nicely!

gnosed:
Don't have the error anymore, but it doesn't neither popup the modal of the confirmation nor apply any transaction in the network

mateuszjasiuk (approved):
Tested and LGTM!

apps/extension/src/provider/Namada.ts
jurevans: Haha, yes, it is indeed. Fixed!

jurevans:
> Error: Could not submit bond tx: Error: Encountered error

That looks like the error you will see when broadcasting to the wrong version of the chain. Our `main` branch is using version `0.31.0` of the SDK which introduced a change to the Tx that would cause this error. Those you said you weren't seeing it anymore? `Bond` and other Tx are working off `main` for `v0.31.0` of Namada

jurevans:
> Don't have the error anymore, but it doesn't neither popup the modal of the confirmation nor apply any transaction in the network

This is a different issue outside of the scope of this PR, and it sounds like the interface you're using and the extension are potentially out of sync. If you're still experiencing this, can you open an issue and specify what interface you're using (`namada.me`, or a local interface), what version of the extension (also, whether it is local or installed from the store), and what browser you're testing on?

gnosed:
Ok, just tested again building from main with Google Chrome (Version 121.0.6167.139 (Official Build) (arm64)) and Extension version 0.2.3 from the Chrome Store. The Unbond fails this time with `Error: Could not submit unbond tx: Error: Could not retrieve from storage the gas cost for token tnam1qxgfw7myv4dh0qna4hq0xdg6lx77fzl7dcem8h7e`, but I guess it's because the token address is still not up-to-date to the one from the SE.

Some UX remark, it would be nice to have some loading indicator, while it's fetching massive amount of data through the rpc. Otherwise, it seems to be broken. Also, my balances keep showing me 0.00 $ which is not the case in reality.

Issue #608, opened 2024-02-02
BUG - Extension & Interface bugs related to SE chain


jurevans:
Tracking known bugs in this ticket

- Balances in interface no longer work properly (has the response changed in Query?)
- Native token address should be an env, and supplied to `Sdk` constructor to enable features like Staking Tx to work with SE chain
- Need a `Memo` field hooked up for Tx, enabled in Tx args in shared
- Bip39 "passphrase" should be hooked up to Extension setup page (mnemonic)

More _TBD_

jurevans:
Closing as of `v0.2.4` release of extension - these are fixed!

PR #611, opened 2024-02-02
feat/608 - Support memo, use native token

jurevans:
relates to #608 

- Updated `Query` to support `query_native_token` - this is called when the chain is initialized or updated, and will fall back to `NAMADA_INTERFACE_NAMADA_TOKEN`, otherwise will fall back to a string. Interface updated to use this token address for `nativeToken`
- `Sdk` constructor now requires a native token address - SDK service updated to pass this into constructor
- Updated balance fetching, this was breaking when connected to SE chain, even when the address was updated
- Only fetching balances for Naan address now, more tokens _TBD_ **NOTE** To add more, we would need to deploy the token addresses as they are on the target chain for `namada.me`, then just iterate over these in the interface
- "NAM"->"Naan" - updated `symbol` of NAM definition to Naan, reference symbol instead of key
- Added `Memo` field to Transfer, IBC Transfer, Bond, & Unbond Tx

### Testing

For everything to work locally
- Install & set up extension, using `apps/namada-interface/.env`:
```
NAMADA_INTERFACE_NAMADA_CHAIN_ID=shielded-expedition.b40d8e9055
NAMADA_INTERFACE_NAMADA_URL=https://proxy.heliax.click/shielded-expedition.b40d8e9055/
```

- In interface, `yarn dev:proxy`
- In extension, navigate to `Network` and set URL to `http://localhost:8010/proxy`
- If you don't have tokens for an account, request them at `faucet.heliax.click`
- Loading and connecting interface should show your token balances
- Submitting Tx should display `Memo` field - I tested these via `tx.rs` in `shared` to ensure they were deserialized correctly and passed to Tx Args
- Essentially, Tx should succeed. If balances don't load on inital connection, you may need to refresh. 
- Token should display as `Naan` and not `NAM`

## TODO

1. I'm looking into an issue when `getChain` does not dispatch correctly on initial connection to a newly installed extension - this is needed to get the correct chain ID, rpc & native token. A refresh gets around it for now
3. Figure out the other Token addresses, and how we can add them via `.env` variables to deploy with Interface. Extension should not be built with any token addresses, as that requires us to resubmit the extension for review


packages/shared/lib/src/sdk/tx.rs
```diff
@@ -445,6 +447,11 @@ fn tx_msg_into_args(tx_msg: &[u8]) -> Result<args::Tx, JsError> {
         _ => None,
     };
 
+    let memo = match memo {
```
mateuszjasiuk, line 450: You can also map Option:
```rs
    let memo = memo.map(|v| v.as_bytes().to_vec());
```

jurevans: Nice, thanks!

jurevans: @mateuszjasiuk RE: You're question about validating Memo - this can be any string, and I did some testing to ensure that it was deserializing into the correct Tx args, however, we will need to validate Tx size when using this field with Ledger, so a large `memo` could cause that to fail. Looking at that in a new PR!

PR #619, opened 2024-02-06
feat: begin replacing redux with jotai

emccorson:
This PR is the first part of replacing redux with jotai in the interface. There are different ways we could handle the data fetching depending on the UI requirements, but it should be easy to change and overall I'm happy enough with how jotai is to use compared to redux.

Currently redux has only been removed for the coins slice, so everywhere else jotai has been added, the existing redux code is still there. Data being fetched through jotai is still being fetched for redux too, so the work is duplicated but I'll fix that in a later PR.

---

### Changed
- Fetch accounts, balances etc. for jotai as well as redux (currently duplicating fetching work but will be fixed later)
- Use jotai instead of redux for displaying data in DerivedAccounts
- Calculate total balance in AccountOverview instead of DerivedAccounts
- Display "-" instead of "0" when balance for an account is loading
- Refactor DerivedAccounts with smaller FiatBalanceDisplay component

### Added
- Add jotai atoms for Namada accounts, balances and chain config
- Add jotai atoms for fiat rates and balanceToFiat function
- Add jotai atom for fiatCurrency setting
- Add temporary jotai atom for reading from redux store

### Removed
- Remove redux coins slice

### Fixed
- Only fetch fiat rates when DerivedAccounts is first mounted and remove apparently broken timestamp code

### DIFF
--- a/apps/namada-interface/src/App/AccountOverview/AccountOverview.tsx
+++ b/apps/namada-interface/src/App/AccountOverview/AccountOverview.tsx
@@ -12,7 +12,7 @@ import { Account, Chain, ExtensionKey, Extensions } from "@namada/types";
 import { formatCurrency } from "@namada/utils";
 import { TopLevelRoute } from "App/types";
 import { AccountsState, addAccounts, fetchBalances } from "slices/accounts";
-import { SettingsState, setIsConnected } from "slices/settings";
+import { setIsConnected } from "slices/settings";
 import { useAppDispatch, useAppSelector } from "store";
 import {
   AccountOverviewContainer,
@@ -27,10 +27,33 @@ import {
 } from "./AccountOverview.components";
 import { DerivedAccounts } from "./DerivedAccounts";
 
+import { atom, useAtomValue, useSetAtom } from "jotai";
+import { loadable } from "jotai/utils";
+import { accountsAtom, balancesAtom } from "slices/accounts";
+import { balanceToFiatAtom } from "slices/coins";
+import { fiatCurrencyAtom } from "slices/settings";
+
 //TODO: move to utils when we have one
 const isEmptyObject = (object: Record<string, unknown>): boolean => {
   return object ? Object.keys(object).length === 0 : true;
 };
+
+const totalBalanceAtom = atom(async (get) => {
+  const accounts = await get(accountsAtom);
+  const balances = get(balancesAtom);
+  const balanceToFiat = await get(balanceToFiatAtom);
+  const fiatCurrency = get(fiatCurrencyAtom);
+
+  return accounts.reduce((acc, account) => {
+    const balance = balances[account.address];
+
+    if (typeof balance === "undefined") {
+      return acc;
+    } else {
+      return acc.plus(balanceToFiat(balance, fiatCurrency));
+    }
+  }, new BigNumber(0));
+});
 
 export const AccountOverview = (): JSX.Element => {
   const navigate = useNavigate();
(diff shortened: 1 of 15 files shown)


apps/namada-interface/src/App/AccountOverview/AccountOverview.tsx
```diff
@@ -27,11 +27,34 @@ import {
...
+import { accountsAtom, balancesAtom } from "slices/accounts";
+import { balanceToFiatAtom } from "slices/coins";
+import { fiatCurrencyAtom } from "slices/settings";
+
 //TODO: move to utils when we have one
 const isEmptyObject = (object: Record<string, unknown>): boolean => {
   return object ? Object.keys(object).length === 0 : true;
 };
 
+const totalBalanceAtom = atom(async (get) => {
```
mateuszjasiuk, line 41: Is this atom defined here because it is specific to this component? :) Asking because most of those are in the slices

apps/namada-interface/src/App/AccountOverview/AccountOverview.tsx
```diff
@@ -27,11 +27,34 @@ import {
...
+  const accounts = await get(accountsAtom);
+  const balances = get(balancesAtom);
+  const balanceToFiat = await get(balanceToFiatAtom);
+  const fiatCurrency = get(fiatCurrencyAtom);
+
+  return accounts.reduce((acc, account) => {
+    const balance = balances[account.address];
+
+    if (balance === undefined) {
```
mateuszjasiuk, line 50: I was always checking with the `typeof`, not exactly sure why. So it turns out the `typeof` check is a bit safer in some edge cases, but `=== undefined` is faster. Anyways, it does not matter in this case and also typescript mitigates some of those :) There are nice answers in the [stackoverflow](https://stackoverflow.com/questions/4725603/variable-undefined-vs-typeof-variable-undefined)

apps/namada-interface/src/slices/accounts.ts
```diff
@@ -135,3 +143,55 @@ const { actions, reducer } = accountsSlice;
...
+////////////////////////////////////////////////////////////////////////////////
+
+const accountsAtom = (() => {
+  const base = atom(new Promise<readonly AccountDetails[]>(() => {}));
+
+  return atom(
+    (get) => get(base),
+    async (_get, set) => {
+      const namada = getIntegration("namada");
+      const accounts = namada.accounts().then(async (accounts) => {
```
mateuszjasiuk, line 158: Is there any particular reason here to use then instead of await?

apps/namada-interface/src/slices/accounts.ts
```diff
@@ -135,3 +143,55 @@ const { actions, reducer } = accountsSlice;
...
+      accounts.forEach(async (account) => {
+        const result = await namada.queryBalances(account.address);
+        const balance = result.reduce(
+          (acc, curr) => ({ ...acc, [curr.token]: new BigNumber(curr.amount) }),
+          {} as Balance
+        );
+        balances[account.address] = balance;
+        set(
+          base,
+          { ...balances } // object clone ensures jotai realises object has changed
```
mateuszjasiuk, line 190: Not sure how helpful that would be in this case but there is an [immer extension](https://jotai.org/docs/extensions/immer) for jotai :)

mateuszjasiuk (approved):
Looks nice! :) Left few questions, but those are not super important

apps/namada-interface/src/slices/accounts.ts
emccorson: In this case it was because I want to reject the accounts promise if accounts is undefined, but I think if I throw an error outside of the `then` callback, it will reject the setter instead. I'm not sure if there is a nicer syntax using await though. Maybe something like:

```
const accounts = (async () => {
  const result = await namada.accounts();
  if (results === undefined) {
    throw new Error("accounts was undefined!");
  }
 return result;
})();

set(base, accounts);
```

apps/namada-interface/src/App/AccountOverview/AccountOverview.tsx
emccorson: Yes, just because it is specific to this component, but I can move it :smile:

mateuszjasiuk: Nah that's fine i think :)

mateuszjasiuk (approved):

jurevans (approved):
LGTM! Everything worked when I tested!

PR #621, opened 2024-02-06
feat: update zip32 path

### LINKED ISSUE #604: Extension: Shielded address derivation update
For our Zip32 derivation of default shielded accounts, we need to update the extension to derive from the following path:

```
m/32'/877'/0'
```

This needs to be verified against imported accounts in the CLI

- User imports a mnemonic via CLI
- User derives a default shielded account - this needs to be equivalent in the CLI

Investigate: What is the best way to migrate existing keys to this derivation path? Perhaps for these accounts, we check the path components in the `path` object for the account, and if `index` is specified, we unspecify it and generate a new key before using.

mateuszjasiuk:
Resolves #604 

### Changes
- Updated path
- Left default payment address
- Changed tests runner to wasm_bindgen_test

### How to test
- import mnemonic in extension
- check viewing key in the vault storage
- import mnemonic in the cli
- check if viewing key in cli is the same as the one in the vault storage


jurevans (approved):
nice, LGTM!

PR #622, opened 2024-02-07
chore: update namada to 0.31.1

mateuszjasiuk:


packages/types/src/tx/tokens/types/index.ts
```diff
@@ -2,7 +2,7 @@ import { RegisteredCoinType, registeredCoinTypes } from "slip44";
 
 const {
   NAMADA_INTERFACE_NAMADA_TOKEN:
-  nativeToken = "tnam1qxgfw7myv4dh0qna4hq0xdg6lx77fzl7dcem8h7e",
```
emccorson, line 5: There are a few more places in the code where this address is hard-coded. Does it need to be changed there too?

emccorson (approved):
Looks good to me and works well! I just left one question.

packages/types/src/tx/tokens/types/index.ts
jurevans: this one will not get used if the extensions `query_native_token` is called, so it's really just a fall back for that case. Otherwise it will always use the `currency.address` from the Chain config returned from the extension, which is added from the token query, in theory

jurevans (approved):
LGTM!

packages/types/src/tx/tokens/types/index.ts
jurevans: so, that said, it doesn't really need to be updated at all. If `Query.query_native_token()` fails, it means the chain is unreachable anyways

mateuszjasiuk: Hmm I think I somethibng did not work for me without this change :D as I did not set native token env at all. But that might've been coincidence. I will double check

mateuszjasiuk: For the sake of completeness I chaned this everywhere :)

PR #623, opened 2024-02-07
fix: Chain service improvements for token

jurevans:
The logic for query native token was a bit incorrect in the `ChainsService` - this has been fixed! The goal of this feature is that the native token will always match the chain that the extension is pointed to, assuming that chain is up and responding, and users should be able to remedy any issue related to this without restarting or upgrading the extension.

### Testing

- Build the extension & interface with pointed to a local node (if you have balances, even better, as it's easier to see it working)
- If your node requires proxying, point the extension to `http://localhost:8010/proxy`
- You should see balances, and you can also inspect the Redux state to confirm the token address, which should be: `tnam1qxgfw7myv4dh0qna4hq0xdg6lx77fzl7dcem8h7e`
- In the Extension, in `Network`, set the values to look like this:
```
shielded-expedition.88f17d1d14
https://proxy.heliax.click/shielded-expedition.88f17d1d14/
```
- Clicking submit, then inspect Redux state, and the `chain.currency.address` should be `tnam1qxvg64psvhwumv3mwrrjfcz0h3t3274hwggyzcee`

You may also test error states, like setting the URL to something that's invalid so that the query fails, and then update with a correct value. An error should result in a chain config that either has an unset `address` or the previous `address`, but upon a successful query, this should always be correct.

**NOTE** The interface is a little buggy, even though the `getChain()`is correct, you will sometimes have to refresh for balances - there are some inefficiencies there, but this PR is to fix the extension

### DIFF
--- a/apps/extension/src/background/chains/service.ts
+++ b/apps/extension/src/background/chains/service.ts
@@ -31,7 +31,7 @@ export class ChainsService {
           const nativeToken = await query.query_native_token();
           defaultChain.currency.address = nativeToken || tokenAddress;
         } catch (e) {
-          console.warn(`Chain is not reachable: ${e}`);
+          console.warn(`Chain is unreachable: ${e}`);
         }
       }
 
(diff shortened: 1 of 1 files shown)


mateuszjasiuk (approved):
Tested! and it works nicely! I have one question though. Is it ok that we always display that the change is successful?
image

jurevans:
> Tested! and it works nicely! I have one question though. Is it ok that we always display that the change is successful? 
I think this is valid, because storage does get updated successfully. If the native token query fails, it still successfully updates the chain info, and we just attempt to query the native token the next time the service is used. I'm not sure if we would want to block them from using this chain info if the request fails, but maybe we do? It would be easy enough to incorporate!

PR #624, opened 2024-02-08
fix: do not display import with passphrase in pk flow

mateuszjasiuk:
When clicking on Private Key we do not want to display: "Import with BIP39 Paassphrase" as it's being ignored anyway.

image


jurevans (approved):
LGTM! Yeah I noticed this too :D

PR #625, opened 2024-02-08
feat: Faucet - set URL to absolute when in production

jurevans:
Use `https://faucet.heliax.click/` for assets when publishing to production


pedrorezende (approved):
Maybe this URL should be defined in a separate .env, but it's exactly what we need for now. LGTM!

Issue #597, opened 2024-01-29
Interface: Implement Gas & Fees for Staking & IBC Tx


jurevans:
In non-Transfer Tx, we are still hardcoded `feeAmount` & `gasLimit` via `txProps`. Following are the locations where these can be found:

IBC:

https://github.com/anoma/namada-interface/blob/2e7dbfd60bf925c2501b22c86a59eada9bc69cf3/apps/namada-interface/src/App/Token/IBCTransfer/IBCTransfer.tsx#L70-L76

Staking:

`Bond`
https://github.com/anoma/namada-interface/blob/2e7dbfd60bf925c2501b22c86a59eada9bc69cf3/apps/namada-interface/src/slices/StakingAndGovernance/actions.ts#L253-L258

`Unbond`
https://github.com/anoma/namada-interface/blob/2e7dbfd60bf925c2501b22c86a59eada9bc69cf3/apps/namada-interface/src/slices/StakingAndGovernance/actions.ts#L290-L295

`Withdraw`
https://github.com/anoma/namada-interface/blob/2e7dbfd60bf925c2501b22c86a59eada9bc69cf3/apps/namada-interface/src/slices/StakingAndGovernance/actions.ts#L318-L325

jurevans:
Closing in favor of #748

PR #626, opened 2024-02-09
feat: staking gas fee improvements

emccorson:
Fixes bond, unbond and withdraw in #597.

---

### Changed
- Use minimum gas price instead of hardcoded price for bond, unbond and withdraw
- Account for fees in maximum valid amount in bonding
- Warn the user when bonding too much to allow for future unbonding and withdrawal
- Refresh balances when a transaction completes (specifically when the toasts atom changes)

### Added
- Add max button to new bonding form
- Display gas fee in bonding and unbonding

### Fixed
- Sum all staking positions when calculating total staked amount
- Allow BigNumber for AmountInput min and max props
- Minor styling fixes

### DIFF
--- a/apps/namada-interface/src/slices/fees.ts
+++ b/apps/namada-interface/src/slices/fees.ts
@@ -1,0 +1,68 @@
+import { Query } from "@namada/shared";
+import BigNumber from "bignumber.js";
+import { atom } from "jotai";
+
+import { accountsAtom } from "slices/accounts";
+import { chainAtom } from "slices/chain";
+
+// TODO: remove harcoding of gas limit
+export const GAS_LIMIT = new BigNumber(20_000);
+
+const minimumGasPriceAtom = (() => {
+  const base = atom(new Promise<BigNumber>(() => {}));
+
+  return atom(
+    (get) => get(base),
+    async (get, set) => {
+      const {
+        rpc,
+        currency: { address: nativeToken },
+      } = get(chainAtom);
+      const query = new Query(rpc);
+
+      const promise = (async () => {
+        const result = (await query.query_gas_costs()) as [string, string][];
+        const nativeTokenCost = result.find(([token]) => token === nativeToken);
+
+        if (!nativeTokenCost) {
+          throw new Error("Error querying minimum gas price");
+        }
+
+        const asBigNumber = new BigNumber(nativeTokenCost[1]);
+
+        if (asBigNumber.isNaN()) {
+          throw new Error("Error converting minimum gas price to BigNumber");
+        }
+
+        return asBigNumber;
+      })();
+
+      set(base, promise);
+    }
+  );
+})();
+
+const isRevealPkNeededAtom = atom(async (get) => {
+  const accounts = await get(accountsAtom);
+  const transparentAccounts = accounts.filter((account) => !account.isShielded);
+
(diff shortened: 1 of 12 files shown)


mateuszjasiuk (approved):
LGTM! Tested and it worked nicely. For some reason in the begining I got information that minimum gas price was not queried, but after few refreshes it worked :D

apps/namada-interface/src/slices/fees.ts
```diff
@@ -0,0 +1,66 @@
...
+
+  const map: Record<string, string | null> = {};
+  transparentAccounts.forEach(async ({ address }) => {
+    const publicKey = await query.query_public_key(address);
+    map[address] = publicKey;
+  });
+
+  return (address: string): boolean => {
+    if (!(address in map)) {
+      throw new Error("address not found in public key map");
```
mateuszjasiuk, line 60: It seems like this is being thrown in situations when you have shieldedf balance unfortunately.
So if you do transfer from transparent to shielded and then refresh and click send buttton :(

apps/namada-interface/src/slices/fees.ts
```diff
@@ -0,0 +1,66 @@
...
+
+const isRevealPkNeededAtom = atom(async (get) => {
+  const accounts = await get(accountsAtom);
+  const transparentAccounts = accounts.filter((account) => !account.isShielded);
+
+  const { rpc } = get(chainAtom);
+  const query = new Query(rpc);
+
+  const map: Record<string, string | null> = {};
+  transparentAccounts.forEach(async ({ address }) => {
```
mateuszjasiuk, line 53: Also if we want to do async/await in loop we need to use `for...of` syntax
```ts
  for await (const { address } of transparentAccounts) { }
```
I think this works because the returned function is not being called immediately

emccorson: Is that the case only if we want each call to the callback to await the previous one finishing?
In this case, I want each call to the callback to run in parallel, but within the callback I need to await the public key in order to use its value.
Basically I'm imagining if it was `map` instead of `forEach`, the return value would be an array of promises which are independent of each other, but since it's `forEach` those promises get thrown away. Have I got this reasoning wrong somewhere?

apps/namada-interface/src/slices/fees.ts
emccorson: Nice spot! That's a silly mistake on my part.

apps/namada-interface/src/slices/fees.ts
mateuszjasiuk: If that's the case then this is fine. It might be a bit confusing because when you want to read the map immediately after the forEach, it's value is `{}` as forEach just calls async functions in loop :P Not sure if we need all of the promises to be resolved, but maybe Promise.all would be a good solution here? :D But yeah if we do not want to wait then this is fine

emccorson: Ah wait, yeah you're totally right there that I should use `Promise.all`, because I was returning the function before the queries have finished. Maybe I need to think about the fetching behaviour a bit more in a later PR, but for now I will just await all public key queries before returning. Thanks for the explanation!

mateuszjasiuk (approved):
Retested and LGTM! :)

PR #628, opened 2024-02-12
tests: approvals tests

mateuszjasiuk:


apps/extension/src/background/approvals/service.test.ts
```diff
@@ -0,0 +1,481 @@
...
+
+      jest.spyOn(service as any, "getPopupTabId").mockResolvedValue(tabId);
+      jest.spyOn(dataStore, "get").mockResolvedValueOnce("data");
+      jest
+        .spyOn(keyRingService, "signArbitrary")
+        .mockResolvedValue(sigResponse);
+      const signer = "signer";
+      const signaturePromise = service.approveSignature(signer, "data");
+
+      await new Promise((resolve) =>
```
emccorson, line 128: What does waiting here do?

apps/extension/src/background/approvals/service.test.ts
```diff
@@ -0,0 +1,481 @@
...
+    });
+  });
+
+  describe("getParamsTransfer", () => {
+    it("should return transfer params", () => {
+      const transferMsgValue = new TransferMsgValue({
+        source: "source",
+        target: "target",
+        token: "token",
+        amount: BigNumber(100),
```
emccorson, line 238: That's cool, I didn't know `new` was optional for BigNumber :smile:

emccorson (approved):
Looks good to me!

apps/extension/src/background/approvals/service.ts
```diff
@@ -100,6 +100,7 @@ export class ApprovalsService {
     if (!data) {
       throw new Error(`Signing data for ${msgId} not found!`);
     }
+    //TODO: Shouldn't we _clearPendingSignature when throwing?
```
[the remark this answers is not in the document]
jurevans: Probably? I don't think it would do anything for `Signing data for ${msgId} not found!`, but on the other errors, it should probably be there.

jurevans (approved):
LGTM!

apps/extension/src/background/approvals/service.test.ts
mateuszjasiuk: Me neither :D

apps/extension/src/background/approvals/messages.ts
```diff
@@ -127,13 +127,17 @@ export class ConnectInterfaceResponseMsg extends Message<void> {
   }
 
   validate(): void {
-    if (!this.interfaceTabId) {
```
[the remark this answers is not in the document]
mateuszjasiuk: Yeah either that or we can just use io-ts also for those checks

apps/extension/src/background/approvals/service.test.ts
mateuszjasiuk: So the reason is `approveSignature` is async but it does not resolve the promise by itself.
Because of that I did not add `await` in front of the call as that would make test hang.
Instead we have to wait a "tick" so `resolverMap` will be filled and `submitSignature` will be able to resolve successfully 😅 It's a bit confusing.

apps/extension/src/background/approvals/service.ts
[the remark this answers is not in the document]
mateuszjasiuk: I will resolve those todos later ;)

PR #632, opened 2024-02-14
bug: select active account and remove fiat currency

### LINKED ISSUE #630: Interface - updates & bugs
Adding relatively small changes to the interface here:

- [ ] Remove conversion to USD/EUR/etc. from interface, display only token balance
  - Remove or disable `coins` slice
  - Remove settings page where currency is selected
  - Remove total balance summing
- [ ] Transfer page is still not setting a selected account when there is only one account, which can result in validation failing for the `amount` field

mateuszjasiuk:
Resolves #630 

To test the "invalid ~~balance~~ amount" bug. First import account without balance and then one with balance.


jurevans (approved):
LGTM!

apps/namada-interface/src/App/AccountOverview/DerivedAccounts/DerivedAccounts.tsx
```diff
@@ -135,6 +107,7 @@ const DerivedAccounts = (): JSX.Element => {
           .map((account) => {
             const { alias, address, isShielded } = account;
             const balance = balances[address];
+            const nativeBalance = typeof balance === "undefined" ? "-" : (balance?.[symbol]?.toString() || "0");
```
emccorson, line 110: The optional chaining on `balance` should be unneeded here.

apps/namada-interface/src/App/AccountOverview/DerivedAccounts/DerivedAccounts.tsx
```diff
@@ -151,7 +124,11 @@ const DerivedAccounts = (): JSX.Element => {
                       )}
                     </DerivedAccountType>
                   </DerivedAccountInfo>
-                  <FiatBalanceDisplay balance={balance} />
+                  <div className="flex items-center">
+                    <span className="text-white">
+                      {Tokens[symbol].symbol} {nativeBalance?.toString()}
```
emccorson, line 129: I think you can remove `?.toString()` here since `nativeBalance` is of type `string`.

emccorson (approved):
Code looks good and works well! I left a couple of minor comments.

apps/namada-interface/src/App/AccountOverview/DerivedAccounts/DerivedAccounts.tsx
mateuszjasiuk: I think balance was undefined sometimes when I was testing it, but I will double check

mateuszjasiuk: Ah right I'm checking that 😅

PR #642, opened 2024-02-21
feat/640 - SDK JS package

### LINKED ISSUE #640: Namada JS SDK
Scoping out a JS SDK wrapper for functionality in `shared`, `types` and probably `crypto` as well.

_Note_ This would be easiest if done from within the monorepo, and could feasibly be adopted by us in our own code.

We can look at `light-sdk` for inspiration, but our API will likely look a little bit different.

SDK & Query could likely be combined - maybe a single class could be instantiated with the RPC settings required for both, or we could just require the RPC endpoint in the Submit calls like `light-sdk`

### SDK
- Build Tx (Transfer, IBC transfer, Eth Bridge  transfer, Bond, Unbond, Withdraw, VoteProposal, RevealPK) - returns bytes
- Sign Tx with Keypair (tx bytes & keys)
- Submit Tx via RPC (final signed tx bytes)
- Append Ledger signature & submit (tx bytes, signature)
- Sign & Verify Arbitrary
- This will wrap the complexity of dealing with serialization, and using string pointer readers in TS for methods that use Zeroize in Rust

### Query

Export Query functions with JS names (instead of Rust)

- Query validators
- Query balances
- Query epoch
- Query gas fees/limit
- Etc.

### Keys

Simple way to generate keypairs for use with Namada

- Mnemonic, Bip44 & Zip32
- Transparent & Shielded Keypairs, Namada addresses

### Ledger HW Wallet

- Move our `Ledger` class into this package, which provides useful integration for the NamadaApp on Ledger

### Signing

- Sign & Verify arbitrary

jurevans:
resolves #640 

- [x] Set up a developer-friendly JS package for interacting with SDK
- [x] Make the "prop" types have consistent naming in `@namada/types`
- [x] Set up unit tests for the API

TODO: 
- [ ] Additional test coverage & documentation
- [ ] Set up publishing for this package


packages/sdk/src/rpc/rpc.ts
```diff
@@ -0,0 +1,212 @@
...
+  async queryNativeToken(): Promise<string> {
+    return await this.query.query_native_token();
+  }
+
+  /**
+   * Query public key
+   * Return string of public key if it has been revealed on chain, otherwise, return null
+   * @async
+   * @param {string} address - Address to query
+   * @returns {string|null} String of public key if found
```
emccorson, line 61: This type doesn't match the TypeScript type. Actually, I thought eslint-plugin-jsdoc should catch that, so it might need to be configured to check against the TypeScript type (that would be really nice to have :smile:).

packages/sdk/src/rpc/types.ts
```diff
@@ -0,0 +1,69 @@
...
+ * GasCosts
+ * [[tokenAddress, gasCost]]
+ */
+export type GasCosts = [[string, string]];
+
+/**
+ * Balance
+ * [[tokenAddress, amount]]
+ */
+export type Balance = [[string, string]];
```
emccorson, line 69: This just allows for a single pair of token address and amount, but if multiple tokens are passed to `queryBalance` there should be multiple address+amount pairs. I was also wondering if it would be slightly nicer user experience to use an object instead of a tuple, so overall something like:

```ts
type Balance = { tokenAddress: string, amount: string }[]
```
I think `GasCosts` will also have the same problem.

packages/sdk/src/rpc/types.ts
```diff
@@ -0,0 +1,69 @@
...
+
+export type Unbonds = {
+  owner: string;
+  validator: string;
+  amount: string;
+  startEpoch: string;
+  withdrawableEpoch: string;
+};
+
+export type StakingPositions = {
```
emccorson, line 42: I think these fields are supposed to be arrays

packages/sdk/src/rpc/rpc.ts
```diff
@@ -0,0 +1,212 @@
...
+  }
+
+  /**
+   * Query bond and unbond details by owner addresses
+   * @async
+   * @param {string[]} owners - Array of owner addresses
+   * @returns {Promise<StakingPositions>} Promise resolving to staking positions
+   */
+  async queryStakingPositions(owners: string[]): Promise<StakingPositions> {
+    const [bonds, unbonds] = await this.query.query_staking_positions(owners);
```
emccorson, line 146: Related to the comment on the `StakingPositions` type, it looks like the type annotations need to be added earlier for TypeScript to catch the problem:
```ts
const [bonds, unbonds]: [BondsResponse[], UnbondsResponse[]] =
  await this.query.query_staking_positions(owners);
```

packages/sdk/package.json
```diff
@@ -0,0 +1,75 @@
...
+    "@ledgerhq/hw-transport-webhid": "^6.28.0",
+    "@ledgerhq/hw-transport-webusb": "^6.28.0",
+    "@namada/chains": "0.2.1",
+    "@namada/crypto": "0.2.1",
+    "@namada/shared": "0.2.1",
+    "@namada/types": "0.2.1",
+    "@namada/utils": "0.2.1",
+    "@zondax/ledger-namada": "^0.0.4",
+    "bignumber.js": "^9.1.1",
+    "bn.js": "^5.2.0",
```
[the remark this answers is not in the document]
mateuszjasiuk: I think it can be removed :)

packages/sdk/src/rpc/rpc.ts
mateuszjasiuk: So there is a typescript recommended ruleset which prevents you from adding type to the args/return values :D
So this:
```ts
  /**
   * Get address and public key from private key
   * @param {string} privateKey - Private key
   * @returns {Address} Address and public key
   */
  getAddress(privateKey: string): Address {
```
becomes this:
```ts
   * Get address and public key from private key
   * @param privateKey - Private key
   * @returns Address and public key
   */
  getAddress(privateKey: string): Address {
```

But this will make jsdoc2md not adding types to md :/ 

We can try to use https://typedoc.org/ instead though

packages/sdk/src/rpc/rpc.ts
mateuszjasiuk: I think we will catch more of those in the 2nd PR as we will replace shared with sdk in the extension :)

packages/sdk/src/rpc/types.ts
mateuszjasiuk: Makes sense! But probably it's worth doing that as separate issue :)

packages/sdk/package.json
```diff
@@ -0,0 +1,75 @@
...
+    "example": "ts-node -r tsconfig-paths/register ./examples/index.ts",
+    "lint": "eslint src --ext .ts,.tsx",
+    "lint:fix": "yarn lint -- --fix",
+    "lint:ci": "yarn lint --max-warnings 0",
+    "wasm:build": "node ./scripts/build.js --release",
+    "wasm:build:multicore": "node ./scripts/build.js --release --multicore",
+    "wasm:build:dev": "node ./scripts/build.js",
+    "wasm:build:dev:multicore": "node ./scripts/build.js --multicore",
+    "wasm:build:node": "./scripts/build-node.sh",
+    "build:docs": "npx jsdoc2md --files ./src/*.ts --configure ./jsdoc2md.json > ./docs/api.md",
```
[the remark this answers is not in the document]
mateuszjasiuk: I'm not sure either :)

packages/sdk/package.json
```diff
@@ -0,0 +1,75 @@
...
+  "name": "@namada/sdk",
+  "version": "0.2.4",
+  "description": "Namada SDK package",
+  "main": "dist/index.js",
+  "types": "dist/index.d.ts",
+  "repository": "https://github.com/anoma/namada-interface/",
+  "author": "Heliax Dev <info@heliax.dev>",
+  "license": "MIT",
+  "scripts": {
+    "example": "ts-node -r tsconfig-paths/register ./examples/index.ts",
```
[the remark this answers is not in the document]
mateuszjasiuk: Should be fixed. I will change examples in next PR anyway so it imports from the `dist`. As this shouldb e closer to the real example.

packages/sdk/docs/api.md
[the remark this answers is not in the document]
mateuszjasiuk: One upside of leaving it in the git is that someone can browse it on github :) Also if we use `plugin:jsdoc/recommended-typescript` then we won't have a problem with types mismatch and eslint should find names mismatches. But I'm not against removing it from git if it makes more sense

packages/sdk/package.json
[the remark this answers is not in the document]
jurevans: maybe I need to just fix my environment, as I don't think it's necessary, but needed to do that locally to make it work. I can try to fix my paths so it works with out it.

packages/sdk/src/rpc/types.ts
emccorson: Resolving as the bug part of this comment is fixed.

packages/sdk/src/rpc/rpc.ts
emccorson: I quickly tried out typedoc and it looks nice. It will also warn us about classes etc. missing from the generated documentation.

emccorson (approved):
Finished review and looks good to me! I think the outstanding comments are just docs or types things that can be sorted out later.

PR #655, opened 2024-03-06
Keplr IBC balances + tokens/integrations refactor

emccorson:
This PR is basically four different things:

1. I've refactored the tokens package to try to remove some of the casts from the codebase and clear up the typing. @jurevans I've integrated some of your work from #631  and I'm hoping the refactors look OK to you! If not, I can just remove my refactor commit from this PR.
2. The ability to decode IBC addresses when querying Keplr balances is added. To test, make sure you have some IBC balances in Keplr for tokens that are present in `CosmosTokens` and then call `queryBalances` on the Keplr integration.
3. The type of `getIntegration` is changed to make it return the specific integration. I've used the [satisfies](https://devblogs.microsoft.com/typescript/announcing-typescript-4-9/#satisfies) operator quite a lot for this, which was a new discovery for me.
```
const namada: Namada = getIntegration("namada"); // previously this would not typecheck
```
4. Keplr atoms for account and balances are written but not hooked up to the IBC form. I didn't add them to the form since I'm expecting the UI is getting changed very soon.

The commits are fairly separable, so if there is anything that shouldn't be added, I can remove it from the PR.


mateuszjasiuk (approved):
It looks good to me, but I think it's worth waiting for Justin as he knows IBC stuff way better.

apps/namada-interface/src/slices/accounts.ts
```diff
@@ -194,48 +201,90 @@ const balancesAtom = (() => {
...
+    (get) => get(base),
+    (_get, set) =>
+      set(
+        base,
+        (async () => {
+          const keplr = getIntegration("cosmos");
+          const accounts = await keplr.accounts();
+
+          // keplr only supports one account
+          return accounts?.[0];
```
jurevans, line 262: Not for this PR specifically, but it would be better if we invoke Keplr with something like(inside the Keplr integration), as this _will_ actually give us access to all accounts at once, assuming the user has approved connections for those chains and they are available in the Keplr extension:
```typescript
(async () => {
  // Define a constant somewhere for all supported chains
  const chains = ["cosmoshub-1", "theta-testnet-1", "osmosis-1"];
  await keplr.enable(chains); // Needs to be invoked when we actually connect Keplr
  const accounts = await Promise.all(chains.map(async (chainId) => {
    return keplr.getKey(chainId);
  }));
    console.log(accounts);
})();
```
We can do that later though, then here we can just return accounts as it is returned from integration!

jurevans (approved):
I think this looks good!

apps/namada-interface/src/slices/accounts.ts
emccorson: Ah, I see! I think I'm getting confused about what an account is on Keplr. Is it correct to say:
- Within Keplr you can have multiple wallets, of which one can be selected at any time
- Each wallet holds multiple addresses (one per chain ID)
- When we call `keplr.getKey(chainID)`, we are requesting the address information for a specific chain ID on the currently selected wallet
- `signer.getAccounts()` just returns the address information for one chain ID for some reason?

In which case I need to fix the jotai atom to be an array of accounts. I got confused by this in the Keplr docs :weary::

>You can get the address/public keys by `getAccounts` method.
It can return the array of address/public key.
But, currently, Keplr extension manages only one address/public key pair.

jurevans: yeah it's weird in Keplr - you can derive custom paths, and I think if you do you will end up with multiple addresses for one chain ID, but I need to confirm that it still does that. I think there will be a use-case for enabling and querying accounts from all chains, but unfortunately that will involve a lot of approval pop-ups for every chain enabled, at least the first time a user connects. You can leave it for now, but either way we should probably use `getKeys` so we have something to display for the account name/alias.

mateuszjasiuk (approved):

PR #656, opened 2024-03-06
fix: unshiedling

mateuszjasiuk:
Bump namada
Fix unshielding
Fix e2e tests(need to double check)

### DIFF
--- a/apps/namada-interface/src/slices/accounts.ts
+++ b/apps/namada-interface/src/slices/accounts.ts
@@ -211,18 +211,12 @@ const balancesAtom = (() => {
         set(base, { ...get(base), [address]: balance });
       });
 
-
-      // Before querying the shielded balances we need to sync the shielded context
-      const shieldedAddresses = accounts
-        .filter((account) => account.isShielded)
-        .map((account) => account.address);
-      await namada.sync(shieldedAddresses);
+      await namada.sync();
 
       const shieldedBalances = await Promise.all(queryBalance(namada, shieldedAccounts, token));
       shieldedBalances.forEach(([address, balance ]) => {
         set(base, { ...get(base), [address]: balance });
       });
-
     }
   );
 })();
--- a/apps/namada-interface/src/slices/accounts.ts
+++ b/apps/namada-interface/src/slices/accounts.ts
@@ -3,7 +3,7 @@ import BigNumber from "bignumber.js";
 import { atom } from "jotai";
 
 import { chains } from "@namada/chains";
-import { getIntegration } from "@namada/integrations";
+import { getIntegration, Integration } from "@namada/integrations";
 import { Account as AccountDetails, ChainKey, TokenType } from "@namada/types";
 
 import { chainAtom } from "slices/chain";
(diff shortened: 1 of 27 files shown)


phy-chain:
@mateuszjasiuk :wave: 
I see that you're adding shielded-sync to the extension, nice !
Is that the reason why currently .insert_spending_key does not work ?
I've tried inserting an xsk in my extension wallet but even if the function call seems to work, nothing actually happens. I dont see the key in my accounts

mateuszjasiuk:
> @mateuszjasiuk 👋 I see that you're adding shielded-sync to the extension, nice ! Is that the reason why currently .insert_spending_key does not work ? I've tried inserting an xsk in my extension wallet but even if the function call seems to work, nothing actually happens. I dont see the key in my accounts

I don't think this is related. Inserting spending key "should" work. We were not using that function really for some time. It's hard to judge your problem without taking a look at what you are trying to do :)

phy-chain:
> > @mateuszjasiuk 👋 I see that you're adding shielded-sync to the extension, nice ! Is that the reason why currently .insert_spending_key does not work ? I've tried inserting an xsk in my extension wallet but even if the function call seems to work, nothing actually happens. I dont see the key in my accounts
> 
> I don't think this is related. Inserting spending key "should" work. We were not using that function really for some time. It's hard to judge your problem without taking a look at what you are trying to do :)

So far, I've added a small "add spending key" flow in the extension, called the `sdk.add_spending_key` function, and after some effort, I got it to "work", meaning that the calls go through without failing/raising an error in the service worker, but then the address does not show anywhere (in my account for instance). Am I missing something ? Should I link it to my account or something ?

apps/namada-interface/src/slices/accounts.ts
```diff
@@ -193,18 +193,55 @@ const balancesAtom = (() => {
...
-        const result = await namada.queryBalances(account.address, [
-          nativeToken || tokenAddress,
-        ]);
-        const balance = result.reduce(
-          (acc, curr) => ({ ...acc, [curr.token]: new BigNumber(curr.amount) }),
-          {} as Balance
-        );
-        set(base, { ...get(base), [account.address]: balance });
+      // Split accounts into transparent and shielded
+      const [transparentAccounts, shieldedAccounts] = accounts.reduce((acc, curr) => {
```
phy-chain, line 197: How do you create an account from inserting a spending key ? 
It  seems that this is my main issue with adding a spending key, it is not registered in the "accounts'

phy-chain: Sure, i''ve pinged you on discord MP !

phy-chain:
@mateuszjasiuk I have tried to run your branch, and my service worker is spamming thousands of requests to my RPC. When I load the page, it starts loading shielded context at block 1, fires queries like mad, reaches height around 15000 or so, then the service worker blows and I have to refresh, goes to inactive, and I never get to the end of the sync. Do you encounter this problem ?

emccorson (approved):
Looks good to me! Just left some minor comments.

mateuszjasiuk:
> @mateuszjasiuk I have tried to run your branch, and my service worker is spamming thousands of requests to my RPC. When I load the page, it starts loading shielded context at block 1, fires queries like mad, reaches height around 15000 or so, then the service worker blows and I have to refresh, goes to inactive, and I never get to the end of the sync. Do you encounter this problem ?

There is no better way for now, we have to query each block. We usually are not testing against long running testnets but rather local chains. This PR works in principle. It will be improved once there are further improvements to the namada sdk

PR #657, opened 2024-03-07
feat: no floating promises in extension

mateuszjasiuk:
Added no floating promises.
In the effects we have to be explicit with void prefix, but I don't think this is a huge deal :)

### DIFF
--- a/apps/extension/src/background/approvals/service.test.ts
+++ b/apps/extension/src/background/approvals/service.test.ts
@@ -94,7 +94,11 @@ describe("approvals service", () => {
     it("should throw an error when popupTabId is not present", async () => {
       jest.spyOn(service as any, "getPopupTabId").mockResolvedValue(undefined);
 
-      expect(service.approveSignature("signer", "data")).rejects.toBeDefined();
+      try {
+        await service.approveSignature("signer", "data");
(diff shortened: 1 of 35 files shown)


apps/extension/src/background/approvals/service.test.ts
```diff
@@ -94,7 +94,11 @@ describe("approvals service", () => {
     it("should throw an error when popupTabId is not present", async () => {
       jest.spyOn(service as any, "getPopupTabId").mockResolvedValue(undefined);
 
-      expect(service.approveSignature("signer", "data")).rejects.toBeDefined();
+      try {
```
emccorson, line 97: I think in these cases you could also just await or return:

```
await expect(service.approveSignature("signer", "data")).rejects.toBeDefined();
```

emccorson (approved):
Looks good to me!

apps/extension/src/background/approvals/service.test.ts
mateuszjasiuk: Yeah I've noticed that a bit later :D

mateuszjasiuk: I fixed that in another PR :)

PR #658, opened 2024-03-07
Feat/fix: wait for tx to resolve/reject using resolversMap

zenodeapp:
Hi there!

So I'm currently working on the IBC shielded transfer app, which communicates with the Namada extension and stumbled upon the following problem:

_Sent transactions to the extension immediately resolve/reject, which make it impossible for the browser to wait for a result and decide whether the tx failed or not._

This PR solves this and also optimizes/fixes/restructures/refactors some of the resolver logic; see more technical details below.

With this fix the frontend is capable of awaiting sent transactions and also able to see whether the user approves or declines the transaction (or closes the window). Declining makes the browser aware of the promises' rejection with an optional message field, while a successful transaction resolves with optional additional details.

ZEN

### DIFF
--- a/apps/extension/src/Approvals/Approvals.tsx
+++ b/apps/extension/src/Approvals/Approvals.tsx
@@ -43,7 +43,7 @@ export const Approvals: React.FC = () => {
   const requester = useRequester();
 
   // Logic for canceling transactions/signatures when closing the window using the close-button.
-  // Perhaps not the cleanest place to have this logic, neither is the shared variable (performUnloadCleanup).
+  // Not really the cleanest place to put this, neither is the use of a shared variable (performUnloadCleanup).
   // TODO: extract this logic, perhaps by creating a separate context.
   useEffect(() => {
     const onUnload = async () => {
--- a/apps/extension/src/Approvals/Approvals.tsx
+++ b/apps/extension/src/Approvals/Approvals.tsx
@@ -1,4 +1,4 @@
-import React, { useState } from "react";
+import React, { useEffect, useState } from "react";
 import { Route, Routes } from "react-router-dom";
 
 import { Container } from "@namada/components";
@@ -12,6 +12,10 @@ import { ApproveTx } from "./ApproveTx/ApproveTx";
 import { ConfirmLedgerTx } from "./ApproveTx/ConfirmLedgerTx";
 import { ConfirmTx } from "./ApproveTx/ConfirmTx";
 import { ConfirmSignature } from "./ConfirmSignature";
+import { RejectSignatureMsg, RejectTxMsg } from "background/approvals";
+import { Ports } from "router";
+import { useRequester } from "hooks/useRequester";
+import { performUnloadCleanup } from "utils";
 
 export enum Status {
   Completed,
@@ -36,6 +40,30 @@ export type SignatureDetails = {
 export const Approvals: React.FC = () => {
   const [details, setDetails] = useState<ApprovalDetails>();
   const [signatureDetails, setSignatureDetails] = useState<SignatureDetails>();
+  const requester = useRequester();
+
+  // Logic for canceling transactions/signatures when closing the window using the close-button.
+  // Perhaps not the cleanest place to have this logic, neither is the shared variable (performUnloadCleanup).
+  // TODO: extract this logic, perhaps by creating a separate context.
+  useEffect(() => {
+    const onUnload = async () => {
+      if(performUnloadCleanup) {
+        if(signatureDetails) {
+          await requester.sendMessage(Ports.Background, new RejectSignatureMsg(signatureDetails.msgId));
+        }
+        
+        if(details) {
+          await requester.sendMessage(Ports.Background, new RejectTxMsg(details.msgId));
+        }
+      }
+    }
+
+    window.addEventListener('beforeunload', onUnload);
+
+    return () => {
+      window.removeEventListener('beforeunload', onUnload);
+    }
+  }, [performUnloadCleanup, signatureDetails, details])
 
   return (
     <Container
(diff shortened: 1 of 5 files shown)


zenodeapp:
**I'll keep track of a changelog here:**

- Feature/fix: transaction-functions in service.ts now leverage the resolverMap to prevent prematurely resolved promises.
- Optimization: remove resolvers after use (garbage collection).
- Refactoring: abstracted logic for creating, resolving and rejecting promises.
- Feature: Generic type parameter <T> added for the createResolver-function.
- Feature: Arbitrary error-handling messages added, defaulting to a "Request rejected"-message.
- Bugfix: small bugfix in error thrown in approveConnectionResponse-function.

zenodeapp:
Alright, think this has most of the logic now.

To explain this PR in simple terms:
- This PR is only for the wallet's extension code.
- The frontend can now wait for a transaction's completion. Visually you could for instance create a loading screen while it awaits the promised result.
- Error handling/messages could therefore be returned for better feedback to the user and give a more streamlined experience.
- Prematurely closing the popup triggers a rejection of the signature or transaction, preventing it to become indeterminate.
- I attempted to refactor the logic and take care of duplicate code. Though do tell me if I made it too tightly coupled.

emccorson:
Thanks for the contribution and detailed explanation! I should hopefully be able to review this next week if no one else gets there first.

One thing to mention is that it is already possible to know if a transaction has completed, because the extension fires an event (`"namada-tx-completed"`, I _think_). But I'll think about how this compares to being able to get a response from the Promise.

zenodeapp:
> Thanks for the contribution and detailed explanation! I should hopefully be able to review this next week if no one else gets there first.
> 
> One thing to mention is that it is already possible to know if a transaction has completed, because the extension fires an event (`"namada-tx-completed"`, I _think_). But I'll think about how this compares to being able to get a response from the Promise.

Ah interesting, didn't know this! That's also indeed a way to go at it! Take your time and no pressure/biggy if you decide this won't make the cut :)!

zenodeapp:
Just a small post-refactoring I'll do. I think it's better to separate the logic of creating a window/popup and attaching a resolver to it. Will commit this after getting out of bed 😴.

zenodeapp:
> Just a small post-refactoring I'll do. I think it's better to separate the logic of creating a window/popup and attaching a resolver to it. Will commit this after getting out of bed 😴.

Alright done.

- I was contemplating to call: `isResolver` => `pending` or `unresolved` since it actually is an unresolved promise if it's still present in the resolverMap. Though kept it as `isResolver`. 
- Also, I wonder if it's actually needed to do this check at the top of `submitTx`, `submitSignature`, `rejectSignature` and `rejectTx` functions. Last commit I re-introduced it back in the `rejectSignature` and `rejectTx` functions. I accidentally removed this extra check previously. The only reason to include it is to be able to get the error thrown: `no resolvers found for tab ID ${popupTabId}` if for some reason the tabId isn't available in the resolversMap anymore. If we don't include this check then it 'fails' in silence since the `resolve` and `reject` functions already check whether they can fulfill the promise or not.

But yes, just a couple of contemplations. In my opinion the check seems a bit overkill, could perhaps remove the `isResolver` function in its entirety. Don't think it causes any troubles if I go through the code. But I might be wrong!

PS: It might actually be necessary to not include this check if I think of this TODO comment: `//TODO: Shouldn't we _clearPendingSignature when throwing?`. So maybe we should remove the whole `isResolver` check.

zenodeapp:
> > Just a small post-refactoring I'll do. I think it's better to separate the logic of creating a window/popup and attaching a resolver to it. Will commit this after getting out of bed 😴.
> 
> Alright done.
> 
> * I was contemplating to call: `isResolver` => `pending` or `unresolved` since it actually is an unresolved promise if it's still present in the resolverMap. Though kept it as `isResolver`.
> * Also, I wonder if it's actually needed to do this `isResolver`-check at the top of `submitTx`, `submitSignature`, `rejectSignature` and `rejectTx` functions. Last commit I re-introduced it back in the `rejectSignature` and `rejectTx` functions as I accidentally removed this extra check previously. The only reason to include it is to be able to get the error thrown: `no resolvers found for tab ID ${popupTabId}` if for some reason the tabId isn't available in the resolversMap anymore. If we don't include this check then it 'fails' in silence since the `resolve` and `reject` functions already check whether they can fulfill the promise or not.
> 
> But yes, just a couple of contemplations. In my opinion the check seems a bit overkill, could perhaps remove the `isResolver` function in its entirety. Don't think it causes any troubles if I go through the code. But I might be wrong!
> 
> PS: It might actually be necessary to not include this check if I think of this TODO comment: `//TODO: Shouldn't we _clearPendingSignature when throwing?`. So maybe it is necessary to remove the whole `isResolver` check as to not cause any loose `pendingSignature`- or `pendingTx`-data to dangle around. But tell me if this check is obsolete, then I'll push a new commit removing this in its entirety. _(Only if this PR gets merged of course. Though parts might be useful, especially since the signature-functions already implemented promises)._

Okay final option lol, which may solve the whole TODO while still including the error throwing if the tabId isn't around: Remove the isResolver in its entirety but add the thrown error message in both `resolve` and `reject` functions as an else-case, like so:

```js
  private resolve(popupTabId: number, result?: unknown) {
    if (popupTabId in this.resolverMap) {
      this.resolverMap[popupTabId].resolve(result);
      delete this.resolverMap[popupTabId];
    } else {
      throw new Error(`no resolvers found for tab ID ${popupTabId}`);
    }
  }

  private reject(popupTabId: number, message?: any) {
    if (popupTabId in this.resolverMap) {
      this.resolverMap[popupTabId].reject(message ?? new Error("Request rejected"));
      delete this.resolverMap[popupTabId];
    } else {
      throw new Error(`no resolvers found for tab ID ${popupTabId}`);
    }
  }
```

 This does require the resolving or rejecting to always happen AFTER clearing pending data..



Choices, choices 😅. I hope the bombardment of details doesn't make ya go nuts.

zenodeapp:
Merged latest into this PR. Luckily already fixed the things that got pushed to main.

zenodeapp:
Oh wow, forgot bout this!

Closing this!

PR #659, opened 2024-03-07
fix: make publicKey a required field in the TxMsgValue class.

zenodeapp:
Hi again!

As I was working on #658 I couldn't seem to understand why the transactions weren't resolving. Checking the console of the extension I realized I didn't pass a publicKey value to my `sendIbcTransfer` causing the extension to panic! I thus made it a required field in the TxMsgValue class. This way it doesn't even get to the extension when a transaction doesn't contain this.

### DIFF
--- a/packages/types/src/tx/schema/tx.ts
+++ b/packages/types/src/tx/schema/tx.ts
@@ -17,8 +17,8 @@ export class TxMsgValue {
   @field({ type: "string" })
   chainId!: string;
 
-  @field({ type: option("string") })
-  publicKey?: string;
+  @field({ type: "string" })
+  publicKey!: string;
 
   @field({ type: option("bool") })
   disposableSigningKey?: boolean;


zenodeapp:
Okay, no wait. This does not match with the SDK.

Hmm. Is there a way to make it known that this is a required field without it panicking like this if it's not set? (or see alternative above).


Also, this is because we're halted now, but this is a similar panic. Is there a way to get these panics to the frontend browser and not inside of the extension's console?

image

zenodeapp:
I'll keep this PR here to keep this open 'as an issue'.

zenodeapp:
Will close this, I realized during testing that this only panics in development mode, not in the current extension (production mode)!

mateuszjasiuk:
> I'll keep this PR here to keep this open 'as a discussion/issue'.

Hello! Thank you for your c ontribution! Can you please close this one and open an issue? We will try to refine the "backlog" regularly and discuss potential problems there. Thanks! We should probably add contrib instructions to the README. We just have a lot of things to do atm.

phy-chain:
@zenodeapp You need to push a message from the service worker, into the main window, then you'd be able to log the errors in the frontend. However, you'd want to enable that only in dev mode

zenodeapp:
> > I'll keep this PR here to keep this open 'as a discussion/issue'.
> 
> Hello! Thank you for your c ontribution! Can you please close this one and open an issue? We will try to refine the "backlog" regularly and discuss potential problems there. Thanks! We should probably add contrib instructions to the README. We just have a lot of things to do atm.

Yes I shall! This is probably not a big prio as I didn't realize it was only in development mode.

But @phy-chain that sounds like what needs to be done indeed. I'm not sure if it gets handled correctly in production mode either though, I'll test this later.

But yes will close this!

PR #660, opened 2024-03-07
fix: action button fill

RuslanGlaznyov:
# What does this PR do? 
Fix the yellow border when the action button is filled

# Context 
This PR is related to this [PR](https://github.com/anoma/namada-interface/pull/517).  Tailwind migration caused a lot of changes to the code base since my last commit, so I decided to create a new PR and close the old one  

# Externals 
## Before fix: 
before

## After fix: 
after

### DIFF
--- a/packages/components/src/ActionButton.tsx
+++ b/packages/components/src/ActionButton.tsx
@@ -2,31 +2,21 @@ import clsx from "clsx";
 import { createElement } from "react";
 import { tv, type VariantProps } from "tailwind-variants";
 
-const colors = {
-  primary: "yellow",
-  secondary: "cyan",
-  black: "black",
-} as const;
-
-type HoverColor = {
-  hoverColor?: keyof typeof colors;
-};
-
 const actionButton = tv({
   base: clsx(
     `group flex items-center cursor-pointer min-h-[2em] overflow-hidden text-black text-center relative`,
-    `[transition:color_0.1s,background_0.5s] ease-out-circ w-full select-none justify-center font-medium`,
+    `hover-fill w-full select-none justify-center font-medium`,
     `active:top-px`
   ),
 
   variants: {
     color: {
       primary:
-        "bg-100-200 bg-gradient-to-b from-50% to-50% from-yellow to-black hover:bg-bottom hover:text-yellow",
+        "bg-100-200 bg-gradient-to-b from-50% to-50% from-yellow to-black hover:bg-0-99 hover:text-yellow",
       secondary:
-        "bg-100-200 bg-no-repeat bg-gradient-to-b from-50% to-50% from-cyan to-black hover:bg-bottom hover:text-cyan",
+        "bg-100-200 bg-no-repeat bg-gradient-to-b from-50% to-50% from-cyan to-black hover:bg-0-99 hover:text-cyan",
       black:
-        "bg-100-200 bg-no-repeat bg-gradient-to-b from-50% to-50% from-black to-yellow hover:bg-bottom text-yellow hover:text-black",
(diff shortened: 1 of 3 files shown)


packages/components/src/ActionButton.tsx
```diff
@@ -2,18 +2,31 @@ import clsx from "clsx";
...
+
+type HoverColor = {
+  hoverColor?: keyof typeof colors;
+};
+
 const actionButton = tv({
   base: clsx(
     `group flex items-center cursor-pointer min-h-[2em] overflow-hidden text-black text-center relative`,
-    `transition-colors duration-100 w-full select-none justify-center font-medium`,
+    `[transition:color_0.1s,background_0.5s] ease-out-circ w-full select-none justify-center font-medium`,
```
pedrorezende, line 18: Unfortunately, `[transition:color_0.1s,background_0.5s]` overrides the easing :( Maybe we could just force the `ease-out-circ` prefixing it with a `!`. Do you know a better solution rather than write the whole cubic-bezier function here?

packages/components/src/ActionButton.tsx
```diff
@@ -113,11 +106,18 @@ export const ActionButton = ({
   hoverColor,
   ...props
 }: ActionButtonProps<keyof React.ReactHTML>): JSX.Element => {
+  // depending on the props, we can conditionally apply a class
+  const hoverColorClass =
```
pedrorezende, line 110: The outline hover is not working anymore. To check that, you can try to create a new key. The "Import Existing Key" should have a black hover. Regarding this piece of  code, maybe it could be placed as part of the tailwind-variants definitions (https://www.tailwind-variants.org/ for more info). Maybe you could use the compound variants section: https://www.tailwind-variants.org/docs/variants#compound-variants

packages/components/src/theme.js
```diff
@@ -1,5 +1,11 @@
 module.exports = {
   theme: {
+    backgroundSize: {
+      "100-200": "100% 220%",
+    },
+    backgroundPosition: {
+      bottom: "0% 99%", // flickering on firefox
```
pedrorezende, line 7: Hmm, I'm not sure if we should override the bottom position with `0 99%`. This seems to lead to other problems in the future.

pedrorezende (changes requested):
Hey @RuslanGlaznyov ! Thanks again for submitting a PR :) This update is really appreciated and it was a very good catch! I've just made a few comments on things I've spotted

packages/components/src/ActionButton.tsx
RuslanGlaznyov: I created hover-fill class in base.css to apply easing

packages/components/src/ActionButton.tsx
RuslanGlaznyov: It turns out to be much easier. I just had to add this to variants: 
```css
 hoverColor: {
      primary: "to-yellow",
      secondary: "to-cyan",
      black: "to-black",
},
  ```
  The outlined button was fixed as well

RuslanGlaznyov:
Hey @pedrorezende! Thanks for your review, I've provided fixes for your comments. Please check when you have a chance. Thanks

pedrorezende:
Hello @RuslanGlaznyov . Thank you very much for this PR and for pointing out this issue. We've made a huge refactor in the ActionButton component last week, and addressed this hairline behaviour using a different approach (#921) . We're sorry for the delay in getting back to you about this, this PR is really good, and we had planned to revisit this as soon as we had some time to make this refactor. We've done this, but we were able to find a simpler approach using css grids. We really appreciate this contribution, and apologize again for the lack of response.

PR #665, opened 2024-03-13
feat: sdk js publish

### LINKED ISSUE #664: Namada SDK publish
### TODO
- [x] build and publish node version
  - [x] create proper node build script(with and without multicore FF)
  - [x] check publish script
- [x] build and publish web version
- [x] make both versions available for import at the same time
  - [ ] ~~make index separate entrypoint from init functions~~
  - [x] fix typings
- [x] use sdk-js-web in the extension

### Next issue

- [ ] ~~make examples use `dist`~~
- [ ] create sdk-examples package
- [ ] use sdk-js-node in the CLI tool
  - [ ] add "node" feature and implement persistence differently
- [ ] add [noUncheckedIndexedAccess](https://www.typescriptlang.org/tsconfig#noUncheckedIndexedAccess) ?
- [ ] replace js snippets with rust libraries in shared

mateuszjasiuk:
Resolves #664

### Changes
- changed lint staged, so
  - it does not lint files outside `src`
  - lint is called with `max-warnings` 0 so we do not commit "invalid" code(fix can't fix everything :))
- removed "local" dependencies from package.json -> they should be resolved only by using typescript paths(this makes building packages for publish way easier)
- separated wasm node builds to: production and dev
- replaced "shared" dependency in extension with "sdk"
- replaced jsdoc2md with typedoc
- added ts-patch with path transform plugin - it's needed because tsc does not change path aliases to relative paths
- added web and nodejs features to the shared rust package
- split masp into masp_web/masp_node and wallet to wallet_web/wallet_node

### Testing
Most important is to test all the extension functionality :)
You can test publish by running `yarn release:np-npm` - it will take a while for `sdk` as it builds wasms for both web and nodejs in production mode


packages/sdk/docs/README.md
```diff
@@ -0,0 +1,54 @@
+@namada/sdk / [Exports](modules.md)
```
mateuszjasiuk, line 1: Need to update readme a bit

packages/sdk/package.json
```diff
@@ -2,25 +2,59 @@
...
+      "require": "./dist/node/sdk/src/initNode.js",
+      "types": "./dist/node/sdk/src/initNode.d.ts"
+    }
+  },
   "repository": "https://github.com/anoma/namada-interface/",
   "author": "Heliax Dev <info@heliax.dev>",
   "license": "MIT",
+  "scriptsComments": {
+    "prepublish": "We make focused install to make sure that we specified all required dependencies",
+    "bundle-dts": "TODO: We bundle dts based on indexWeb.d.ts, but it should have the same api as indexNode.d.ts",
```
mateuszjasiuk, line 28: Remove this

packages/shared/lib/Cargo.toml
```diff
@@ -34,6 +37,7 @@ wasm-bindgen-rayon = { version = "1.0", optional = true }
 console_error_panic_hook = "0.1.6"
 zeroize = "1.6.0"
 hex = "0.4.3"
+reqwest = "0.11.25"
```
mateuszjasiuk, line 40: Check if reqwest if used

packages/types/src/tx/messages/index.ts
```diff
@@ -22,4 +22,14 @@ export class Message<T extends Schema> implements IMessage<T> {
       throw new Error(`Unable to deserialize message: ${e}`);
     }
   }
+  public static dec<T extends Schema>(
```
mateuszjasiuk, line 25: Change name to "decode" or  sth

packages/shared/scripts/build.js
```diff
@@ -1,7 +1,13 @@
 const { parseArgs } = require("node:util");
 const { spawnSync, execSync } = require("child_process");
 
+const tagets = ["web", "nodejs"];
```
jurevans, line 4: small typo here -> `targets`

jurevans (approved):
LGTM! Extension still works, just left one very minor comment!

jurevans (approved):
Looks good!

packages/sdk/docs/README.md
```diff
@@ -0,0 +1,51 @@
...
+yarn wasm:build[:node]:dev
+
+# Build wasm with debugging for multicore
+yarn wasm:build[:node]:dev:multicore
+```
+
+## Usage
+
+```typescript
+import { Sdk } from "@namada/sdk"
```
emccorson, line 26: Does this example still need updated?

packages/sdk/docs/classes/Rpc.md
```diff
@@ -0,0 +1,413 @@
...
+[sdk/src/rpc/rpc.ts:82](https://github.com/anoma/namada-interface/blob/7542445f/packages/sdk/src/rpc/rpc.ts#L82)
+
+___
+
+### queryPublicKey
+
+▸ **queryPublicKey**(`address`): `Promise`\<`undefined` \| `string`\>
+
+Query public key
+Return string of public key if it has been revealed on chain, otherwise, return null
```
emccorson, line 246: Type signature is fixed now but comment says "return null".

packages/sdk/src/mnemonic.ts
```diff
@@ -54,15 +56,18 @@ export class Mnemonic {
   /**
    * Validate a mnemonic string, raise an exception providing reason
```
emccorson, line 57: I think this comment is outdated now that the return type has changed.

packages/sdk/src/rpc/rpc.ts
```diff
@@ -204,11 +204,15 @@ export class Rpc {
...
+   * @returns void
    */
-  async broadcastTx(signedTx: SignedTx): Promise<void> {
+  async broadcastTx(signedTx: SignedTx): Promise<string> {
     const { txMsg, tx } = signedTx;
-    await this.sdk.process_tx(tx, txMsg);
+    return await this.sdk.process_tx(tx, txMsg);
+  }
+
+  async shieldedSync(vks: string[]): Promise<void> {
```
emccorson, line 215: Missing docs here.

packages/sdk/src/rpc/rpc.ts
```diff
@@ -204,11 +204,15 @@ export class Rpc {
   /**
    * Broadcast a Tx to the ledger
    * @async
-   * @param {SignedTx} signedTx - Transaction with signature
-   * @returns {void}
+   * @param signedTx - Transaction with signature
+   * @returns void
```
emccorson, line 208: Now returns string instead of void.

packages/integrations/src/Namada.ts
```diff
@@ -87,10 +87,14 @@ export default class Namada implements Integration<Account, Signer> {
     tokens: string[] = []
   ): Promise<TokenBalances> {
     const balances = (await this._namada?.balances({ owner, tokens })) || [];
+    console.log(balances);
```
emccorson, line 90: Maybe meant to remove?

packages/crypto/.eslintrc.js
emccorson, line 1: There is no `lint:ci` script for crypto, so AFAIK this will not get used unless manually called.

emccorson: Ah wait, I just realised it must be for lint staged :smile:

packages/sdk/tsconfig.web.json
```diff
@@ -0,0 +1,22 @@
...
+    "node_modules",
+    "scripts"
+  ],
+   "compilerOptions": {
+    "declaration": true,
+    "outDir": "dist/web",
+    "module": "esnext",
+    "moduleResolution": "bundler",
+    "plugins": [
+      // Transform paths in output .js files
```
emccorson, line 15: Probably a silly question, but does this still work even though comments are not allowed in JSON files? :smile:

apps/extension/src/background/keyring/service.ts
```diff
@@ -423,12 +479,6 @@ export class KeyRingService {
       await this.broadcaster.updateBalance();
     } catch (e) {
       console.warn(e);
-      await this.broadcaster.completeTx(
```
emccorson, line 426: Why does the event for tx failure get removed here?

emccorson (commented):
Looks really nice! I found possibly one bug, but otherwise just minor comments/questions.

tsconfig.base.json
```diff
@@ -29,7 +29,13 @@
       "@namada/types/*": ["../../../packages/types/src/*"],
 
       "@namada/utils": ["../../../packages/utils/src"],
-      "@namada/utils/*": ["../../../packages/utils/src/*"]
+      "@namada/utils/*": ["../../../packages/utils/src/*"],
+
+      "@namada/sdk/web": ["../../../packages/sdk/src/indexWeb.ts"],
```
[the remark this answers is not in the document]
mateuszjasiuk: Yeah! That's the idea

apps/extension/src/provider/messages.ts
```diff
@@ -1,4 +1,4 @@
-import { SupportedTx } from "@namada/shared";
+import type { SupportedTx } from "@namada/sdk/web";
```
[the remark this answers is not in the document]
mateuszjasiuk: I do not remember if there was a problem here specifically, but importing type only helps avoiding circular dependencies :)

apps/extension/src/background/sdk/service.ts
```diff
@@ -1,34 +1,30 @@
...
 
-    return new Sdk(rpc, chain.currency.address || tokenAddress);
+    return new SdkService(sdk);
   }
 
-  async getQuery(): Promise<Query> {
-    const rpc = await this.getRpc();
-    return new Query(rpc);
+  getSdk(): Sdk {
+    return this.sdk;
```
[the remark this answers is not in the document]
mateuszjasiuk: RIght! It will "fix" itself after another init. I will take a look how to make it a bit better :)

packages/sdk/package.json
```diff
@@ -2,25 +2,58 @@
...
   "version": "0.2.4",
   "description": "Namada SDK package",
-  "main": "dist/index.js",
-  "types": "dist/index.d.ts",
+  "exports": {
+    "./web": {
+      "import": "./dist/web/sdk/src/indexWeb.js",
+      "types": "./dist/web/sdk/src/indexWeb.d.ts"
+    },
+    "./web-init": {
```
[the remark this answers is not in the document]
mateuszjasiuk: It's done like that because otherwise there was some problem in unit tests as we build wasm for tests as nodejs target. When we couple init with index, TS will complain that module produced by wasm pack does not export some init functions. Something like that :D So it's not really "needed" outside of tests. Maybe we could improve that in future by running tests in browser or sth?

packages/sdk/src/initWeb.ts
```diff
@@ -0,0 +1,43 @@
+// We have to use relative improts here othewise ts-patch is getting confused and produces wrong paths after compialtion
+import { init as initCrypto } from "../../crypto/src/init";
+import { init as initShared } from "../../shared/src/init";
+import { initThreadPool } from "../../shared/src/init-thread-pool";
+
+/**
+ * Initialize the SDK memory
+ * @async
+ * @returns
```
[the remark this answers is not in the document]
mateuszjasiuk: You are right, prettier does not catch this case though :( Maybe it assumes that return can be described using multiple lines?

packages/sdk/src/indexNode.ts
```diff
@@ -0,0 +1,47 @@
+import { Query as QueryWasm, Sdk as SdkWasm } from "@namada/shared";
+import { webcrypto } from "node:crypto";
+import { Sdk } from "./sdk";
+export * from "./index";
+export * from "./utils";
+
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+(globalThis as any).crypto = webcrypto;
```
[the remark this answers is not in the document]
mateuszjasiuk: It's a polyfill to share crypto implementation between browser and nodejs

packages/sdk/docs/README.md
mateuszjasiuk: I've updated README here https://github.com/anoma/namada-interface/pull/695/files :)

packages/sdk/src/mnemonic.ts
mateuszjasiuk: Yeah we should avoid mentioning any type information in the description :D

packages/sdk/src/rpc/rpc.ts
mateuszjasiuk: This unfortunately won't be caught by jsdoc plugin :(

packages/sdk/src/rpc/rpc.ts
mateuszjasiuk: This is weird, jsdoc plugin detects that there is "wrong" description but not that there is none :| I'm pretty sure it worked in the beginning :D

packages/sdk/src/rpc/rpc.ts
mateuszjasiuk: I changed return description to optional :)

packages/sdk/tsconfig.web.json
mateuszjasiuk: I can tell you more! :) Even trailing commas work, the reason is that typescript uses it's own parser :P

apps/extension/src/background/keyring/service.ts
mateuszjasiuk: Looks like I removed it by accident :D Thanks!

jurevans (approved):
Changes look good to me!

Issue #662, opened 2024-03-12
Fetching shielded balance via SDK nor extension working


zenodeapp:
Hey,

![image](https://github.com/anoma/namada-interface/assets/108588903/08ad53a9-e713-4d22-a880-81f65d47c2ae)

Whenever I attempt to query a shielded balance for a viewing key via the SDK, I get just a lot of POST request to my rpc endpoint. This just continue to go on. While I don't mind to wait whether it will actually result in something, this will definitely not be feasible for a multiple user front-end.

The extension way gives a:
`Error: No results have been found on Vault`-error.

Again, this is probably due to the current shielded-sync approach and you guys are likely working at it. But wanted to express this as to see whether it's a me-problem or a common one amongst us all.

ZEN

zenodeapp:
Close this asap if it's already a known issue. Don't wanna let the issues tab become bloated like the main `namada`-repository 😅.

thousandsofthem:
Extension without balances is pretty useless, if i can't see own balance then i can't use shielded transactions. Don't work for me as well

pnam1609:
This is my PR to reduce time fetching MASP. Now i still working on it https://github.com/anoma/namada-interface/pull/667

mateuszjasiuk:
Hello. To give you more context about the issue. The problem is that atm we have to query each block to produce shielded context which takes a long time. For SE, as of now, it means querying 160 000 blocks. If you use interface(namada.me) this won't work as we timeout query function after some time(there was a reason for that which is not really that important to this conversation :)).

What is more:
- if you did not do any shileding/unshielding this means querying 160 000 each time as context is only saved when it detects some change in relation to viewing keys that you provided. This also means that once you add another shielded account you have to query everything from the beginning
- if you made shielding transfer at block 100 000, and once you fetched shielded context, this means that every next shielded context fetch will start from block 100 000
- currently published extension uses Namada SDK in version 0.31.8 which is outdated and some of the functionalities might not work with the SE

Upsides:
- we have a [0.32.0 bump PR](https://github.com/anoma/namada-interface/pull/673) which removes the timeout for `query_balance`
- there was a recent change to the Namada SDK which allows you to pass a block index that is a starting index of the shielded context fetch, which allows to write some improvements, like storing index of last fetched block in frontend or whatnot

zenodeapp:
[...]

PR #667, opened 2024-03-13
Feat: Reduce time when fetching masp

pnam1609:
### Issue
Link: [Fetching shielded balance via SDK nor extension working](https://github.com/anoma/namada-interface/issues/662)

### Solution
- First, Make fetches by block ranges instead of all blocks.
- Second, Use Promise pool from @supercharge/promise-pool library to process multiple fetch block ranges in parallel. It significantly reduced time when fetching block.
- Then save the context of the rage block to a new temporary Indexed DB with key as startBlock-EndBlock and value as context after serializing this ranges block.It void loss data if the process is down. Only need fetching again missed ranges block.
- Once finished fetching and saving all ranges blocks to temporary Indexed DB . Will call a function to get all the data from temporary Indexed DB  and load it into the context. 
- The last step is saved to the root db.
- Also the solution help get the progress of process and display in to user.

### Image
- Button to naviagte shielded sync page
image
- Shielded sync page
image

Note:
- I have another change for Namada source. And change depency in cargo.toml to my change. PR: https://github.com/anoma/namada/pull/2898

### DIFF
--- a/apps/namadillo/package.json
+++ b/apps/namadillo/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@namada/namadillo",
-  "version": "1.0.4",
+  "version": "1.0.5",
   "description": "Namadillo",
   "repository": "https://github.com/anoma/namada-interface/",
   "author": "Heliax Dev <info@heliax.dev>",
--- a/apps/namadillo/src/App/Governance/ViewJson.tsx
+++ b/apps/namadillo/src/App/Governance/ViewJson.tsx
@@ -36,7 +36,7 @@ type PgfIbcTargetJson = {
 };
 
 type PgfPaymentData = {
-  continuous: PgfTargetJson[];
+  continuous: (PgfTargetJson | PgfIbcTargetJson)[];
   retro: (PgfTargetJson | PgfIbcTargetJson)[];
 };
 
@@ -93,7 +93,13 @@ const formatData = (proposal: Proposal): DataJson => {
       const continuous = [
         ...pgfActions.continuous.add,
         ...pgfActions.continuous.remove,
-      ].map(formatPgfTarget);
+      ].map((item) => {
+        if ("internal" in item) {
+          return formatPgfTarget(item);
+        } else {
+          return formatIbcPgfTarget(item);
+        }
+      });
 
       const retro = pgfActions.retro.map((retroItem) => {
         if ("internal" in retroItem) {
--- a/apps/namadillo/src/App/Governance/VoteInfoCards.tsx
+++ b/apps/namadillo/src/App/Governance/VoteInfoCards.tsx
@@ -2,7 +2,13 @@ import { NamCurrency } from "App/Common/NamCurrency";
 import { twMerge } from "tailwind-merge";
 
 import { SkeletonLoading } from "@namada/components";
-import { AddRemove, PgfActions, Proposal } from "@namada/types";
+import {
+  AddRemove,
+  PgfActions,
+  PgfIbcTarget,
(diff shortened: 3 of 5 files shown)


apps/extension/package.json
```diff
@@ -40,13 +40,16 @@
     "@ledgerhq/hw-transport": "^6.30.0",
     "@ledgerhq/hw-transport-webhid": "^6.28.0",
     "@ledgerhq/hw-transport-webusb": "^6.28.0",
+    "@material-ui/core": "^4.12.4",
```
phy-chain, line 43: Is that necessary ? It drags an old version babel/runtime with it, which combine are quite huge, and an open door for potentail security issues through babel runtime : https://github.com/advisories/GHSA-67hx-6x53-jw92

phy-chain:
Ping me on Discord so we can chat more easily, I want to help ! @phychain

mateuszjasiuk:
@pnam1609 I really appriciate the effort. I think it will be hard to merge this soon. The reason is you worked under some assumptions about how things should work and look like. The proper flow would be:
- create an issue
- discuss solution in the created issue
- settle on requirements
- work on functionality
- create PR

TuanTQ15:
Cool, thank you for your hard work. We need it ASAP

pnam1609:
Hi @mateuszjasiuk , Thanks for your response. I hope it's available as soon as possible and i can contribute more for project.

nodeopsdev:
@pnam1609 Really nice ! I appreciate your comprehensive solution, which outlines a systematic approach to optimize block fetching by implementing block ranges, utilizing Promise pools for parallel processing, and ensuring data integrity through temporary IndexedDB storage. Your solution not only enhances efficiency by reducing fetch times but also provides a robust mechanism for handling interruptions and displaying progress to the user. Awesome valuable contribution.

mateuszjasiuk:
Hello,
I think if we want to merge it quicker we have to chage few things :):

1. UI - We did not have a plan to create Shielded Sync button. User should not need to remember to do shielded sync every time they want to, for example, see up to date balance. So for the interface we call the "sync" each time we query for the shielded balance. So in general I would move all the UI changes to the separate issue so we can discuss it further :)

2. Changes to Namada repo - I don't think your changes will be merged in near future or at all. On the upside, there is a recent change in the Namada SDK(0.32.0) that allows you to specify the starting index of the [fetch function](https://github.com/anoma/namada/blob/5e0b1620f7f1bf68f9d7ebcef1a624eb60dba7cd/crates/sdk/src/masp.rs#L711), which should help with your solution.

WDYT?

pnam1609:
> @pnam1609
> 
> Hello, I think if we want to merge this quicker we have to chage few things :):
> 
> 1. UI - We did not have a plan to create Shielded Sync button. User should not need to remember to do shielded sync every time they want to, for example, see up to date balance. So for the `namada.me` we call the "sync" each time we query for the shielded balance. So in general I would move all the UI changes to the separate issue so we can discuss it further :)
> 2. Changes to Namada repo - I don't think your changes will be merged in near future or at all. On the upside, there is a recent change in the Namada SDK(0.32.0) that allows you to specify the starting index of the [fetch function](https://github.com/anoma/namada/blob/5e0b1620f7f1bf68f9d7ebcef1a624eb60dba7cd/crates/sdk/src/masp.rs#L711), which should help with your solution.
> 
> WDYT?

pnam1609:
Hi @mateuszjasiuk ,
1. For this shielded sync button and UI for that. Let me remove this.
2. I will change the dependency and will check to logic code again.

PR #671, opened 2024-03-19
Dispatch events by origin + add tx hash in TxCompleted event

### LINKED ISSUE #666: Extension: return transaction hash on send
At this moment namada-extension does not expose Transfer/Transaction status nor transaction hash after request was made.  

It's not possible to build good UI without it, bare minimum would be status (sent successfully or rejected/error) and transaction hash in case of success

### LINKED ISSUE #670: Extension events are not sent when the tab changes
The extension dispatches events at various times, such as when the accounts change or the network configuration changes. The extension stores the tabs of sites connected to the extension, and when it sends an event, it gets sent directly to each tab in the extension storage:

```typescript
// ExtensionBroadcaster.ts
async sendMsgToTabs(msg: Message<unknown>): Promise<void> {
  const tabs = await syncTabs(this.localStorage, this.requester);

  try {
    tabs?.forEach(({ tabId }: TabStore) => {
      // We do not care about the result or the order of sendMessageToTab
      void this.requester.sendMessageToTab(tabId, Ports.WebBrowser, msg);
    });
  } catch (e) {
    console.warn(e);
  }
}
```

If the tab changes, the new tab won't receive events, even if it is a site that has been allowed to connect to the extension.

This should be fixed to allow connected sites to receive events regardless of tab ID.

emccorson:
Closes #666. Fixes #670.

To test the events are dispatched by origin, not tab ID:

1. Connect the interface to the extension.
2. Add an event listener for an extension event in the interface console e.g. `window.addEventListener("namada-network-changed", e => console.log(e))`.
3. Trigger an event in the extension e.g. change the chain ID. Notice the event is logged in the interface.
4. Open the interface in a new tab and repeat steps 2 and 3. Notice the event is still logged.
5. Navigate to a different website in the original tab and repeat steps 2 and 3. Notice the event is **not** logged.
6. Revoke the connection to the interface in the extension. Repeat steps 2 and 3. Notice the event is **not** logged.

Notes:
- I am not sure if I need to write a migration function for the extension storage since I removed the tabs store.
- I wasn't able to test that I'm returning the correct hash in the TxCompleted event since `namada client tx-result` errors with `block indexing is disabled`.

---

### Added
- Include inner tx hash in successful TxCompleted event payload

### Removed
- Remove tab storage and functions to track tabs

### Fixed
- Dispatch extension events based on whether origin is allowed instead of based on tab ID


apps/extension/src/background/approvals/service.ts
```diff
@@ -397,7 +397,6 @@ export class ApprovalsService {
 
     if (allowConnection) {
       try {
-        await this.keyRingService.connect(interfaceTabId);
```
mateuszjasiuk, line 400: I've merged to main more approvals tests so something might fail because of this line removal :)

mateuszjasiuk (approved):
LGTM! Tested and everyting seems to be working fine!

mateuszjasiuk (approved):

jurevans (approved):
nice, this looks good!

PR #673, opened 2024-03-22
chore: update namada to 0.32.0

mateuszjasiuk:
- bumped namada
- removed timeout from balance query
- added eslintrc.js to the shared package


jurevans (approved):
LGTM!

PR #685, opened 2024-03-24
Faucet: Add handling for chains API

### LINKED ISSUE #534: Faucet: Add handling for chains API
There is a new endpoint we can query for testnets: `https://chains.heliax.click/`, which will give us something like the following:

```
{
  "chains": [
    {
      "chain_id": "public-testnet-15.0dacadb8d663",
      "valid_from": "2023-12-18T17:00:00Z",
      "valid_until": "2024-01-20T00:00:00Z",
      "network_type": "Testnet",
      "namada_version": "https://github.com/anoma/namada/releases/tag/v0.28.1"
    }
  ]
}
```
This will allow us to query for start time without first having the chain ID (and we can also construct the faucet API from this chain ID).

givingmittal:
Closes #534

### DIFF
--- a/apps/faucet/src/App/App.tsx
+++ b/apps/faucet/src/App/App.tsx
@@ -22,27 +22,32 @@ import { FaucetForm } from "App/Faucet";
 import { chains } from "@namada/chains";
 import { useUntil } from "@namada/hooks";
 import { Account, AccountType } from "@namada/types";
-import { API } from "utils";
+import { API, ChainsAPI } from "utils";
 import dotsBackground from "../../public/bg-dots.svg";
 import { CallToActionCard } from "./CallToActionCard";
 import { CardsContainer } from "./Card.components";
 import { Faq } from "./Faq";
 
-const DEFAULT_URL = "http://localhost:5000";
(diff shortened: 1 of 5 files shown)


apps/faucet/src/utils/chainsApi.ts
```diff
@@ -0,0 +1,66 @@
+import {
```
mateuszjasiuk, line 1: Can't you just reuse the current api.ts ?

apps/faucet/src/App/App.tsx
```diff
@@ -133,14 +138,37 @@ export const App: React.FC = () => {
...
-    if (nowUTC < startsAtToMilliseconds) {
-      setIsTestnetLive(false);
-    }
 
     // Fetch settings from faucet API
     (async () => {
       try {
+        let startsAtToMilliseconds = startsAt * 1000;
+        let startsAtText = defaults.startsAtText
+        try{
```
mateuszjasiuk, line 147: It would be nice to move this `try...catch` block to the separete function. I think that would make code more readable. Also it should help with getting rid off mutations in multiple places. At the moment it's a bit difficult to follow what is going on.

PR #696, opened 2024-03-28
fix: Extension release version should read from extension package.json

jurevans:
- [x] Bump extension version to match last submitted version
- [x] Extension build asset reads version from `apps/extension/package.json`
- [x] Also fixed issue with `_redirects` being copied into a folder and not picked up by Netlify


mateuszjasiuk (approved):
LGTM!

Issue #706, opened 2024-04-01
Balance isn't automatically update when receiving token


kyrtill6b:
Issue:
- When another account transfer token, the application namada-interface is not update automatically update the balances. Only reload the page manually to reload balance

emccorson:
Just to clarify, the problem is that when an account receives tokens from a transfer, the interface does not update? If that's the case, that is not currently supported by the interface.

I'm not sure whether we will choose to support this feature. We would need to investigate the technical difficulty of having something like that. I will mark this as PoC/maybe for now.

kyrtill6b:
Hi @emccorson, thank you for your quick response. Yes the issue is that when an account receives tokens from a transfer, the interface does not update. In my opinion, the application should have this feature like another wallet (For example: Metamask, Kelpr,...). This feature helps enhance user experience when they use the application.

pnam1609:
In my view, this feature is indispensable for the functionality of this application. 
I looked at [707](https://github.com/anoma/namada-interface/pull/707) and this solution works as described. It's worth noting that it used the Cosmos SDK library to implement this, which will trigger query balancing when a new block change occurs. I wonder if there is any library that supports only catching the event when receiving the token instead of catching all changes of the event block. It makes for better performance. However, [707](https://github.com/anoma/namada-interface/pull/707) still is a good contribution.

pedrorezende:
@pnam1609 thanks for the contribution! Just a quick update: we're going to revisit this PR on phase 3. You can check the suggested phases for Namada [here](https://namada.net/mainnet-launch)

PR #707, opened 2024-04-01
Feat: Adding Feature update balance when having new block

kyrtill6b:
### Issue:
- [Balance isn't automatically update when receiving token](https://github.com/anoma/namada-interface/issues/706)

### Solution
- Using library from Cosmos to config websocket catch event of new block change
- Depend on the event, trigger the function reload balance for transparent account and save it into redux.

### DIFF
--- a/apps/namada-interface/package.json
+++ b/apps/namada-interface/package.json
@@ -7,6 +7,8 @@
   "license": "MIT",
   "private": true,
   "dependencies": {
+    "@cosmjs/socket": "^0.32.3",
+    "@cosmjs/tendermint-rpc": "^0.32.3",
     "@namada/components": "0.2.1",
     "@namada/hooks": "0.2.1",
     "@namada/integrations": "0.2.1",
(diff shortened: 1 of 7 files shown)


kyrtill6b:
Hi @mateuszjasiuk @emccorson, Could you please take a moment to look at my PR. Thanks in advance.

emccorson:
Thank you for your contribution.

I haven't reviewed this yet, but I would say at first glance there are a few problems to overcome before this PR can be merged. Firstly, we need to decide whether the issue reported in #706 is something we want to support. Also, we are currently overhauling the interface, so a lot of code is already in the process of changing, including replacing redux with jotai.

For now, I think this PR can stay as a prototype solution for #706 if we decide to support this.

neopinhuy:
Hmm, I think this is a necessary feature because almost every wallet like Metamask, Kelpr, or Leap has it. The purpose of this is to provide convenience, transparency, and avoid minor errors. It's all about user experience, and I believe it's essential. Thank you for your contribution to it.
@kyrtill6b

PR #713, opened 2024-04-02
Migrating dev server to Vite

pedrorezende:
We've been struggling with the lack of hot module replacement in several moments of the development. I've tried to tweak / change the Webpack config without much success - it kept reloading the whole structure on every change. 

So I took the first step to definitely solve this issue: migrate from Webpack to Vite. I would really appreciate your comments and thoughts about this.

#722


apps/namada-interface/vite.config.js
```diff
@@ -0,0 +1,19 @@
+import react from "@vitejs/plugin-react";
+import { defineConfig } from "vite";
+import tsconfigPaths from "vite-tsconfig-paths";
+
+// https://vitejs.dev/config/
+export default defineConfig({
+  plugins: [react(), tsconfigPaths()],
+  define: {
+    "process.env": {},
```
mateuszjasiuk, line 9: Is this enough to load envs?

apps/namada-interface/package.json
```diff
@@ -48,7 +48,7 @@
     "watch": "webpack-cli --watch",
     "start": "webpack-dev-server",
```
mateuszjasiuk, line 49: You can remove this  I guess :)

mateuszjasiuk (commented):
👋 I think if it improves overall dev experience then we should go for it. It does not seem to work for me though. I'm getting:
`Uncaught SyntaxError: The requested module '/@fs/home/mj/Projects/heliax/namada-interface/packages/shared/src/shared/snippets/shared-801e3fcea73932a3/src/sdk/mod.js?t=1712224447927' does not provide an export named 'fetchAndStoreMaspParams' (at shared.js:2:40)`
When opening interface :(

apps/namada-interface/vite.config.js
pedrorezende: It was a temp fix that I left behind :/ Checking that now

pedrorezende:
> 👋 I think if it improves overall dev experience then we should go for it. It does not seem to work for me though. I'm getting: `Uncaught SyntaxError: The requested module '/@fs/home/mj/Projects/heliax/namada-interface/packages/shared/src/shared/snippets/shared-801e3fcea73932a3/src/sdk/mod.js?t=1712224447927' does not provide an export named 'fetchAndStoreMaspParams' (at shared.js:2:40)` When opening interface :(

Will make a clean install to try to reproduce this error o/

PR #716, opened 2024-04-02
Feat: saving last block query for next fetching

huvny-de:
### Issue
#662

### Solution
Depend on new version of SDK is Namada 0.32.0, to minimize block fetching during shielded syncing, the approach involves passing both the start_query_height and last_query_height parameters fetch function, thus restricting fetching to a specified range. To ensure the correct values for these parameters, the following solution can be implemented:
- Retrieve the start_query_height from local storage.
- Initiate a new function to obtain the latest block, which will determine the last_query_height as the latest block on the network.
- Sync only within the range defined by these two values.
- Upon completion of the sync process, store the value of the latest block in local storage to facilitate subsequent shielded syncing.

### DIFF
--- a/apps/extension/src/background/keyring/keyring.ts
+++ b/apps/extension/src/background/keyring/keyring.ts
@@ -816,12 +816,13 @@ export class KeyRing {
     return Result.ok(null);
   }
 
-  async queryLastBlock() {
+  async queryLastBlock(): Promise<number> {
     try {
       const query = this.sdkService.getSdk().rpc;
       return await query.queryLastBlock();
     } catch (e) {
       console.warn(e);
+      return NaN;
     }
   }
 
--- a/apps/extension/src/background/keyring/service.ts
+++ b/apps/extension/src/background/keyring/service.ts
@@ -538,7 +538,7 @@ export class KeyRingService {
     return await this.sdkService.getSdk().masp.hasMaspParams();
   }
 
-  async queryLastBlock(): Promise<number | undefined> {
+  async queryLastBlock(): Promise<number> {
     return await this._keyRing.queryLastBlock();
   }
 
@@ -550,15 +550,14 @@ export class KeyRingService {
     const startHeight = (await this.localStorage.getLatestSyncBlock())
       ?.lastestSyncBlock;
     const lastHeight = await this.queryLastBlock();
+    const start_height = startHeight ? BigInt(startHeight) : undefined;
+    await this.sdkService.getSdk().rpc.shieldedSync(vks, start_height);
 
     if (startHeight !== undefined && lastHeight) {
       await this.localStorage.setLatestSyncBlock({
         lastestSyncBlock: lastHeight,
       });
     }
-
-    const start_height = startHeight ? BigInt(startHeight) : undefined;
-    await this.sdkService.getSdk().rpc.shieldedSync(vks, start_height);
   }
 
   async queryBalances(
--- a/packages/sdk/src/rpc/rpc.ts
+++ b/packages/sdk/src/rpc/rpc.ts
@@ -216,6 +216,7 @@ export class Rpc {
    * Sync the shielded context
    * @async
    * @param vks - Array of viewing keys
+   * @param start_height - Bigint epoch height to start syncing from
    * @returns
    */
   async shieldedSync(vks: string[], start_height?: bigint): Promise<void> {


Wiemsdorfba:
Tks @huvny-de  for providing this solution. It is indeed necessary to optimize the synchronization process, and this solution effectively addresses the issue by minimizing unnecessary data fetching and storing essential parameters for future synchronization

This solution optimizes shielded syncing, minimizing resource usage while ensuring accurate and efficient data retrieval from the Namada network

Wiemsdorfba:
Tks @huvny-de for providing this solution. It is indeed necessary to optimize the synchronization process, and this solution effectively addresses the issue by minimizing unnecessary data fetching and storing essential parameters for future synchronization

This solution optimizes shielded syncing, minimizing resource usage while ensuring accurate and efficient data retrieval from the Namada network

apps/namada-interface/src/slices/accounts.ts
```diff
@@ -185,6 +185,8 @@ const accountsAtom = (() => {
   );
 })();
 
+const LATEST_SYNCED_HEIGHT_STORAGE = "latest-synced-height"
```
mateuszjasiuk, line 188: I think last block info should be stored in extension local storage. 
https://github.com/anoma/namada-interface/blob/ec3dec8070219f29ccf95e8a50c880da3f032566/apps/extension/src/storage/LocalStorage.ts#L86

packages/integrations/src/Namada.ts
```diff
@@ -90,11 +90,15 @@ export default class Namada implements Integration<Account, Signer> {
...
     };
   }
 
-  public async sync(): Promise<void> {
-    await this._namada?.shieldedSync();
+  public async sync(startHeight?: number, lastHeight?: number): Promise<void> {
+    await this._namada?.shieldedSync({ startHeight, lastHeight });
+  }
+
+  public async queryLastBlock(): Promise<number | undefined> {
```
mateuszjasiuk, line 101: Once you use extension local storage this function is not needed anymore :)

packages/shared/lib/src/query.rs
```diff
@@ -278,12 +291,22 @@ impl Query {
...
+        };
+
         shielded
             .fetch(
                 &self.client,
                 &DefaultLogger::new(&WebIo),
-                None,
-                None,
+                start_query_height,
+                last_query_height,
```
mateuszjasiuk, line 309: I'm pretty sure Namada will query up to last block if you pass None.

apps/namada-interface/src/slices/accounts.ts
```diff
@@ -223,7 +225,13 @@ const balancesAtom = (() => {
         set(base, { ...get(base), [address]: balance });
       });
 
-      await namada.sync();
+      const lastHeight = await namada.queryLastBlock();
+      const startHeightStorage = Number(localStorage.getItem(LATEST_SYNCED_HEIGHT_STORAGE));
+      const startHeight = startHeightStorage ? startHeightStorage : undefined;
+      await namada.sync(startHeight,lastHeight);
+      if(lastHeight){
```
mateuszjasiuk, line 232: Looks like formatting is a bit off. Please run `yarn prepare` in the root of monorepo to setup husky.

packages/shared/lib/src/query.rs
```diff
@@ -278,12 +291,22 @@ impl Query {
 
         let _ = shielded.save().await?;
 
+        let start_query_height = match start_height {
```
mateuszjasiuk, line 294: You can also just `start_height.map(BlockHeight)` - should work :)

mateuszjasiuk (changes requested):
Thanks! I think this is a good start for shielded sync improvements. Left few comments :)

huvny-de:
Hi @mateuszjasiuk , Thanks for your comment. I fixed all of it. Please take a look on that.

mateuszjasiuk (changes requested):
In general it looks good. I've tested it locally and it seems to be working nicely :)
There are few more things to change but they should be easy.
First, linting errors (you can run `yarn lint` in the root of the monorepo to see them):
```bash
@namada/extension
 | 
 | /home/mj/Projects/heliax/namada-interface/apps/extension/src/background/keyring/keyring.ts
 |   819:3  error  Missing return type on function  @typescript-eslint/explicit-function-return-type
 | 
 | ✖ 1 problem (1 error, 0 warnings)
 | 
 | `yarn lint:ci` failed with exit code 1
@namada/sdk
 | 
 | /home/mj/Projects/heliax/namada-interface/packages/sdk/src/rpc/rpc.ts
 |   215:3  warning  Missing JSDoc @param "start_height" declaration  jsdoc/require-param
 | 
 | ✖ 1 problem (0 errors, 1 warning)
 |   0 errors and 1 warning potentially fixable with the `--fix` option.
```
Also, we need to reset `lastestSyncBlock` every time we import new account. The reason is we might import an account that was used for shielding in block 10, but latestSyncBlock is 20 or whatever. So to get accurate context we need to sync from the beginning unfortunately :(

PR #720, opened 2024-04-04
Handle "Error: unreachable" message when importing short private key

stanisloe:
If user attempts to import private key of length less than 64 symbols namada extension continues with the import process but in the end throws `Failed while "Encrypting and storing private key.". Error: unreachable` to the user. While it is unlikelly someone will be importing private key of value `123` it is possible to trim few characters from your private key when you copy it and receive error which is not very informative. ( Which happened to me and thought there is something wrong with extension )

Changes:
Validation for private key minimum length and error message in case wrong length was provided.
Validation before submitting private key to extension background.

Tests:
Built and imported extension to my browser and tested that it is not possible to proceed with short key and meaningfull validation error message is presented to the user.
Also checked the normal flow with proper key.

### DIFF
--- a/apps/extension/src/Setup/ImportAccount/SeedPhraseImport.tsx
+++ b/apps/extension/src/Setup/ImportAccount/SeedPhraseImport.tsx
@@ -47,27 +47,28 @@ export const SeedPhraseImport: React.FC<Props> = ({ onConfirm }) => {
     Array.from(mnemonicsRange)
   );
 
-  const privateKeyError = (() => {
-    const validation = validatePrivateKey(filterPrivateKeyPrefix(privateKey));
+  const validatePkAndFormatErrorMessage = (key: string): string => {
+    const validation = validatePrivateKey(filterPrivateKeyPrefix(key));
     if (validation.ok) {
       return "";
     } else {
       switch (validation.error.t) {
-        case "TooLong":
-          return `Private key must be no more than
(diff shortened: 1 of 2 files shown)


apps/extension/src/Setup/ImportAccount/SeedPhraseImport.tsx
```diff
@@ -47,27 +47,31 @@ export const SeedPhraseImport: React.FC<Props> = ({ onConfirm }) => {
     Array.from(mnemonicsRange)
   );
 
-  const privateKeyError = (() => {
-    const validation = validatePrivateKey(filterPrivateKeyPrefix(privateKey));
+  const validatePkAndFormatErrorMessage = (key: string): string => {
```
mateuszjasiuk, line 50: I think it's fine to call it `validatePk`. We can also return `{valid: boolean, msg: string}` or something, as it seems a bit more explicit :)

mateuszjasiuk (changes requested):
LGTM! Left a comment about function name, also make sure your branch is up to date with main(I did not check that) :) Other than that works nicely!

apps/extension/src/Setup/ImportAccount/SeedPhraseImport.tsx
stanisloe: sure. updated function name and return type 👍

jurevans (commented):
LGTM!

PR #724, opened 2024-04-05
Removing Airdrop app

pedrorezende:
Removes Airdrop project from the list of apps, but not from our hearts <3


mateuszjasiuk (approved):
LGTM! There are some references to airdrop in `yarn.lock` so it's worth running `yarn install` :)

PR #728, opened 2024-04-08
Feat: adding txh display in toast notification

TuanTQ15:
### Link
https://github.com/anoma/namada-interface/issues/729

### Work
- Toast notification display the transaction hash
- When clicking the transaction hash will be redirect to explorer to view detail transaction

### Screenshot:
Before:
image

After:
image


mateuszjasiuk:
Hi @TuanTQ15, thank you for the contribution. Unfortunately we are in the middle of completely changing the UI of the interface so even if we merge this it won't show up in final version. :( In general all the UI changes should be discussed first in the issue so you avoid doing unecessary work. I will close this, sorry for that!

PR #732, opened 2024-04-10
feat/721: Return and display invalid word in mnemonic phrase

### LINKED ISSUE #721: Extension: Mnemonic validation should indicate invalid word
If a user enters an invalid mnemonic word, our validation method should return the index so we can inform the user which word was mistyped.

_NOTE_ We currently use `tiny-bip39`, which has an `ErrorKind::InvalidWord` enum, but this does not return the invalid word index. However, we can use https://github.com/anoma/tiny-bip39/tree/invalid-word-index which has this feature added.

- Update mnemonic validation to return of invalid word
- Update extension UI for importing mnemonics to highlight word so user may resolve the issue easily

jurevans:
Resolves #721 

- [x] `@namada/crypto` should return error containing index of invalid word when validating mnemonic
- [x] Mnemonic phrase import should validate mnemonic and highlight invalid word if found

NOTE: Checked with Chris Holt - we'll be keeping this design for now.

### Screenshot

In addition to returning the error message containing the invalid index, the word input is highlighted for the user to fix:

Screen Shot 2024-04-17 at 11 26 13 AM

### DIFF
--- a/apps/extension/src/Setup/Common/SeedPhraseList.tsx
+++ b/apps/extension/src/Setup/Common/SeedPhraseList.tsx
@@ -7,6 +7,7 @@ type SeedPhraseListProps = {
   words: string[];
   sensitive?: boolean;
   columns?: number;
+  invalidIdx?: number;
   onChange?: (index: number, value: string) => void;
   onPaste?: (index: number, e: React.ClipboardEvent<HTMLInputElement>) => void;
 };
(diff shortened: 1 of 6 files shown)


mateuszjasiuk (approved):
LGTM! :)
I've noticed one thing, when first word is invalid we get different error message
obraz

jurevans:
> LGTM! :) I've noticed one thing, when first word is invalid we get different error message 
@mateuszjasiuk Thanks! I will fix this!

Issue #675, opened 2024-03-22
Extension: Add select fiat currency to settings


jurevans:
- User should be able to select their currency, and have the conversion rate applied to token balance

Figma designs:
https://www.figma.com/file/NFyHbLZXBSl3aUsMxtffvV/Namada-End-User-Interface-V1.0-%2F-Phase-1%3A-Block-Party?type=design&node-id=3968-2&mode=design&t=mWoF2N3EVo2YwmLj-0

emccorson:
Probably needs to be done after #674.

emccorson:
Won't be done now that requirements have changed.

PR #735, opened 2024-04-11
feat: add extension currency select page

emccorson:
This PR adds the currency select page to the extension.

### Expected behaviour:
- User can select a currency
- Selected currency persists after the page is reloaded
- Selecting a currency is not used anywhere to display a balance. This will be hooked up in a later PR.

Partially completes #675.


apps/extension/src/App/Settings/Currency.tsx
```diff
@@ -0,0 +1,82 @@
...
+import { GapPatterns, Stack, TickedRadioList } from "@namada/components";
+import { FiatCurrency } from "@namada/types";
+import { PageHeader } from "App/Common";
+
+import { ExtensionKVStore } from "@namada/storage";
+import { KVPrefix } from "router/types";
+import { LocalStorage } from "storage";
+import browser from "webextension-polyfill";
+
+const localStorage = new LocalStorage(
```
mateuszjasiuk, line 12: Not sure if we use this anywhere else in the extension, but we might make it a hook or sth :)

apps/extension/src/App/Settings/Currency.tsx
```diff
@@ -0,0 +1,82 @@
...
+  })
+);
+
+const FIAT_CURRENCIES = [
+  { code: "USD", symbol: "$" },
+  { code: "EUR", symbol: "€" },
+  { code: "GBP", symbol: "£" },
+  { code: "PLN", symbol: "zł" },
+  { code: "BRL", symbol: "R$" },
+  { code: "PHP", symbol: "₱" },
```
mateuszjasiuk, line 25: PHP best currency 😸

mateuszjasiuk (approved):
LGTM! It would be nice to add `cursor: pointer` on currencies hover :)

jurevans (approved):
Nice! This looks good

emccorson:
This won't go in now that the extension requirements have changed.

PR #738, opened 2024-04-12
Feat: staking rewards

### LINKED ISSUE #679: SDK: Claim staking rewards
- Update SDK code to support staking reward claiming
- ~Update interface to support reward claims~ Moved to #741
- Update SDK js lib package to support this

Figma designs:

https://www.figma.com/file/NFyHbLZXBSl3aUsMxtffvV/Namada-End-User-Interface-V1.0-%2F-Phase-1%3A-Block-Party?type=design&node-id=3968-2&mode=design&t=mWoF2N3EVo2YwmLj-0

emccorson:
Adds support for querying and claiming rewards.

I used `BigNumber` as a return type in the SDK, but I'm not sure whether we would rather have that as `string` to stick to non-library TypeScript types.

Closes #679.

---

### Added
- Add ability to submit claim rewards TX
- Add SDK method to query available rewards

### DIFF
--- a/apps/extension/src/background/keyring/service.ts
+++ b/apps/extension/src/background/keyring/service.ts
@@ -6,6 +6,7 @@ import {
   AccountType,
   Bip44Path,
   BondMsgValue,
+  ClaimRewardsMsgValue,
   DerivedAccount,
   EthBridgeTransferMsgValue,
   IbcTransferMsgValue,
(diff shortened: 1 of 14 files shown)


apps/extension/src/background/keyring/service.ts
```diff
@@ -264,6 +265,44 @@ export class KeyRingService {
...
+        claimRewardsMsgValue,
+        txMsgValue
+      );
+      await this.broadcaster.completeTx(
+        msgId,
+        TxType.ClaimRewards,
+        true,
+        innerTxHash
+      );
+      // TODO: not sure if it is correct to broadcast staking update
```
mateuszjasiuk, line 291: It looks like staking update only updates the list of validators, and AFAIK the rewards are not part of bonded stake so it should not be needed :)

jurevans (approved):
This looks good to me! (after conflicts are resolved)

emccorson:
Superseded by #932.

PR #743, opened 2024-04-15
feat: circle fit component

emccorson:
A utility component for fitting text in a circle.

I'm not super happy with how complicated the code got, but it seems to work fine and responds to different border sizes.

Usage:
```html
<div className="w-[500px] h-[500px] border-8 border-yellow rounded-[100%]">
  <CircleFit>
    Some text to fit in the circle
  </CircleFit>
</div>
```

2024-04-15-160828_703x719_scrot

### DIFF
--- a/packages/components/src/CircleFit.tsx
+++ b/packages/components/src/CircleFit.tsx
@@ -1,0 +1,120 @@
+import { tv } from "tailwind-variants";
+
+const circleFit = tv({
+  slots: {
+    wrapper: "w-full h-full",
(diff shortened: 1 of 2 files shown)


euharrison:
@emccorson is this still relevant? Even if we close it, we can see the code in case it's useful for the future

emccorson:
@euharrison this was mostly an experiment since the designs call for fitting things inside circles. I think it could be useful, but I can close it for now and leave it in the back of my mind if we need it.

Issue #583, opened 2024-01-24
Extension: Refactor injected Namada signer


jurevans:
This issue is to update the `Signer` provider to handle signing functions only:

- Build Tx from interface, so sign functions only accept "built" tx (serialized tx and signing data) and signer address
- Remove all `submit...` functions from signing, as extension will only sign
- Add `sign` functionality in signing client to return serialized signed Tx for broadcast (from interface)
- Validate chainId when signing - the `chainId` in extension `Settings` should match what is specified in the Tx
  - In `sign_tx()` (in `shared`) we can access the `chain_id` of the `Tx` once it's deserialized: `let chain_id = tx.header.chain_id.to_string()` or something like that, and validate against what is stored in the extension. This may mean that the `signing.sign()` functions in our SDK JS package will require a `chainId`, which will be provided by the extension when `keyring.sign()` is invoked. The API exposed by the Extension provider/integration will not need to provide the chain ID as it is a part of the Tx that is already built.

PR #749, opened 2024-04-25
WIP: feat/744: Refactoring provider to accept props

### LINKED ISSUE #744: Extension: Move memo field & transaction fee estimate to Extension
Currently, memo field is handled in the interface, as well as the estimated gas fees. These should both be present on the Approvals view for all transactions. 

- Add memo field to extension's Approval screens for all Tx
- Add estimated gas fee to Approval view
- We should probably remove this from the old interface as well for testing purposes

**NOTE** - Currently, the Tx data is serialized on the interface side, which means when we _add_ the memo field to the Tx in the extension, we'd essentially need to deserialize the Tx and serialize a new version of it with the `memo` field added, and commit it to storage for that particular `msgId`. This is pretty ridiculous, so another issue may be to pass Tx in as props to the extension, where it will have `memo` added on approval, then finally serialized before sending to the SDK. Will discuss.

Figma designs:

https://www.figma.com/file/NFyHbLZXBSl3aUsMxtffvV/Namada-End-User-Interface-V1.0-%2F-Phase-1%3A-Block-Party?type=design&node-id=3968-2&mode=design&t=mWoF2N3EVo2YwmLj-0

jurevans:
Partially resolves #744 

As the extension will take responsibility of memo and fees, I'm moving the serialization of the tx data provided by the interface to just before the tx is submitted. This will allows us to append any info the user specifies in the approval process, as well as enable us to display the `data` of the Tx, as this will also allow querying for all tx details within the extension.

- [x] Provider should pass props into the extension instead of serialized data
- [x] Stored tx should keep all unserialized data of the tx
- [ ] The `QueryPendingTx` message should return the entirety of the stored Tx so that we may display props for `View data </>` (later PR)
- [ ] Before submitting the Tx, serialize so that it may be parsed in Rust

Immediately following this PR, Memo & Fees UI will be added to the extension.

### DIFF
--- a/apps/extension/src/Approvals/Approvals.tsx
+++ b/apps/extension/src/Approvals/Approvals.tsx
@@ -4,9 +4,9 @@ import { Route, Routes } from "react-router-dom";
 import { TxType } from "@heliax/namada-sdk/web";
 import { Container } from "@namada/components";
 
+import { TxMsgProps } from "@namada/types";
 import { AppHeader } from "App/Common/AppHeader";
 import { TopLevelRoute } from "Approvals/types";
-import { PendingTxDetails } from "background/approvals";
 import { ApproveConnection } from "./ApproveConnection";
 import { ApproveSignature } from "./ApproveSignature";
 import { ApproveTx } from "./ApproveTx/ApproveTx";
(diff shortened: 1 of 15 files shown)


jurevans:
Closing as not needed due to work in https://github.com/anoma/namada-interface/pull/754 for issue #583

PR #754, opened 2024-05-01
feat/583: Refactor extension signing provider

### LINKED ISSUE #583: Extension: Refactor injected Namada signer

jurevans:
Resolves #583 

- [x] Remove all `submit...` functionality from extension
- [x] Remove any use of `query` and `RevealPk`
- [x] Add `sign()` method to extension (`shared`, extension providers, etc.)
- [x] Add `sign()` method to SDK JS
- [x] Sign should resolve with signed Tx bytes in `ConfirmTx` and `ConfirmLedgerTx`
- [x] Hook up balances & transfer Tx in interface to test signing
- [x] Debugging why transfers aren't being applied, only fees are deducted
- [x] Re-enable approving multiple Tx, returning all signatures
- [x] `sign()` function should validate Extension's `chainId` setting against `chain_id` in the Tx

Follow-up PRs:
- Update and test for Ledger approvals
- Update and test for MASP

_NOTE_ - as a consequence of this, I removed the `ledger`, `offscreen` and `web-workers...` services, as these will be handled outside of the extension

_NOTE_ - Moved `TxMsgValue` & `TxProps` -> `WrapperTxMsgValue` & `WrapperTxProps` to avoid confusion of wrapper args and tx type-specific props (transfer, bond, etc.)

### Testing

This PR incorporates these changes for `Send` transparent tokens, so we should test:

- Sending tokens _to_ a new account (watch console logs for status, as extension will not dispatch events as it no longer has Submit functionality! As such, I left a few logs to watch the process happening)
- Sending tokens _from_ that new account (to invoke `RevealPK` tx

_NOTE_ Currently (for me) the first time I attempt to send a Tx, the UI crashes :( Not exactly sure why, but on subsequent tries it works. I'll take a look, but am not too concerned about supporting the old interface, this is just a test!

As the extension no longer broadcasts for Submit Tx (success or fail), that will need to be handled in interface. If I have time, I'll dispatch a `fetchBalances` call just so we don't have to refresh, but that will only be to make testing easier


apps/extension/src/background/approvals/handler.test.ts
```diff
@@ -48,26 +44,18 @@ describe("approvals handler", () => {
...
-        {
-          txMsg: "txMsg",
-          specificMsg: "specificMsg",
-        },
-      ],
-      AccountType.Mnemonic
-    );
-    handler(env, approveTxMsg);
-    expect(service.approveTx).toBeCalled();
+    // const approveTxMsg = new ApproveTxMsg(
```
pedrorezende, line 47: Can we remove this comment block?

apps/extension/src/background/chains/handler.ts
pedrorezende, line 1: Do we still need references for the chainId?

apps/extension/src/background/chains/messages.ts
pedrorezende, line 1: Same here, if we don't need these references, can we remove this `chains/message.ts` file?

apps/extension/src/background/keyring/handler.ts
```diff
@@ -21,7 +21,6 @@ import {
   RenameAccountMsg,
   RevealAccountMnemonicMsg,
   SaveAccountSecretMsg,
-  ScanAccountsMsg,
   SetActiveAccountMsg,
   TransferCompletedEvent,
```
pedrorezende, line 25: A few events to be removed here as well

pedrorezende (commented):
Hello! I think there are a few missing things that we still need to remove from the extension. Feel free to merge with them, so we can create another refactoring round. The only thing I would really ask to change is how we handle the form submission

apps/extension/src/test/init.ts
```diff
@@ -63,13 +62,10 @@ export const init = async (): Promise<{
...
   const utilityStore = new KVStoreMock<UtilityStore>(KVPrefix.Utility);
   const localStorage = new LocalStorage(new KVStoreMock(KVPrefix.LocalStorage));
-  const revealedPKStore = new RevealedPKStorage(
-    new KVStoreMock(KVPrefix.RevealedPK)
-  );
   const vaultStorage = new VaultStorage(new KVStoreMock(KVPrefix.IndexedDB));
   const namadaRouterId = await getNamadaRouterId(localStorage);
   const requester = new ExtensionRequester(messenger, namadaRouterId);
-  const txStore = new KVStoreMock<TxStore>(KVPrefix.LocalStorage);
+  const txStore = new KVStoreMock<PendingTx>(KVPrefix.LocalStorage);
```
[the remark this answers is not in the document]
jurevans: It is still necessary as we do not actually pass the transactions to the approval popup, so once it is submitted to the extension, it is stored (in memory only, not persisted should the extension be reloaded) then cleared once signed or rejected. The approval popup now only knows about the `msgId` of the stored Tx (there is a limit to the number of characters we include in the popup query string, so the Tx, which is large, needs to live somewhere else). These `PendingTx` are not persisted in any way beyond the approval process

apps/extension/src/background/chains/handler.ts
jurevans: Yes! In our discussions it was requested that we specify the chain ID in the extension for whichever chain we want to interact with. There is a GH issue to validate the `chainId` of any tx when signing against this stored value, so I left this in for that use case.

apps/extension/src/background/keyring/handler.ts
jurevans: Thanks!

apps/extension/src/background/chains/messages.ts
jurevans: We should keep this - perhaps it could be simplified a bit, but we will need `chainId` to use for validating against Tx chain ID

apps/extension/src/Approvals/ApproveSignArbitrary.tsx
```diff
@@ -17,7 +17,9 @@ type Props = {
   setSignatureDetails: (details: SignatureDetails) => void;
 };
 
-export const ApproveSignature: React.FC<Props> = ({ setSignatureDetails }) => {
+export const ApproveSignArbitrary: React.FC<Props> = ({
+  setSignatureDetails,
```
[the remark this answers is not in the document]
jurevans: Sure!

jurevans:
@pedrorezende Refer to this issue for some follow-up tasks related to this: https://github.com/anoma/namada-interface/issues/770 - I think with this PR we should focus on unblocking the interface so far as integrating with the Extension is concerned, though I will address all of your comments here! Anything that isn't immediately necessary can go into this issue!

apps/extension/src/Approvals/ApproveSignTx/ApproveSignTx.tsx
```diff
@@ -0,0 +1,83 @@
...
+        <div>
+          {signer && (
+            <p className="text-xs">
+              Signer: <strong>{shortenAddress(signer)}</strong>
+            </p>
+          )}
+        </div>
+      </Stack>
+      <Stack gap={3} direction="horizontal">
+        <ActionButton onClick={handleApproveClick}>Approve</ActionButton>
```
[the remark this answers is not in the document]
jurevans: Ok, I updated this! We should still use `onClick` for the `Reject` action button though, correct?

apps/extension/src/Approvals/ApproveSignArbitrary.tsx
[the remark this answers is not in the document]
jurevans: Updated!

apps/extension/src/background/keyring/service.ts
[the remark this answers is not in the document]
jurevans: Yes! Thanks, I should have caught that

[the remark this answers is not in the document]
jurevans: Got it, thanks!

apps/extension/src/background/approvals/handler.test.ts
jurevans: I was planning to re-enable this test as `ApproveTxMsg` has changed

apps/extension/src/background/keyring/handler.ts
jurevans: Got it, thanks!

jurevans:
> LGTM! Maybe we could open an issue to make one more review before next release, just to make sure we're not keeping any other functionality not related to signing.

I opened https://github.com/anoma/namada-interface/issues/771 for this purpose! We can look at that once the Ledger & UI changes are in, and _after_ we integrate with the new interface (in case we catch anything that can be improved)

PR #776, opened 2024-05-21
chore: update namada to 0.36.0

mateuszjasiuk:
+ compiles
- fee_unshield got removed which means unshielding will stop working (we do not need it to work for phase 1 tho)


jurevans (approved):
LGTM!

PR #778, opened 2024-05-24
feat: indexer integration part 1

mateuszjasiuk:


apps/namadillo/src/App/Staking/ValidatorName.tsx
```diff
@@ -18,7 +18,7 @@ export const ValidatorName = ({
       <aside>
         <ValidatorThumb
           imageUrl={validator.imageUrl}
-          alt={validator.alias}
+          alt={validator.alias || validator.address}
```
jurevans, line 21: Yeah, I think this is totally a valid approach in these cases!

apps/namadillo/src/slices/fees.ts
```diff
@@ -12,11 +12,16 @@ export const minimumGasPriceAtom = atomWithQuery<BigNumber>((get) => {
   return {
     queryKey: ["minimum-gas-price", nativeToken],
     queryFn: async () => {
-      const { rpc } = await getSdkInstance();
-      const result = await rpc.queryGasCosts();
-      const nativeTokenCost = result.find(([token]) => token === nativeToken);
+      const api = new DefaultApi();
+      const gasTableResponse = await api.apiV1GasTableGet();
+
+      // TODO: Can nativeToken ever be undefined?
```
jurevans, line 18: Was just thinking about this, and I think we should fall back to a default address: `tnam1qxgfw7myv4dh0qna4hq0xdg6lx77fzl7dcem8h7e` from the point of view of the interface so it should never be undefined. So if it is not set in `.env` or the `queryNativeToken` call fails, I feel like `nativeToken` should still be defined in the app. Just a thought!

apps/namadillo/src/slices/proposals/functions.ts
```diff
@@ -180,276 +172,141 @@ const decodeProposalType = async (
...
   account: Account
 ): Promise<boolean> => {
-  const votes = await fetchProposalVotes(chain, id);
+  const votes = await fetchProposalVotes(id);
   return votes.some((vote) => vote.address === account.address);
 };
 
 export const fetchVotedProposalIds = async (
-  chain: Chain,
+  _chain: Chain,
```
jurevans, line 299: I think it's safe to just remove this and update where applicable, but we can do that later!

apps/namadillo/src/slices/validators.ts
```diff
@@ -34,50 +40,116 @@ export type MyValidator = {
...
+};
+
+export const allValidatorsAtom = atomWithQuery((get) => {
+  const chainParameters = get(chainParametersAtom);
+  return {
+    queryKey: ["all-validators", chainParameters.dataUpdatedAt],
+    enabled: !!chainParameters,
+    queryFn: async () => {
+      const parameters =
+        chainParameters.data?.unbondingPeriodInDays || BigInt(0);
```
pedrorezende, line 73: Should we keep using `BigNumber` here instead?

apps/namadillo/src/slices/validators.ts
```diff
@@ -34,50 +40,116 @@ export type MyValidator = {
...
-      const rpcUrl = get(rpcUrlAtom);
-      const query = new Query(rpcUrl);
-      const myValidatorsRes = await query.query_my_validators([
-        account.data?.address,
+      const unbondingPeriod =
+        chainParameters.data?.unbondingPeriodInDays || BigInt(0);
+      const api = new DefaultApi();
+      const [bondsResponse, totalVotingPowerResponse] = await Promise.all([
+        // TODO: is address always available?
+        api.apiV1PosBondAddressGet(account.data!.address!),
```
pedrorezende, line 112: I think we should move these two queries to their respective atoms, avoiding querying it multiple times

apps/namadillo/src/slices/validators.ts
```diff
@@ -34,50 +40,116 @@ export type MyValidator = {
...
+      account.data?.address,
+      chainParameters.dataUpdatedAt,
+    ],
+    queryFn: async () => {
+      const unbondingPeriod =
+        chainParameters.data?.unbondingPeriodInDays || BigInt(0);
+      const api = new DefaultApi();
+      const [unbondsResponse, totalVotingPowerResponse] = await Promise.all([
+        // TODO: is address always available?
+        api.apiV1PosUnbondAddressGet(account.data!.address!),
```
pedrorezende, line 144: Same here, if separated in two different atoms, we would save this repeated query here

pedrorezende (approved):
LGTM! With a few minor comments that can be addressed later. The only things I would really change before merging is the default port. Port 5000 is the default for a few MacOS services

apps/namadillo/src/App/Sidebars/YourStakingDistribution.tsx
```diff
@@ -15,6 +16,8 @@ export const YourStakingDistribution = ({
     MyValidator | undefined
   >();
 
+  console.log(myValidators);
```
emccorson, line 19: Should this be removed?

apps/namadillo/src/hooks/useValidatorFilter.tsx
```diff
@@ -13,7 +13,9 @@ const filterValidators = (validator: Validator, search: string): boolean => {
   const preparedSearch = search.toLowerCase().trim();
   return (
     validator.address.toLowerCase().indexOf(preparedSearch) > -1 ||
-    validator.alias.toLowerCase().indexOf(preparedSearch) > -1
+    (validator.alias || validator.address)
```
emccorson, line 16: Minor nitpick, but I think this will search the address twice if the alias is undefined, so you could probably just use `validator.alias?` or something like that.

apps/namadillo/src/slices/fees.ts
emccorson: Just to add, the interface currently won't load at all if the native token is not set in `.env`.

emccorson (approved):
Looks good to me! I was having some problems running the indexer locally, but the code looks good :+1:

apps/namadillo/src/slices/fees.ts
mateuszjasiuk: 👍

apps/namadillo/package.json
```diff
@@ -7,6 +7,8 @@
   "license": "MIT",
   "private": true,
   "dependencies": {
+    "@anomaorg/namada-indexer-client": "0.0.8",
```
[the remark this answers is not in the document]
mateuszjasiuk: We can do it later on if we feel this is annoying in any way. Generating the API is way faster, plus we are sure that it's up to date with swagger :)

apps/namadillo/src/slices/proposals/functions.ts
```diff
@@ -180,276 +172,141 @@ const decodeProposalType = async (
...
+    // TODO: do we need to escape the string? I think react does it for us
+    content = proposal.content;
+  }
 
   return {
-    ...deserialized,
-    ...fetchedStatus,
+    id: BigInt(proposal.id),
+    author: proposal.author,
     content,
```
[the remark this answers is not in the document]
mateuszjasiuk: I double checked and content has to be string->string map so we are good :)

apps/namadillo/src/slices/proposals/functions.ts
```diff
@@ -180,276 +172,141 @@ const decodeProposalType = async (
...
-      typeFiltered.filter(
-        (p) =>
-          p.content.title?.toLowerCase().includes(search.toLowerCase()) ||
-          p.id.toString().includes(search)
-      )
-    );
+export const fetchProposalById = async (id: bigint): Promise<Proposal> => {
+  const api = new DefaultApi();
+  const proposalPromise = api.apiV1GovProposalIdGet(Number(id));
+  const totalVotingPowerPromise = api.apiV1PosVotingPowerGet();
```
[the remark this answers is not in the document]
mateuszjasiuk: We need to hide turnout/quorum for completed proposals as proposal author can also change it's status which might change quorum value :/

apps/namadillo/src/lib/query.ts
```diff
@@ -123,12 +128,12 @@ export const signTxArray = async <T>(
     );
 
     if (!signedTxs) {
-      throw new Error("Signing failed");
+      throw new Error("Signing failed: No signed transactions returned");
```
jurevans, line 131: This is fine, just a note though: My Ledger signing PR handles this a bit differently, where we can see the chain ID validation message if there is mismatch, but we can leave this for now!

apps/namadillo/src/slices/fees.ts
jurevans: yeah, I discussed with @pedrorezende a bit about this issue

jurevans (approved):
Changes LGTM!

PR #781, opened 2024-05-28
chore: use local version of web-ext

euharrison:
Use local version of `web-ext` by adding `yarn run` in the script
Ref: https://classic.yarnpkg.com/lang/en/docs/cli/run/

Also, update the readme removing the global installation of `web-ext`


jurevans (approved):
LGTM!

PR #794, opened 2024-05-29
Creating Settings Page

### LINKED ISSUE #787: Namadillo: Settings Page
The settings section in Namadillo should be implemented as a modal, accessible via the cog icon at the top of the page. This modal should overlay the current page being viewed, but have its own distinct URL. Within the modal, users should be able to select the current fiat currency and configure advanced settings, such as the RPC URL. The image below displays how it should look like.

image

pedrorezende:
This PR implements Settings page and implements a modal navigation, keeping the background page active and blurred.

Closes #787 .

### DIFF
--- a/apps/namadillo/src/App/Settings/Advanced.tsx
+++ b/apps/namadillo/src/App/Settings/Advanced.tsx
@@ -1,16 +1,20 @@
 import { ActionButton, Input, Stack } from "@namada/components";
+import SettingsRoute from "App/Settings/routes";
 import { useAtom } from "jotai";
 import { useState } from "react";
+import { useLocation, useNavigate } from "react-router-dom";
 import { rpcUrlAtom } from "slices/settings";
 
 export const Advanced = (): JSX.Element => {
+  const navigate = useNavigate();
+  const location = useLocation();
   const [currentRpc, setCurrentRpc] = useAtom(rpcUrlAtom);
   const [rpc, setRpc] = useState(currentRpc);
 
   const onSubmit = (e: React.FormEvent): void => {
     e.preventDefault();
     setCurrentRpc(rpc);
-    window.history.back();
+    navigate(SettingsRoute.index(), { replace: true, state: location.state });
   };
 
   return (
--- a/apps/namadillo/src/App/Settings/Advanced.tsx
+++ b/apps/namadillo/src/App/Settings/Advanced.tsx
@@ -1,0 +1,35 @@
+import { ActionButton, Input, Stack } from "@namada/components";
+import { useAtom } from "jotai";
+import { useState } from "react";
+import { rpcUrlAtom } from "slices/settings";
+
+export const Advanced = (): JSX.Element => {
+  const [currentRpc, setCurrentRpc] = useAtom(rpcUrlAtom);
+  const [rpc, setRpc] = useState(currentRpc);
+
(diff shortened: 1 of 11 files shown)


apps/namadillo/src/App/Settings/Advanced.tsx
```diff
@@ -0,0 +1,35 @@
...
+  return (
+    <form
+      className="flex flex-col flex-1 justify-between text-base"
+      onSubmit={onSubmit}
+    >
+      <Stack gap={6}>
+        <Input
+          type="text"
+          value={rpc}
+          label="RPC Url"
```
emccorson, line 25: The version of the design I saw just had this label as "RPC". There was also an "LCD" field, but I don't know what that is.

apps/namadillo/src/App/Settings/Advanced.tsx
emccorson, line 1: The design I saw had the Advanced page containing a list that included "RPC Endpoints", and then that page was the one here.

apps/namadillo/src/App/Settings/Advanced.tsx
```diff
@@ -0,0 +1,35 @@
...
+import { rpcUrlAtom } from "slices/settings";
+
+export const Advanced = (): JSX.Element => {
+  const [currentRpc, setCurrentRpc] = useAtom(rpcUrlAtom);
+  const [rpc, setRpc] = useState(currentRpc);
+
+  const onSubmit = (e: React.FormEvent): void => {
+    e.preventDefault();
+    setCurrentRpc(rpc);
+    window.history.back();
```
emccorson, line 13: It probably doesn't matter, but navigating directly to `http://localhost:5173/settings/advanced` and changing the RPC will direct you out of the application.

apps/namadillo/src/App/Settings/CurrencySelectorEntry.tsx
```diff
@@ -0,0 +1,35 @@
...
+  currency,
+  selected,
+  onClick,
+}: CurrencySelectorEntryType): JSX.Element => {
+  return (
+    <li>
+      <button
+        onClick={onClick}
+        className={twMerge(
+          clsx(
```
emccorson, line 21: Does it make a difference to call both `twMerge` and `clsx` here?

emccorson (approved):
Looks good to me! A few comments, but I'm not sure how much is important to address.

- After closing the settings modal, clicking back in browser re-opens the modal
- There's no settings button shown if the extension is not connected (maybe this is one for Chris actually)
- Probably an unimportant edge case, but if you navigate directly to http://localhost:5173/settings in the browser, the background doesn't blur

apps/namadillo/src/App/Settings/Advanced.tsx
pedrorezende: Yeah, I saw that, but I considered it as a placeholder for another field. I've added the "Url" to be more specific

apps/namadillo/src/App/Settings/CurrencySelectorEntry.tsx
pedrorezende: The `twMerge` is necessary to override the `border-transparent` with `border-yellow` in case the selected is true :)

PR #810, opened 2024-05-30
Adding Namada Footer

### LINKED ISSUE #785: Namadillo: Namada footer on every page
We should review the layout and add the Namada footer to every main section that requires it. The image below shows an example of the correct placement of the footer.

image

pedrorezende:
Adding Namada footer and tweaking spacements.

Closes #785


jurevans (approved):
Tested, and changes look good!

PR #815, opened 2024-05-30
feat/770: Signing updates for Ledger HW Wallet

### LINKED ISSUE #770: Extension: Signing updates for Ledger HW wallet
There are a couple areas we can improve and finalize approvals/signing before releasing the new version of the extension:

- Update to latest `ledger-namada` package, test with app from `hub.zondax.ch`, and:
- finish hooking up Ledger HW wallet signing to the new approval process. It has only been made to _build_, but needs finalizing and testing

If we think of any other small tasks, I will simply add them to this issue.

jurevans:
**NOTE** This is updated as of NamadaApp `v0.0.26` and NPM package `v0.0.6`, however, there will be a new release of the these to support the change in Transfers & MASP, though Txs currently support in Namadillo should work!. 

Resolves #770

- [x] Update Ledger-Namada app to `v0.0.6`
- [x] Hook up Ledger approvals for signing - Add ability to submit multiple Tx for signing, and prompt user to sign each on Ledger, then the signed Txs to interface
- [x] TxDetails should show all Txs being signed by Ledger - if more than one, the data should be an array. Non-Ledger details should still be a single Tx object as it is a batch
- [x] Broadcast the signed Ledger txs in interface, and user should see success notifications for each Tx
- [x] UI updates (from @euharrison) merged here from #949 

**TODO**
- [ ] Update to latest Ledger-Namada app when it is released

- Would be nice to make use of `Tx` `wallet_filter` before passing Tx to Ledger for signing - can we invoke this before submitting to Ledger, then pass the original Tx with signatures applied before batching and broadcasting? Maybe a good follow-up issue as it currently works without this, but may encounter memory issues in the future.
- Remaining Ledger-related issue is importing payment address & viewing key for shielded accounts, but we are still awaiting some MASP-related changes on the Ledger app

### TESTING

- Install NamadaApp `v0.0.26` from https://hub.zondax.ch/namada (or whichever the latest version is there) to your Ledger HW Wallet (this won't work in Firefox!)
- Re-run `yarn` to get updated Ledger dependencies
- You will need balance to test signing with Ledger, so I set up a local node based on Namada `v0.41.0`. Transfer funds to your Ledger via CLI (the first Tx will include a RevealPK -  if this is the case, you should see in the TxDetails data that it is an array of Txs with a single commitment each, which is how Ledger must handle these)
- Import your Ledger wallet into the extension, and with it selected, issue requests in the interface - you should be prompted to review on your Ledger, and successfully submit! If RevealPK was included on a single Bond tx, you should see two success notifications (as they were broadcasted separately)
- Issue Tx from a non-Ledger account that hasn't had its PK revealed: You should see in Tx details that it is a single Tx with 2 or more commitments


mateuszjasiuk (approved):
LGTM! Tested all the txs(except redelegate) with ledger and everything works nicely!

PR #816, opened 2024-05-30
Checking for console.log

pedrorezende:
Me and @jurevans were talking about paying a drink (alcoholic or not) for the interface team for every `console.log` left behind in the code.

Because I'm in clearly disadvantage of this, I've opened this PR to check for missing `console.log` during `husky` routine, and also fixed the one I've left behind in `myValidatorsAtom`.

### DIFF
--- a/.lintstagedrc.json
+++ b/.lintstagedrc.json
@@ -1,7 +1,7 @@
 {
-  "*.{js,jsx,ts,tsx,json}": ["prettier --write"],
-  "./apps/**/src/*.{js,jsx,ts,tsx}": ["eslint --fix --max-warnings=0"],
-  "./packages/**/src/*.{js,jsx,ts,tsx}": ["eslint --fix --max-warnings=0"],
-  "./storybook/src/*.{js,jsx,ts,tsx}": ["eslint --fix --max-warnings=0"],
-  "./e2e/src/*.{js,jsx,ts,tsx}": ["eslint --fix --max-warnings=0"]
+  "*.{js,jsx,ts,tsx}": [
+    "prettier --write",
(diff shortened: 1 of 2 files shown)


.lintstagedrc.json
```diff
@@ -1,7 +1,7 @@
 {
-  "*.{js,jsx,ts,tsx,json}": ["prettier --write"],
-  "./apps/**/src/*.{js,jsx,ts,tsx}": ["eslint --fix --max-warnings=0"],
-  "./packages/**/src/*.{js,jsx,ts,tsx}": ["eslint --fix --max-warnings=0"],
-  "./storybook/src/*.{js,jsx,ts,tsx}": ["eslint --fix --max-warnings=0"],
-  "./e2e/src/*.{js,jsx,ts,tsx}": ["eslint --fix --max-warnings=0"]
+  "*.{js,jsx,ts,tsx}": [
+    "prettier --write",
+    "eslint --rule 'no-console: error' --fix --max-warnings=0"
```
emccorson, line 4: Is there a reason to put it in the husky config instead of the eslintrc file? If in the eslintrc, this CI will also be able to check.

pedrorezende: Because we don't want to fire lint warnings for console.log during the development process

jurevans (approved):
LGTM!

.lintstagedrc.json
pedrorezende: That's why I think it should be considered errors just in the lintstaged, because you can bypass it if you want, without affecting the CI or the dev experience. But in my last commit I've added an exception for `console.warn` and `console.error`, so I think we'll be fine :)

PR #832, opened 2024-06-02
Creating a gas usage selection modal

### LINKED ISSUE #788: Namadillo: Gas amount selection modal
This modal should enable users to specify the amount of gas they wish to spend on their transactions, taking the minimum transaction fee into account for validation purposes.

Additionally, we need to establish a method for calculating the range that defines the 'Average' and 'High' gas amount settings.

Please refer to the image below for a visual representation of the expected interface

image

pedrorezende:
This modal aims to provide a way to select the amount of gas a user would like to spend in a single transaction. This first version will only work with NAM tokens.

Closes #788


euharrison:
Replaced by https://github.com/anoma/namada-interface/pull/1585

Issue #789, opened 2024-05-29
Namadillo: refactor notifications and event handling


pedrorezende:
Currently, notifications are being processed within a specific store slice. However, it would be more effective to dispatch events for every significant blockchain interaction, such as the successful completion of a staking transaction. This approach will provide greater flexibility in managing responses, allowing us to dispatch more than just notifications.

Issue #837, opened 2024-06-04
Namadillo: listen to indexer events


pedrorezende:
Instead of polling the RPC endpoint for new balances, Namadillo should be able to subscribe to indexer events in order to listen for wallet / staking balance changes, for example.

[See #123](https://github.com/anoma/namada-indexer/pull/123)

Issue #830, opened 2024-06-02
Need to refresh a few times before balances show up!


ChrisHoltDesign:
If there is a small load time for balances would be good to impliment a loading or syncing animation. Will this work as part of the traffic light system Awa mentioned

PR #838, opened 2024-06-04
Refactor Notifications and adding polling

pedrorezende:
This PR refactors toast notifications, reinserting events into transaction stages. It adds base events for four types of transactions: Bond, Unbond, ReDelegate and Withdraw, each one containing three different states that could be combined with: Pending, Success, and Error (ex: `Bond.Success`, or `Unbond.Error` ). 

This PR also adds a temporary solution for updating the balance after a transaction is dispatched. Now the interface polls the rpc endpoint for a few seconds after a Success event is dispatched. A new issue was opened to refactor this in the future: #837 

Fixes: #789  #830


apps/namadillo/src/slices/accounts.ts
```diff
@@ -2,14 +2,14 @@ import { getIntegration } from "@namada/integrations";
...
   ChainKey,
-  CosmosTokenType,
   TokenBalances,
   TokenType,
 } from "@namada/types";
 import BigNumber from "bignumber.js";
 import { getSdkInstance } from "hooks";
 import { atom } from "jotai";
 import { atomWithQuery } from "jotai-tanstack-query";
+import { shouldUpdateAmountAtom } from "./etc";
```
jurevans, line 12: This is a nitpick, but instead of `Amount` should we use `Balance`? Like `shouldUpdateBalanceAtom`?

jurevans (approved):
LGTM!

PR #841, opened 2024-06-05
More Governance fixes

### LINKED ISSUE #805: Validate proposal ID from URL query properly
The code to read and validate a proposal ID from the URL params should be put in a utility function somewhere to reduce duplication with ViewJson, ProposalAndVote and SubmitVote. It should also validate properly that we successfully parsed a bigint.

### LINKED ISSUE #811: Opening a proposal page should scroll to the top of the page
When you click to open a proposal from the All Proposals table, the page loads at the bottom instead of scrolling back to the top. The extension already does something like this I think so we can probably reuse that.

emccorson:
### Changed
- Rename Governance overview route to "/" to match Staking
- Make Governance route wildcard to match Staking

### Fixed
- Reuse and fix up proposal ID query param logic
- Add scroll to top behaviour when changing page
- Fix back button not working on Governance page

Fixes #805. Fixes #811.

### DIFF
--- a/apps/namadillo/src/App/Governance/Governance.tsx
+++ b/apps/namadillo/src/App/Governance/Governance.tsx
@@ -8,10 +8,7 @@ import GovernanceRoutes from "./routes";
 export const Governance: React.FC = () => (
   <main className="w-full">
     <Routes>
-      <Route
-        path={`${GovernanceRoutes.overview()}`}
-        element={<GovernanceOverview />}
-      />
+      <Route path="/*" element={<GovernanceOverview />} />
       <Route
         path={`${GovernanceRoutes.proposal()}`}
         element={<ProposalAndVote />}
--- a/apps/namadillo/src/App/Governance/Governance.tsx
+++ b/apps/namadillo/src/App/Governance/Governance.tsx
@@ -1,41 +1,26 @@
-import { useSanitizedLocation } from "@namada/hooks";
-import { useEffect } from "react";
-import { Route, Routes, useNavigate } from "react-router-dom";
+import { Route, Routes } from "react-router-dom";
 import { GovernanceOverview } from "./GovernanceOverview";
 import { ProposalAndVote } from "./ProposalAndVote";
 import { SubmitVote } from "./SubmitVote";
 import { ViewJson } from "./ViewJson";
 import GovernanceRoutes from "./routes";
 
-export const Governance: React.FC = () => {
-  const navigate = useNavigate();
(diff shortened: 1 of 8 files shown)


apps/namadillo/src/hooks/useProposalIdParam.ts
```diff
@@ -0,0 +1,33 @@
+import { useEffect } from "react";
+import { useNavigate } from "react-router-dom";
+
+import { useSanitizedParams } from "@namada/hooks";
+import GovernanceRoutes from "App/Governance/routes";
+
+const parseProposalId = (asString: string | undefined): bigint | null => {
```
pedrorezende, line 7: Nitpick: I would say we should keep BigNumber to make everything standard / compatibility, but we're not making any calculations with that, so I think it's okay. Any thoughts?

apps/namadillo/src/App/Governance/Governance.tsx
```diff
@@ -1,41 +1,26 @@
...
-        <Route path={`${GovernanceRoutes.viewJson()}`} element={<ViewJson />} />
-      </Routes>
-    </main>
-  );
-};
+export const Governance: React.FC = () => (
+  <main className="w-full">
+    <Routes>
+      <Route
+        path={`${GovernanceRoutes.overview()}`}
```
pedrorezende, line 12: We can also make this Route with the property `index`

pedrorezende (approved):
LGTM, just a few minor comments! o/

apps/namadillo/src/hooks/useProposalIdParam.ts
emccorson: I had a further think about this. I think `bigint` is good because it more closely matches the values that are valid proposal IDs (basically `u64` in Rust), whereas `BigNumber` includes more values that are not valid (e.g. non-integers). `bigint` is a bit of pain sometimes to use because it can't be used to index objects or as a React key, but `BigNumber` would have the same problems. If we do need to get a `BigNumber` representation we can always do `BigNumber(someBigInt.toString())`. What do you think? I'll leave it in this PR for now and if you still think it should be changed we can add it later.

apps/namadillo/src/App/Governance/Governance.tsx
emccorson: `index` didn't seem to work (I think because of a missing trailing `/`), so I just used `"/*"` to match the Staking routes.

PR #854, opened 2024-06-06
feat/779 - SDK: Incorporate Tx batching functionality

### LINKED ISSUE #779: Extension/Interface: Update to support Tx batching
Now that https://github.com/anoma/namada/pull/3103 is merged into `v0.36.1`, we can make use of batch Tx instead of building, signing, and issuing multiple Tx sequentially (this will save gas and should simplify SDK JS API)

(_NOTE_ Will add more details here as I research how to implement this!)

- Tx should be built as a batch always, and we can consider a single Tx to just be a batch of 1 item (in `packages/shared`)
- SDK JS implements this as a part of the API (in `packages/sdk`)
- Interface is updated to support this change, and the work-around in fee calculation removed as it will not be needed (we do not need to multiple fee for a tx by number of tx)

jurevans:
Resolves #779 

- [x] Introduce support for batch Tx in `shared` & `sdk`
- [x] Incorporate this step into interface Tx
- [x] Sign batch Tx correctly and confirm each Tx is applied to chain
- [x] Update `process_tx` response (and `rpc.broadcastTx` response in SDK) to allow clients to see status of individual Tx within the batch


apps/namadillo/src/lib/query.ts
```diff
@@ -91,22 +91,30 @@ export const buildTxArray = async <T>(
...
+    txArray.push({ type: revealPublicKeyType, tx: revealPkTx.tx });
   }
 
   const encodedTxs = await Promise.all(
     queryProps.map((props) => txFn.apply(tx, [wrapperTxProps, props]))
   );
 
-  const typedEncodedTxs = encodedTxs.map((encodedTx, index) => ({
-    encodedTx,
+  const initialTx = encodedTxs[0].tx as BuiltTx;
```
emccorson, line 101: I think this will compile without casting with `as BuiltTx`, and also in the call to `buildBatch` below.

apps/namadillo/src/lib/query.ts
```diff
@@ -91,22 +91,30 @@ export const buildTxArray = async <T>(
 
   if (!publicKey) {
     const revealPkTx = await tx.buildRevealPk(wrapperTxProps, account.address);
-    txArray.push({ type: revealPublicKeyType, encodedTx: revealPkTx });
+    txArray.push({ type: revealPublicKeyType, tx: revealPkTx.tx });
```
emccorson, line 94: Just a question, but is there a reason for not including the reveal PK TX in the batch?

emccorson (approved):
It works well and I just left a couple of small comments. I'm not super familiar with the TX code though so I would get a second review :smile:

apps/namadillo/src/lib/query.ts
jurevans: Yes! I separated reveal PK so that the batch contains one Tx type, so it has its own approval for now to make it clear that it is a Tx that needs to be approved and sent. Maybe we would want to include it, but for displaying details and such, it's much easier to think about if a batch is of one Tx type

apps/namadillo/src/lib/query.ts
jurevans: I will take a look! I'd really like to get rid of this `BuiltTx` entirely and clean up some of this Tx code. I feel it is unnecessarily complicated

apps/namadillo/src/lib/query.ts
jurevans: I think there is no technical reason that would prevent us from adding it into the batch, but from a UX point of view, the interface always sends a batch of a single tx type at a time. If we decide we want to do it, it will be trivial to add it in. I will ask!

apps/namadillo/src/lib/query.ts
jurevans: ah, yeah, I was messing with having a separate type for batches but removed it, so that was something I forgot to change back :)

apps/extension/src/background/keyring/messages.ts
```diff
@@ -350,3 +351,30 @@ export class DeleteAccountMsg extends Message<
     return DeleteAccountMsg.type();
   }
 }
+
+export class QueryAccountDetailsMsg extends Message<
```
pedrorezende, line 355: Sorry if I'm missing something, but are we using QueryAccountDetailsMsg anywhere? I've noticed this message type hasn't been added to the handler function

pedrorezende (approved):
LGTM

apps/extension/src/background/keyring/messages.ts
jurevans: You're right! This was copied from my Ledger update PR, I can remove it here

jurevans: Fixed!

PR #855, opened 2024-06-07
feat: add pagination to proposals table

### LINKED ISSUE #843: Add pagination to legacy governance proposals list
Screenshot 2024-06-05 at 10 54 30

Can we take the pagination we have currently on the validator list and apply it to the governance proposal archive list.

lets maybe lock it at 50 in the internal scroll before a new page is needed

It might take a few months to fill this out beyond 50 but would be good to have :) 

nb. Not needed for live or upcoming proposals

emccorson:
Also rename ValidatorsTable to TableWithPaginator and refactor to allow use with things other than validators.

Closes #843.

### DIFF
--- a/apps/namadillo/src/App/Common/TableWithPaginator.tsx
+++ b/apps/namadillo/src/App/Common/TableWithPaginator.tsx
@@ -1,0 +1,100 @@
+import { StyledTable, TableHeader, TableRow } from "@namada/components";
+import { FormattedPaginator } from "App/Common/FormattedPaginator";
+import clsx from "clsx";
+import { useCallback, useEffect, useMemo, useRef, useState } from "react";
+
+type TableWithPaginatorProps<T> = {
+  id: string;
+  headers: (TableHeader | React.ReactNode)[];
+  itemList: T[];
+  renderRow: (item: T, index: number) => TableRow;
+  resultsPerPage?: number;
+  initialPage?: number;
+  children?: React.ReactNode;
+} & Pick<React.ComponentProps<typeof StyledTable>, "tableProps" | "headProps">;
+
+export const TableWithPaginator = <T,>({
+  id,
+  headers,
+  renderRow,
+  itemList,
+  resultsPerPage = 100,
+  initialPage = 0,
+  tableProps,
+  headProps,
+  children,
+}: TableWithPaginatorProps<T>): JSX.Element => {
+  const containerRef = useRef<HTMLDivElement>(null);
+  const [page, setPage] = useState(initialPage);
+  const [rows, setRows] = useState<TableRow[]>([]);
+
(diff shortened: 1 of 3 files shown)


apps/namadillo/src/App/Governance/AllProposalsTable.tsx
```diff
@@ -3,14 +3,10 @@ import { useState } from "react";
...
-import {
-  Stack,
-  StyledSelectBox,
-  StyledTable,
-  TableRow,
-} from "@namada/components";
+import { Stack, StyledSelectBox, TableRow } from "@namada/components";
 import { Proposal, isProposalStatus, proposalStatuses } from "@namada/types";
 import { Search } from "App/Common/Search";
+import { TableWithPaginator } from "App/Common/TableWithPaginator";
```
jurevans, line 9: Just a question: Is this now generic enough to live in `@namada/components`? I don't think it's important to move it soon, if at all, just a thought!

apps/namadillo/src/App/Common/TableWithPaginator.tsx
```diff
@@ -0,0 +1,102 @@
...
+  id: string;
+  headers: (TableHeader | React.ReactNode)[];
+  itemList: T[];
+  renderRow: (item: T, index: number) => TableRow;
+  resultsPerPage?: number;
+  initialPage?: number;
+  children?: React.ReactNode;
+} & Pick<React.ComponentProps<typeof StyledTable>, "tableProps" | "headProps">;
+
+export const TableWithPaginator = <T,>({
```
jurevans, line 16: Just a question: Is this now generic enough to live in @namada/components? I don't think it's important to move it soon, if at all, just a thought!

jurevans (approved):
LGTM! Just a couple small comments

apps/namadillo/src/App/Common/TableWithPaginator.tsx
```diff
@@ -0,0 +1,102 @@
...
+        "grid grid-rows-[auto_max-content] flex-1 overflow-hidden w-full gap-2"
+      )}
+    >
+      {styledTable}
+      {pagination}
+      {children}
+    </div>
+  );
+
+  return <></>;
```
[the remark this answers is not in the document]
emccorson: Oops, nice spot! Fixed now.

apps/namadillo/src/App/Common/TableWithPaginator.tsx
emccorson: I was wondering about this too. I'm not 100% sure it's not too interface specific yet and it does depend on the react-paginate dependency, so I left it here for now, but I definitely think this could be considered for a move to components in the future.

PR #856, opened 2024-06-07
Namadillo: Refactoring extension connection

### LINKED ISSUE #852: Namadillo: Review extension detection
When you access the `/staking` page without the extension installed, it fails to detect that the extension is missing. We should also add the loading screen that was previously used, which appears after accessing the page and before the detection routine begins.

pedrorezende:
This PR refactors extension connecting hook, working only with jotai state instead of local state, and also fixes some connection bugs.

Closes #852


jurevans (approved):
Nice, this looks clean!

PR #861, opened 2024-06-07
feat/860 - Update Sdk instance with new settings

### LINKED ISSUE #860: SDK: Provide mechanism to re-initialize wasm in instance with new settings
The current `sdk` package does not provide a direct way of updating itself should new values for `rpcUrl` and `nativeToken` be needed by the client. In its current state, it is required that we create a new instance of Sdk, but this shouldn't necessarily be needed, since we only need to reload the wasms that depend on those values.

- Provide a call on the Sdk instance to update `url` and `nativeToken`, which will reload both the `Sdk` and `Query` wasms
- Make `nativeToken` optional, as some testnets (such as localnets) may not need this value to change

Allowing this instance to be updated will clean up logic in the client, as we can maintain one global instance of Sdk, and simply update to fix all queries/broadcasts.

jurevans:
Resolves #860 

- [x] Sdk instance can be updated with new `url` and `nativeToken` (`Sdk.updateNetwork()`)which will reload the wasms
- [x] Regen docs

### DIFF
--- a/packages/sdk/docs/classes/Crypto.md
+++ b/packages/sdk/docs/classes/Crypto.md
@@ -40,7 +40,7 @@ Class Crypto handles AES encryption tasks
 
 #### Defined in
 
-[sdk/src/crypto/crypto.ts:20](https://github.com/anoma/namada-interface/blob/4a9f889e/packages/sdk/src/crypto/crypto.ts#L20)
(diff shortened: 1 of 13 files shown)


pedrorezende (approved):
Cool! LGTM!

PR #862, opened 2024-06-10
Misc. UI fixes

emccorson:
The main change here is making the proposals table columns have fixed widths so it doesn't move about when the data changes. I think my solution is a bit hacky so I'm not very happy with it, so if there is a better way then please let me know about it.

---

### Fixed
- Make proposals table columns fixed widths
- Fix more instances of wrong black color
- Remove vote button on live proposals list
- Fix lintstaged braces warning


pedrorezende (approved):
LGTM!

PR #871, opened 2024-06-12
Governance pre-release final-ish touches

emccorson:
### Changed
- Temporarily switch off proposal persistent cache
- Temporarily hide placeholder vote info text
- Default to highest vote in proposal summary

### Added
- Add success/failure notifications for voting

### Removed
- Remove ProposalAndVote error handling code (can be done better elsewhere)

### Fixed
- Readd proposals table filter and search
- Use indexer to check if user voted on proposal
- Make the vote button read "Edit Vote" once the user has voted
- Fix some text in notifications
- Add footer to proposal page
- Fix panel gaps on proposal page

### DIFF
--- a/apps/namadillo/src/slices/proposals/functions.ts
+++ b/apps/namadillo/src/slices/proposals/functions.ts
@@ -6,8 +6,6 @@ import {
   ProposalTypeEnum as IndexerProposalTypeEnum,
   VotingPower as IndexerVotingPower,
 } from "@anomaorg/namada-indexer-client";
-import { EncodedTx } from "@heliax/namada-sdk/web";
-import { getIntegration } from "@namada/integrations";
 import {
   Account,
   AddRemove,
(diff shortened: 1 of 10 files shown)


apps/namadillo/src/slices/proposals/functions.ts
```diff
@@ -308,71 +351,33 @@ export const fetchVotedProposalIds = async (
...
+): Promise<TransactionPair<VoteProposalProps>[]> => {
   try {
-    const signedTx = await signingClient.sign(
-      signer,
-      txArray.map(({ tx }) => tx)
+    const { tx } = await getSdkInstance();
+
+    const voteProposalProps = {
+      signer: account.address,
+      proposalId: BigInt(1),
```
pedrorezende, line 366: Not sure if this is intended, because we have `proposalId` param unused

pedrorezende (approved):
Looks really good :) Before merging, could you please just take a look at the comment I made to check if something was left behind?

apps/namadillo/src/slices/proposals/functions.ts
emccorson: Oh nice spot! That is definitely a mistake. I've fixed it and added a commit which will make the linter catch this error.

mateuszjasiuk (approved):
LGTM!

PR #874, opened 2024-06-12
Namadillo: Better error handling

### LINKED ISSUE #839: Namadillo: Error handling components
We should create few components for error handling:
[] Text error messages
[] Box error messages
[] Page error messages

pedrorezende:
Adds new components for error handling, applied a few of them on Staking and Account Overview.
I think we can now close #839 . If we need to apply these components in other locations, maybe we can create separate issues to track it


apps/namadillo/src/App/Common/RouteErrorBoundary.tsx
```diff
@@ -0,0 +1,59 @@
...
+      className={clsx(
+        "flex items-center justify-center flex-1 w-full h-full",
+        "relative text-yellow h-[100svh]"
+      )}
+    >
+      <div className="flex flex-col gap-2 items-center max-w-[40%] text-center">
+        <i className="text-7xl">
+          <MdOutlineFace5 />
+        </i>
+        <strong className="text-2xl">Ops... an error occurred</strong>
```
emccorson, line 36: Ops -> Oops

apps/namadillo/src/App/Common/RouteErrorBoundary.tsx
emccorson, line 1: What do you think about hiding the error message in production? I don't have a strong opinion either way though.

apps/namadillo/src/App/Staking/AllValidatorsTable.tsx
emccorson, line 1: This component still has an early return line which means we don't get to see the new error component:

```
if (validators.isError) return <>Error!</>;
```

It also still dispatches the old error toast.

emccorson (approved):
This is really nice :sunglasses: 

I think there are a couple of reasons we might not get to see the new error components in some cases:

1. There is other logic in the component that returns early on error
2. The queries that dependent on other queries (the ones that use `enabled`) have the status `pending` even when the queries they use have status `error`. That means those queries stay pending forever. I'll have a think about how we might fix that.

But those things could follow in a later PR and the new error handling itself looks good!

jurevans (approved):
LGTM! Nothing to add that wasn't covered in other comments

apps/namadillo/src/App/Common/RouteErrorBoundary.tsx
jurevans: Maybe initially we can keep it for odd situations users find themselves in? Might help to more quickly sort things out

apps/namadillo/src/App/Common/AtomErrorBoundary.tsx
```diff
@@ -0,0 +1,76 @@
+import clsx from "clsx";
+import { AtomWithQueryResult } from "jotai-tanstack-query";
+import { MdErrorOutline } from "react-icons/md";
+import { twMerge } from "tailwind-merge";
+
+type AtomErrorBoundaryProps = {
+  result: AtomWithQueryResult | AtomWithQueryResult[];
+  niceError: string;
+  successIf?: boolean;
```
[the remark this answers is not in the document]
pedrorezende: I think we can remove that, tbh. I don't remember the initial motivation that led me to add that

apps/namadillo/src/App/Staking/AllValidatorsTable.tsx
pedrorezende: Oh good catch!

apps/namadillo/src/App/Common/RouteErrorBoundary.tsx
pedrorezende: I think this is important, mainly in the early phases of Namadillo. We will never use any tracking service like Sentry, so it will be easier for us to identify any unexpected behaviour this way.

apps/namadillo/src/App/Staking/AllValidatorsTable.tsx
pedrorezende: I really like the toast tho :) but will go for the boundary because of the retry button

emccorson (approved):

PR #879, opened 2024-06-14
feat: update indexer client to 0.0.11

mateuszjasiuk:


apps/namadillo/src/slices/fees.ts
```diff
@@ -1,11 +1,75 @@
...
+export type TxKind =
+  | "Bond"
+  | "Unbond"
+  | "Redelegation"
+  | "Withdraw"
+  | "ClaimRewards"
+  | "VoteProposal"
+  | "Unknown";
+
+const txKindFromIndexerTxKind = (
```
pedrorezende, line 18: Can we rename this fn? Maybe `txKindFromIndexer` only?

apps/namadillo/src/App/Staking/IncrementBonding.tsx
```diff
@@ -83,12 +84,13 @@ const IncrementBonding = (): JSX.Element => {
...
     );
     invariant(gasPrice.data, "Gas price loading is still pending");
+    invariant(gasLimits.isSuccess, "Gas limit loading is still pending");
     createBondTransaction({
       changes: parseUpdatedAmounts(),
       account,
       gasConfig: {
         gasPrice: gasPrice.data!,
-        gasLimit: GAS_LIMIT,
+        gasLimit: gasLimits.data!.Bond.native,
```
pedrorezende, line 93: Maybe this approach will not be sufficient for batch transactions :( What do you think?

pedrorezende (approved):
Approving because we need this one in asap 🚀

PR #880, opened 2024-06-17
Small fixes 2024 06 17

emccorson:

### DIFF
--- a/apps/namadillo/src/App/Common/NamInput.tsx
+++ b/apps/namadillo/src/App/Common/NamInput.tsx
@@ -1,0 +1,11 @@
+import { AmountInput } from "@namada/components";
+import { Tokens } from "@namada/types";
+
+type NamInputProps = Omit<
+  React.ComponentProps<typeof AmountInput>,
+  "maxDecimalPlaces"
+>;
+
+export const NamInput: React.FC<NamInputProps> = (props) => (
+  <AmountInput {...props} maxDecimalPlaces={Tokens.NAM.decimals} />
+);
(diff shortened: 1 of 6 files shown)


apps/namadillo/src/App/Settings/SettingsPanel.tsx
```diff
@@ -58,10 +58,15 @@ export const SettingsPanel = (): JSX.Element => {
         </header>
         <Routes>
           <Route index element={<SettingsMain />} />
-          <Route
-            path={`${SettingsRoutes.currencySelection()}`}
-            element={<CurrencySelector />}
-          />
+          {
```
pedrorezende, line 61: I think we don't need a flag here. It's okay to just remove it

apps/namadillo/src/App/Settings/SettingsMain.tsx
```diff
@@ -13,10 +13,15 @@ export const SettingsMain = (): JSX.Element => {
   return (
     <div className="flex flex-1 justify-between flex-col pb-4">
       <ul className="flex flex-col gap-2">
-        <SettingsPanelMenuItem
-          url={`${SettingsRoutes.currencySelection().url}`}
-          text="Currency"
-        />
+        {
+          // TODO: Remove flag when we have fiat currency data
+          process.env.NAMADA_INTERFACE_UNHIDE_CURRENCY && (
```
pedrorezende, line 18: We can just remove this option. It will take a while to add this again

pedrorezende (approved):
I know the PR is still in draft, but I would like to approve the current changes! I think we've worked in the same points for the Settings routes, but I just removed them (and I would go for it, because it will be really easy to add them back + it will take a while to have the fiat currency working with actual data)

PR #882, opened 2024-06-18
temp: remove fiat currency display

### LINKED ISSUE #877: Namadillo: Temporarily hide fiat functionality
We don't have any data yet on the exchange rate for NAM with fiat currencies, so we should probably switch off the currency select in the settings and the "$0" text on the account overview.

emccorson:
Revert this once we have fiat currency data.

Closes #877.


pedrorezende (approved):
LGTM!

PR #883, opened 2024-06-18
feat: show real times, time remaining for proposals

emccorson:
### Added
- Show real dates/times for proposals instead of epochs
- Show the time remaining for ongoing proposals

### DIFF
--- a/apps/namadillo/src/App/Governance/AllProposalsTable.tsx
+++ b/apps/namadillo/src/App/Governance/AllProposalsTable.tsx
@@ -10,7 +10,12 @@ import { useAtomValue } from "jotai";
 import { useEffect, useState } from "react";
 import { GoCheckCircleFill, GoInfo } from "react-icons/go";
 import { useNavigate } from "react-router-dom";
-import { showProposalStatus, showProposalTypeString } from "utils";
+import {
+  proposalStatusToString,
+  proposalTypeStringToString,
+  secondsToDateString,
+  secondsToTimeString,
+} from "utils";
 import { StatusLabel, TypeLabel } from "./ProposalLabels";
 import GovernanceRoutes from "./routes";
 
@@ -171,7 +176,7 @@ export const AllProposalsTable: React.FC<ExtensionConnectedProps> = (props) => {
               id: status,
               value: (
                 <TableSelectOption>
-                  {showProposalStatus(status)}
+                  {proposalStatusToString(status)}
                 </TableSelectOption>
               ),
               ariaLabel: status,
@@ -207,7 +212,7 @@ export const AllProposalsTable: React.FC<ExtensionConnectedProps> = (props) => {
               id: type,
               value: (
                 <TableSelectOption>
-                  {showProposalTypeString(type)}
+                  {proposalTypeStringToString(type)}
                 </TableSelectOption>
               ),
               ariaLabel: type,
@@ -299,6 +304,9 @@ const Status: React.FC<CellProps> = ({ proposal }) => (
   <StatusLabel status={proposal.status} className="ml-auto" />
 );
 
-const VotingEnd: React.FC<CellProps> = ({ proposal }) => (
-  <div className="text-right">Epoch {proposal.endEpoch.toString()}</div>
+const VotingEnd: React.FC<CellProps> = ({ proposal: { endTime } }) => (
+  <div className="text-right">
+    <div>{secondsToTimeString(endTime)}</div>
+    <div className="text-neutral-450">{secondsToDateString(endTime)}</div>
+  </div>
 );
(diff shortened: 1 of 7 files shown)


apps/namadillo/src/App/Governance/ProposalHeader.tsx
```diff
@@ -211,37 +211,63 @@ const StatusLabel: React.FC<{
...
     </div>
   );
 };
 
+const TimeRemaining: React.FC<{
+  proposalQuery: AtomWithQueryResult<Proposal>;
+}> = ({ proposalQuery }) => {
+  const text = (() => {
+    if (
+      proposalQuery.status !== "pending" &&
```
pedrorezende, line 242: I think the only possible condition here is the `proposalQuery.status` being equals to `success`

apps/namadillo/src/App/Governance/ProposalHeader.tsx
```diff
@@ -211,37 +211,63 @@ const StatusLabel: React.FC<{
...
         className="text-right"
       />
     </div>
   );
 };
 
+const TimeRemaining: React.FC<{
+  proposalQuery: AtomWithQueryResult<Proposal>;
+}> = ({ proposalQuery }) => {
+  const text = (() => {
```
pedrorezende, line 240: Maybe this could be a simple function instead of a self-invoking one

apps/namadillo/src/App/Governance/AllProposalsTable.tsx
```diff
@@ -299,6 +304,9 @@ const Status: React.FC<CellProps> = ({ proposal }) => (
   <StatusLabel status={proposal.status} className="ml-auto" />
 );
 
-const VotingEnd: React.FC<CellProps> = ({ proposal }) => (
-  <div className="text-right">Epoch {proposal.endEpoch.toString()}</div>
+const VotingEnd: React.FC<CellProps> = ({ proposal: { endTime } }) => (
+  <div className="text-right">
+    <div>{showTime(endTime)}</div>
+    <div className="text-[#878787]">{showDate(endTime)}</div>
```
pedrorezende, line 310: Can we reuse `text-neutral-400` or `text-neutral-450` here?

pedrorezende (commented):
Hello o/ Left a few comments before approval

apps/namadillo/src/App/Governance/AllProposalsTable.tsx
emccorson: The theme file has `400: "#999999"` and `450: "#838383"` so it looks like we don't have a named colour for `#878787` yet. Do you think I should use 400 or 450 anyway (changing the colour) or define something new? I was getting a bit confused with the designs before because there are a lot of different greys but I wasn't sure if they should all be distinct.

apps/namadillo/src/App/Governance/ProposalHeader.tsx
emccorson: I was wondering too about this pattern. I quite like IIFEs for scoping the logic to just the component in the case where it's more complicated than would work with a ternary statement. I can think of a couple of potential drawbacks:

1. It might be slightly less performant (although not sure about this). Something like `useCallback` would probably be slightly better.
2. If we wanted to test just the function logic, we couldn't get access to it since it's inside the component. Defining it as a normal function outside the component would work here, but the function will only be used by the `TimeRemaining` component so I would prefer to scope it to just the component (defining everything in a new file would also allow this, or having an IIFE wrapping the whole component and function).

What do you think?

apps/namadillo/src/App/Governance/ProposalHeader.tsx
emccorson: Oh nice! That makes more sense.

apps/namadillo/src/App/Governance/AllProposalsTable.tsx
pedrorezende: No worries, sometimes I get confused too. I think the best approach is to add the closest color we have defined in our theme file. This way we avoid adding so many different shades and make the final UI without standard colors

apps/namadillo/src/App/Governance/ProposalHeader.tsx
pedrorezende: I think useCallback would avoid recreating the function many times, but I think it would be better to make such refactor if we decide to optimize this code for some reason (I don't see it as something necessary now, prefer clarity in this case).

We can also scope the function inside the component, but without being a IIFE. Something like `renderText()` or `getText()`.

pedrorezende (approved):
Thanks for the changes, @emccorson ! :) LGTM!

apps/namadillo/src/App/Governance/ProposalHeader.tsx
emccorson: In that case, I think I will leave this as is just now and if there is a refactor in the future we can look at this again (the same pattern is used in a few of the components in this file).

PR #884, opened 2024-06-18
feat: chain id from indexer

mateuszjasiuk:
Also:
- real apr
- real unbonding time
- some fixes to unbonds table

### DIFF
--- a/apps/namadillo/src/App/Staking/UnbondingAmountsTable.tsx
+++ b/apps/namadillo/src/App/Staking/UnbondingAmountsTable.tsx
@@ -26,10 +26,7 @@ export const UnbondingAmountsTable = (): JSX.Element => {
     for (const myValidator of myUnbonds.data) {
       const { validator, unbondedAmount, withdrawableAmount } = myValidator;
 
-      const amount =
-        unbondedAmount?.gt(0) ? unbondedAmount!
-        : withdrawableAmount?.gt(0) ? withdrawableAmount!
-        : new BigNumber(0);
+      const amount = new BigNumber(unbondedAmount || withdrawableAmount || 0);
 
       if (amount.gt(0)) {
         rowsList.push({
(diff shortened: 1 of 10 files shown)


apps/namadillo/src/App/Staking/UnbondingAmountsTable.tsx
```diff
@@ -24,8 +24,14 @@ export const UnbondingAmountsTable = (): JSX.Element => {
 
     const rowsList: TableRow[] = [];
     for (const myValidator of myUnbonds.data) {
-      const { validator, unbondedAmount } = myValidator;
-      if (myValidator.unbondedAmount?.gt(0)) {
+      const { validator, unbondedAmount, withdrawableAmount } = myValidator;
+
+      const amount =
```
mateuszjasiuk, line 29: Not the nicest but the quickest solution for now

apps/namadillo/src/slices/validators.ts
```diff
@@ -216,22 +237,37 @@ const toMyValidators = (
...
       totalVotingPower,
-      unbondingPeriod
+      unbondingPeriod,
+      apr
     );
+    const secondsLeft = BigInt(Number(indexerUnbond.withdrawTime) - timeNow);
+
+    // TODO: later return from the backend
+    const canWithdraw = secondsLeft <= BigInt(0);
+    const timeLeft = timeFromSeconds(canWithdraw ? BigInt(0) : secondsLeft);
```
pedrorezende, line 256: Maybe we should think about adding a date library like `date-fns` to make things easier. Right now I'm getting the following output `0 Days: 9 Hrs: 18 Mins` , I think it should read something like `9h 18 min`

pedrorezende (approved):
Few minor comments that we can open issues in the future to fix, if you want!

jurevans (approved):
Nice, LGTM!

PR #886, opened 2024-06-19
bug: Approve sign promise should resolve when rejected

jurevans:
Fixes an issue where approval promise was not being resolved when the transaction was rejected.

- Updated to resolve promise on reject sign tx
- Moved `rejectSignature` -> `rejectSignArbitrary` to avoid confusion
- Updated tests

### Testing

- Create a transaction, and submit
- In the first view in the popup, click `Reject` -_NOTE_ the interface does not currently display a response, but (in the case of Bond) the `Stake` button will be available again. You can see the error message in the console


emccorson (approved):
Looks good to me!

pedrorezende (approved):
LGTM!

PR #887, opened 2024-06-20
feat: add error messages in toasts

emccorson:
I'm guessing with the styling/layout of these toasts with error messages, so UX comments especially welcome!

---

### Changed
- Stop error toasts from auto-closing on timeout

### Added
- Add error messages in toasts
- Show error toast when signing fails

### Fixed
- Remove "Error:" prefix from extension error messages
- Fix success toasts showing for stake/redelegate errors

### DIFF
--- a/apps/extension/src/provider/InjectedProxy.ts
+++ b/apps/extension/src/provider/InjectedProxy.ts
@@ -60,7 +60,11 @@ export class InjectedProxy {
         }
 
         if (result.error) {
-          reject(new Error(`${result.error}`));
+          const error =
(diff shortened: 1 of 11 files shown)


jurevans (approved):
LGTM! Only thought I have is that maybe we should use word wrapping in the error details. Mismatched chain ID could look like:
```
Signing failed: chain_id luminara-position.5eef
10f5ab82 does not match Tx header chain_id l
uminara-position.5eef10f5ab83
```
which might be a bit confusing and hard to read.

PR #890, opened 2024-06-24
chore: switch to use namada_sdk instead of namada

tzemanovic:
In https://github.com/anoma/namada/pull/3402 we're removing the `namada` crate that contained various unrelated stuff. This PR switches to use `namada_sdk` that will provide the same modules instead (note that this will not compile until https://github.com/anoma/namada/pull/3402 is released, hopefully in v0.40.0, but I did check it against a local path)

### DIFF
--- a/packages/shared/lib/Cargo.lock
+++ b/packages/shared/lib/Cargo.lock
@@ -1927,7 +1927,7 @@ dependencies = [
 [[package]]
 name = "ibc"
 version = "0.53.0"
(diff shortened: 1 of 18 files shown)


tzemanovic:
https://github.com/anoma/namada/pull/3402 didn't make it into the v0.40 release but it should be included in the next one

tzemanovic:
the refactors landed in namada 0.42 - I've update the PR

jurevans (approved):
LGTM! Working great for me

PR #892, opened 2024-06-24
App Setup + Managing default settings

pedrorezende:
This PR contains a few improvements on how Namadillo manages the app settings. These new features are listed below:

- App now displays some useful error messages depending on the type of error experienced by the user
- If Namadillo is being set up without any indexer or prior config, a new indexer settings screen will appear
- Introduces a new config.toml file on /public folder, where you can set up a default config for the installation
- Adds an atom to verify if the indexer url is valid
- Refactors page loader


apps/namadillo/src/hooks/useExtensionEvents.tsx
```diff
@@ -0,0 +1,23 @@
...
+  );
+
+  // Register handlers:
+  useEventListenerOnce(Events.AccountChanged, async () => {
+    await defaultAccount.refetch();
+    balances.refetch();
+  });
+
+  useEventListenerOnce(Events.ConnectionRevoked, () =>
+    setNamadaExtensionConnected("idle")
```
emccorson, line 21: Not introduced in this PR, but `ConnectionRevoked` gets fired when any connection is revoked, so interfaces listening to this event have to check whether they are still connected in the event handler. If I'm remembering correctly, it's this way because we couldn't know which tab the extension should send the event to, so it sends it to all tabs.

apps/namadillo/src/hooks/useSdk.tsx
```diff
@@ -2,24 +2,30 @@ import initSdk from "@heliax/namada-sdk/inline-init";
...
 import { createContext, useContext, useEffect, useState } from "react";
-import { nativeTokenAtom, rpcUrlAtom } from "slices/settings";
+import { nativeTokenAddressAtom } from "slices/chainParameters";
+import { rpcUrlAtom } from "slices/settings";
 
 export const SdkContext = createContext<Sdk | null>(null);
 
 const initializeSdk = async (): Promise<Sdk> => {
   const { cryptoMemory } = await initSdk();
   const store = createStore();
```
emccorson, line 12: Doesn't need to be changed in this PR and I don't know if it's actually causing any problems yet, but we should probably use a global instance of `store` with a provider in `index.tsx`, otherwise atoms will end up reading different values depending on where they're called.

apps/namadillo/src/polyfills.ts
```diff
@@ -0,0 +1,5 @@
+// eslint-disable-next-line @typescript-eslint/ban-ts-comment
+// @ts-ignore: Unreachable code error
+BigInt.prototype.toJSON = function (): string {
```
emccorson, line 3: Just a question, but where is this getting used?

emccorson (commented):
Looks really nice and works well!
I found a couple of bugs which I left in the comments.

apps/namadillo/src/slices/settings.ts
```diff
@@ -53,27 +80,48 @@ export const hideBalancesAtom = atom(
...
+ */
+export const rpcUrlAtom = atom((get) => {
+  const userDefinedRpc = get(namadilloSettingsAtom).rpcUrl;
+  if (userDefinedRpc) return userDefinedRpc;
 
-export const indexerUrlAtom = atom(
-  (get) => get(namadilloSettingsAtom).indexerUrl,
-  changeSettings<string>("indexerUrl")
-);
+  const tomlRpc = get(defaultServerConfigAtom).data?.indexer_url;
```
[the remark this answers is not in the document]
pedrorezende: Oh damn. My bad!

apps/namadillo/src/App/AppSetup.tsx
```diff
@@ -0,0 +1,103 @@
...
+
+  // Checks if the indexer details are valid.
+  if (indexerHeartbeat.isError) {
+    return (
+      <AtomErrorBoundary
+        result={indexerHeartbeat}
+        containerProps={errorContainerProps}
+        niceError={
+          <>
+            Unable to connect to indexer, please check your internet connection.
```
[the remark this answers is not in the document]
pedrorezende: Thanks!

apps/namadillo/src/slices/chain.ts
```diff
@@ -1,5 +1,37 @@
...
+  const indexerRpcUrl = get(indexerRpcUrlAtom);
+  const indexerUrl = get(indexerUrlAtom);
+  const tomlConfig = get(defaultServerConfigAtom);
+
+  return {
+    queryKey: ["chain", indexerUrl, chainParameters, indexerRpcUrl],
+    staleTime: Infinity,
+    ...queryDependentFn(async () => {
+      const rpcUrl = get(rpcUrlAtom);
+      const extensionChainId = (await integrations.namada.getChain())?.chainId;
```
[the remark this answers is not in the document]
pedrorezende: Sorry about that, thank you for catching this one

apps/namadillo/src/hooks/useSdk.tsx
```diff
@@ -12,7 +12,12 @@ const initializeSdk = async (): Promise<Sdk> => {
...
   const rpcUrl = store.get(rpcUrlAtom);
   const nativeToken = store.get(nativeTokenAddressAtom);
-  const sdk = getSdk(cryptoMemory, rpcUrl, "", nativeToken);
+
+  if (!nativeToken.isSuccess) {
```
pedrorezende, line 16: I'm not sure if this is the best approach for sdk initialization, but now the native token is queried in chain parameters atom, not loaded from a .env file. Ideas?

apps/namadillo/src/slices/settings.ts
```diff
@@ -1,35 +1,62 @@
...
+});
+
 export const namadilloSettingsAtom = atomWithStorage<SettingsStorage>(
   "namadillo:settings",
   {
     version: "0.1",
     fiat: "usd",
     hideBalances: false,
-    rpcUrl: process.env.NAMADA_INTERFACE_NAMADA_URL || "",
     indexerUrl: process.env.NAMADA_INTERFACE_INDEXER_URL || "",
```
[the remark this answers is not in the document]
pedrorezende: Hmm I've let this one as another option to initialize the indexerUrl, but I think it makes sense to remove everything related to .env files to avoid confusion :)

apps/namadillo/src/App/AppSetup.tsx
```diff
@@ -0,0 +1,103 @@
...
+    );
+  }
+
+  // Waiting for the extension to be detected along with other chain settings.
+  if (!extensionReady || chain.isPending) {
+    return <PageLoader />;
+  }
+
+  // Handles chain connection errors.
+  if (chain.isError) {
```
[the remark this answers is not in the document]
pedrorezende: Added a handle for that! Thanks! However, couldn't do much about the _http 200 error_ returned by vite xD

apps/namadillo/src/hooks/useExtensionEvents.tsx
pedrorezende: Can we address this one in a new PR?

apps/namadillo/src/hooks/useSdk.tsx
pedrorezende: I've replaced `createStore` by `getDefaultStore` , because we're not setting any store in the Provider

apps/namadillo/src/polyfills.ts
pedrorezende: We've defined a few bigint types in the @anomaorg/namada-indexer-client api. I haven't checked how the react-query handles changes in their dependencies, but probably they're stringified and hashed, and that's why this polyfill is needed

apps/namadillo/src/hooks/useExtensionEvents.tsx
emccorson: Added as #898!

apps/namadillo/src/hooks/useSdk.tsx
emccorson: I guess the problem here is we want the SDK instance to depend on the native token and the RPC URL, which both come from atoms, but `initializeSdk` is just a normal function.

What about putting the global SDK instance inside an atom `sdkAtom` and replacing the calls to `getSdkInstance` with `useAtomValue(sdkAtom)` where possible (which should be possible for most cases)? And for whenever we need to access the SDK outside of React, we can change the definition of `getSdkInstance` to `store.get(sdkAtom)`, and callers of that function just need to be careful to make sure they are not getting an old SDK if the RPC/native token has changed.

If we do that, it should allow us to write the SDK atom to wait on `nativeTokenAddressAtom` being ready, and will also solve any problems of getting an old SDK instance inside our query atoms.

Alternatively, `initializeSdk` could return a Promise that resolves only when the native token is ready.

What do you think?

emccorson (approved):
Changes look good! I especially like the config.toml syntax error solution :sunglasses:

Issue #893, opened 2024-06-25
Namadillo: Decide on a changelog process


emccorson:
We should have a changelog for Namadillo and decide on how we want to do this.

My preference would be to use something like unclog (the tool the main namada team uses), which means if a PR introduces user-facing changes (for Namadillo, this means the host of Namadillo), it should add a new file in a changelog directory. Then at release time, the tool adds all changelog entries to the CHANGELOG.md file. (Alternatively, we could consider just writing a shell script for this to avoid a dependency on unclog and keep the process very simple.)

I think this will lead to better results than automatically generating a changelog based on git history. It is also better than directly writing to CHANGELOG.md, which would cause a lot of merge conflicts.

A good general guide:
https://keepachangelog.com

A more specific (but possibly too restrictive) guide:
https://common-changelog.org/

PR #894, opened 2024-06-25
Namadillo release configuration

emccorson:
This PR adds scripts and documentation for releasing Namadillo.

I have created a temporary release on my fork to show what a real release will look like:
https://github.com/emccorson/namada-interface/actions/runs/9656944071
https://github.com/emccorson/namada-interface/releases/tag/namadillo-1.0.0

There needs to be some follow-up work to define a process for adding changelog entries. For now, I have loosely based this PR on how the main namada team does changelog entries and created #893 for changelog discussion.

---

### Added
- Add scripts to release namadillo
- Add release process documentation
- Add changelog
- Add README for distribution

### Removed
- Remove web vitals from namadillo

### DIFF
--- a/.github/workflows/release-namadillo.yml
+++ b/.github/workflows/release-namadillo.yml
@@ -1,0 +1,44 @@
+name: Create Namadillo draft release
+on:
+  workflow_dispatch:
+    inputs:
+      INCREMENT:
+        required: true
+        type: string
+        default: "patch"
+        description: "Version increment (major/minor/patch etc.)"
+
+jobs:
+  release:
+    runs-on: ubuntu-latest
+    defaults:
+      run:
+        working-directory: ./apps/namadillo
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v3
+
+      - name: Install yarn dependencies
+        uses: ./.github/actions/yarn-cache
+
+      - name: Restore Rust cache
+        uses: ./.github/actions/rust-cache
+        with:
+          cache-name: build
+
+      - name: Install protoc
+        run: sudo apt-get install -y protobuf-compiler
+
+      - name: Build Namadillo
+        run: yarn build
+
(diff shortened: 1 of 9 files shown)


pedrorezende (commented):
@emccorson is it possible to remove any legacy CI routine from the namada-interface in this PR as well?

apps/namadillo/scripts/package.sh
```diff
@@ -0,0 +1,33 @@
+#!/usr/bin/env bash
```
pedrorezende, line 1: Another question, what would be the benefit of using a custom script instead of actions/upload-artifact, like in the wallet.yml ci?

pedrorezende (approved):
LGTM, with a few comments!

.github/workflows/release-namadillo.yml
```diff
@@ -0,0 +1,44 @@
...
+        with:
+          cache-name: build
+
+      - name: Install protoc
+        run: sudo apt-get install -y protobuf-compiler
+
+      - name: Build Namadillo
+        run: yarn build
+
+      - name: Configure git identity
```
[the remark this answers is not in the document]
emccorson: This is needed because release-it creates a commit, which will fail if there is no git identity configured first.

apps/namadillo/scripts/package.sh
emccorson: There are more steps in this script that we couldn't do with actions/upload-artifact as far as I know, such as appending the build information to the README or possibly renaming the README-dist file to README.

mateuszjasiuk (approved):
Nice! Looks good to me :)

emccorson:
> @emccorson is it possible to remove any legacy CI routine from the namada-interface in this PR as well?

@pedrorezende I just checked and I don't see anything left over from the old interface, so I think we should be good :+1:

PR #897, opened 2024-06-26
feat/835 -  Tx details in extension

### LINKED ISSUE #835: Extension: Display Tx details in Approval window for signing
We need to re-introduce details about the Tx being signed in the Approval views.

- Extension will at the very least know what _type_ of Tx is being signed for display purposes
- Add a query to obtain Tx details from the `PendingTx`
- Currently, extension receives serialized values for `tx` & `signing` data as constructed by `build` functions in the SDK. Is it possible to deserialize the Tx completely within the extension from `tx` solely?
- Alternatively, we can pass the original Tx constructing args to this call to make them available to the extension, though if we can avoid that, we probably should, just to keep the API as simple as possible
- Whichever values will be required should be wrapped in the type returned by the SDK JS package's `buildTx` functions, for simplicity's-sake. The TxType can be appended to the route used to invoke the approval, so we can use the `enum` value, e.g.: `/approve-sign-tx/1/....` where `1` references the enum value, or `/approve-sign-tx/transfer/...`

The route could simply follow this structure:

```
/approve-sign-tx/{accountType}/{txType}/{msgId}
```

Where `accountType` determines whether we use Extension key store, or Ledger HW Wallet, and `txType` tells us how to deserialize & display the Tx, and `msgId` gives us the key for the `PendingTx`

See Figma designs for Tx approvals:
https://www.figma.com/design/NFyHbLZXBSl3aUsMxtffvV/Namada-End-User-Interface-V1.0-%2F-Phase-1%3A-Block-Party?node-id=4285-26253&t=k9h95cJpkjsPAh8e-0

## Sign Arbitrary

We should handle "details" the same way when signing arbitrary data as well. We query the details for `signArbitrary` calls, which will always return a string. Then, attempt to parse the string as JSON, and format, something like:

```
try {
  const json = JSON.parse(data);
  const formatted = JSON.stringify(data, null, 2);
  setData(formatted);
} catch (e) {
  setData(data); // unchanged
}
```

And display in a `<pre></pre>` per the designs

Figma:
https://www.figma.com/design/NFyHbLZXBSl3aUsMxtffvV/Namada-End-User-Interface-V1.0-%2F-Phase-1%3A-Block-Party?node-id=3968-2&t=da4IZ3LoUqFISuyV-0#812357271

jurevans:
Resolves #835 

- [x] Add ability to deserialize Tx in `shared` and `sdk` - we have a `to_bytes()` which will return borsh-serialized values that can be deserialized into typed props
- [x] Add query in extension to access pending Tx, and retrieve these details by deserializing Tx bytes
- [x] Display details in `ApproveSignTx`
- [x] Display JSON details when clicking `View Data`
- [x] Add query in extension to access pending `Sign Arbitrary` data, which will return a string that we can attempt to parse as JSON, display in `ApproveSignArbitrary` `View Data` view
- [x] Update SDK docs

### TESTING

__NOTE__ Best to start with a clean install of the extension and new account to test the entire flow. Also, this should probably be tested with the Luminare testnet!

#### Testing Tx details
- Install new copy of extension based on this branch
- Create/Import account (if you want to see RevealPk being added to batch, create an account here, then transfer funds via the faucet at https://faucet.luminara.icu/)
- Once connected in Namadillo, create some Bonds, then submit for approval (if this is the first time this account is used on chain, you will see a RevealPk tx in the list - NOTE that we don't have to display this, I'm just iterating through all Tx in a batch)
- Click `View Data` to see the entire object of the batch Tx, go back & approve
- Ensure that submitting transaction works!

#### Testing Sign-Arbitrary details
- In Namadillo, enable `Sign Arbitrary` in settings, and submit a sign arbitrary request by clicking the icon
- In `View Data`, you should see the string that was passed to the extension for signing
- You can also enter valid JSON into text area, then when you submit it for signing, `View Data` should parse and format this

### TODO

Styles will be addressed in a follow-up PR


apps/namadillo/src/hooks/useWasmHashes.tsx
```diff
@@ -0,0 +1,56 @@
+import { Namada, useIntegration } from "@namada/integrations";
+import { WasmHash } from "@namada/types";
+import { chainAtom, nativeTokenAddressAtom } from "atoms/chain";
+import { rpcUrlAtom } from "atoms/settings";
+import { useAtomValue } from "jotai";
+import { useCallback, useEffect, useState } from "react";
+import { getSdkInstance } from "./useSdk";
+
+export const useWasmHashes = (): void => {
```
pedrorezende, line 9: I think this whole hook could be refactored into an atom. chain, rpcUrl, and  nativeToken could be added as a list of dependencies for this new atom. It would be good to add this in the AppSetup component, so we can show a loading or handle the error right in the initialization

apps/extension/src/Approvals/ApproveSignTx/ApproveSignTx.tsx
```diff
@@ -69,10 +90,61 @@ export const ApproveSignTx: React.FC<Props> = ({ setDetails }) => {
...
+          onClick={() => navigate(TopLevelRoute.ApproveSignTxDetails)}
+        >
+          View Data
+        </ActionButton>
+      </div>
       <Alert type="warning">
         Sign this {accountType === AccountType.Ledger ? "Ledger " : ""}
         transaction?
       </Alert>
+      {details?.txDetails?.commitments &&
```
pedrorezende, line 105: I think we could separate this switch from the body. But we're gonna work in the ux anyways, so no worries about that for now o/

pedrorezende (approved):
TypeScript part looks really really nice! Haven't review the Rust part, maybe @mateuszjasiuk and @emccorson could do it better than me, so consider this as an approval for the TS part. 

Just made two small comments that we can address later. I also saw some problems after trying to unbond, re-delegate and when the gas is too low. We can address everything in different PRs, I think this one is very important to get in.

apps/namadillo/src/hooks/useWasmHashes.tsx
jurevans: Yeah, I was starting to think the same thing. Will take a look!

jurevans:
> Just made two small comments that we can address later. I also saw some problems after trying to unbond, re-delegate and when the gas is too low. We can address everything in different PRs, I think this one is very important to get in.

That's weird about Unbond - that was working before in a previous test, I should be able to fix that quickly as I'm getting the same issue now :/

jurevans:
@pedrorezende can you check Unbond again? I realized that validation was letting me send invalid unbond Tx, where the Unbond amount is more that what is currently staked. I tried submitting multiple Unbonds _up to_ the stake amount, and they are working for me

packages/shared/lib/src/sdk/mod.rs
```diff
@@ -334,20 +339,20 @@ impl Sdk {
         for built_tx in built_txs.into_iter() {
             let tx_bytes = &built_tx.tx_bytes()?;
```
emccorson, line 340: Not in this PR, but is there any difference between calling `tx_bytes` then `Tx::try_from_slice` to get the Tx instead of `built_tx.tx`?

packages/sdk/src/tx/tx.ts
```diff
@@ -449,4 +454,52 @@ export class Tx {
...
+
+  /**
+   * Method to retrieve JSON strings for all commitments of a Tx
+   * @param txBytes - Bytes of a transaction
+   * @param wasmHashes - Array of wasm paths with their associated hash
+   * @returns a TxDetails object
+   */
+  deserialize(
+    txBytes: Uint8Array,
+    wasmHashes: { path: string; hash: string }[]
```
emccorson, line 466: Could maybe use `WasmHash` type here?

packages/shared/lib/src/sdk/args.rs
```diff
@@ -0,0 +1,611 @@
...
+    token: String,
+    fee_amount: String,
+    gas_limit: String,
+    chain_id: String,
+    public_key: Option<String>,
+    memo: Option<String>,
+}
+
+impl WrapperTxMsg {
+    pub fn new(
```
emccorson, line 44: I think you can already create instances of `WrapperTxMsg` like this just by defining the struct and leaving out the `new` method (but it's been a while since I wrote any Rust so I could be wrong about that :smile:).

emccorson (approved):
Looks good to me and works well!

One thing I noticed is that in the details `amount` is in namnam but `feeAmount` is in NAM, so for a transaction of 10 NAM I am seeing `"amount": "10000000"`.

Not related to this PR, but I was thinking if we are refactoring the transaction code, it might be nice to review the type names, because I find it a bit confusing just now with the different TX wrapper types (`EncodedTx`, `EncodedTxData` x 2, `TransactionPair` etc.).

apps/extension/src/Approvals/ApproveSignTx/ApproveSignTx.tsx
jurevans: yeah, I didn't like this either, but didn't want to impose much structure since I knew you would look at it :)

jurevans:
> Looks good to me and works well!
> 
> One thing I noticed is that in the details `amount` is in namnam but `feeAmount` is in NAM, so for a transaction of 10 NAM I am seeing `"amount": "10000000"`.
> 
> Not related to this PR, but I was thinking if we are refactoring the transaction code, it might be nice to review the type names, because I find it a bit confusing just now with the different TX wrapper types (`EncodedTx`, `EncodedTxData` x 2, `TransactionPair` etc.).

You're absolutely right! I didn't want to touch the interface too much, just switched those bits from an array to a single batch Tx - the naming is confusing but that's what was there - I will change these out when I work on the build/signing/broadcast APIs next!

About the fee & amount - I'm not doing any thing to convert the denoms now, but perhaps we should, this is just what's in the Tx, so we should definitely consider whether we want to display as is, or do some conversion. Come to think of it, I think the detailed list of Tx should probably be in NAM (to match what the user entered), and in the JSON view we just show the raw data? That could probably be handled in the UX tasks!

packages/shared/lib/src/sdk/args.rs
jurevans: the problem is is that I moved it into `tx.rs`, so anything outside of that file must use a public accessor/setter. Those fields are all treated as private for these wasm-bindgen structs, and it's annoying, but that's why I added it!

packages/shared/lib/src/sdk/mod.rs
jurevans: `Tx::try_from_slice(built_tx.tx)?` should work, or `borsh::from_slice(built_tx.tx)`. I'm planning to kill this struct in a follow-up PR so I'll leave it for now.  I hate using it this way, and want to pass a simplified object around instead of this struct with methods. It's ugly to use outside of Rust at the moment. I think a better approach would have a simpler type without methods, just a group of args we need. Since I'm using serde / gloo_utils already for the vec of `signing_data` bytes, maybe that would provide a simpler interface for all of the transaction data outside of Rust, just a plain object. That's something I'm looking at next! And hopefully get rid of `wrapper_tx_msg` like you mentioned - if it happens to be needed for broadcast, the interface already has that data so in any case it is not needed here.

packages/sdk/src/tx/tx.ts
jurevans: Yes! Thanks, I missed that!

packages/shared/lib/src/sdk/args.rs
emccorson: Ah, I see! I thought I must be missing something :smile:

PR #899, opened 2024-06-28
Refactoring Store

pedrorezende:
This PR refactors the current store structure, renaming `/slices` to `/atoms`. and unifying the files inside `/store` . It also splits the current atoms files into different files: atoms, services, and functions. 

It also starts a movement of placing types in the root of the project, which I think it might be easier to understand the whole structure of the project.


mateuszjasiuk (approved):
LGTM!

PR #900, opened 2024-06-29
Namadillo: Review Governance Queries

pedrorezende:
This PR fixes some queries in the Governance overview, and also adds an "empty" message when there are no live or upcoming governance proposals.

### DIFF
--- a/apps/namadillo/src/App/Governance/GovernanceOverview.tsx
+++ b/apps/namadillo/src/App/Governance/GovernanceOverview.tsx
@@ -11,6 +11,7 @@ import {
 import { useAtomValue } from "jotai";
 import { AllProposalsTable } from "./AllProposalsTable";
 import { LiveGovernanceProposals } from "./LiveGovernanceProposals";
+import { ProposalListPanel } from "./ProposalListPanel";
 import { ProposalsSummary } from "./ProposalsSummary";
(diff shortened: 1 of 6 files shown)


apps/namadillo/src/App/Governance/GovernanceOverview.tsx
```diff
@@ -22,53 +23,55 @@ export const GovernanceOverview: React.FC = () => {
...
-  );
+  const liveProposals =
+    allProposals.data?.filter((proposal) => proposal.status === "ongoing") ||
+    [];
+
+  const upcomingProposals =
+    allProposals.data?.filter((proposal) => proposal.status === "pending") ||
+    [];
+
+  useNotifyOnAtomError(activeAtoms, [
```
emccorson, line 36: Should we still keep this toast now that the AtomErrorBoundary is in? I am happy either way.

emccorson (approved):
Looks good! Everything works nicely and I didn't find any problems.

apps/namadillo/src/App/Governance/GovernanceOverview.tsx
pedrorezende: I think we can keep this. If we see that is too much, we can remove it at any time. I think it gives more emphasis to the error

PR #902, opened 2024-07-01
Namadillo: disconnecting only from the specified origin

### LINKED ISSUE #898: Namadillo: ConnectionRevoked event handler doesn't check which connection was revoked
`ConnectionRevoked` gets fired when any connection is revoked, so interfaces listening to this event have to check whether they are still connected in the event handler. If I'm remembering correctly, it's this way because we couldn't know which tab the extension should send the event to, so it sends it to all tabs.

See `apps/namadillo/src/hooks/useExtensionEvents.tsx`.

pedrorezende:
Before this PR, Namadillo was disconnected after removing any origin in the extension. Now we pass the removed origin in the revoke event. 

Fixes #898


pedrorezende:
Closing because, as talked internally, handling the revoke connection this way might lead to a security breach

PR #904, opened 2024-07-03
Namadillo: Fees

pedrorezende:
This PR adds back the fees panel with its respective warning

### DIFF
--- a/apps/namadillo/src/App/Common/TransactionFees.tsx
+++ b/apps/namadillo/src/App/Common/TransactionFees.tsx
@@ -1,27 +1,33 @@
 import clsx from "clsx";
-import { useGasEstimate } from "hooks/useGasEstimate";
+import { useAtomValue } from "jotai";
+import { TxKind, gasLimitsAtom, minimumGasPriceAtom } from "slices/fees";
 import { NamCurrency } from "./NamCurrency";
 import { TextLink } from "./TextLink";
+
 type TransactionFeesProps = {
+  txKind: TxKind;
   numberOfTransactions: number;
   className?: string;
 };
 
 export const TransactionFees = ({
+  txKind,
   numberOfTransactions,
   className,
 }: TransactionFeesProps): JSX.Element => {
-  const { calculateMinGasRequired } = useGasEstimate();
-  const minimumGas = calculateMinGasRequired(numberOfTransactions);
+  const gasLimits = useAtomValue(gasLimitsAtom);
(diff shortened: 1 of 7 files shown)


apps/namadillo/src/atoms/fees/services.ts
```diff
@@ -30,7 +30,8 @@ export const fetchMinimumGasPrice = async (
     ({ token }) => token === nativeToken
   );
   invariant(!!nativeTokenCost, "Error querying minimum gas price");
-  const asBigNumber = new BigNumber(nativeTokenCost.amount);
+  // TODO: this should be removed after indexer error is fixed!
```
emccorson, line 33: It looks like this is fixed now.

apps/namadillo/src/App/Common/TransactionFees.tsx
```diff
@@ -1,28 +1,42 @@
...
-        forceBalanceDisplay={true}
-      />
+    <div className={clsx("flex flex-col", className)}>
+      <div className="text-white text-sm">
+        Transaction fee:{" "}
+        <NamCurrency
+          className="font-medium"
+          forceBalanceDisplay={true}
+          amount={gasPrice.data.multipliedBy(
+            gasLimits.data[txKind].native.multipliedBy(numberOfTransactions)
```
emccorson, line 35: Will this still be correct in cases where a reveal PK is needed?

emccorson (commented):
Looks good! I just a left of couple of comments and I would like to quickly check this again after the merge conflicts are fixed.

apps/namadillo/src/App/Common/TransactionFees.tsx
pedrorezende: I think I need to review it again after the batch tx PR gets in

jurevans: With batch Tx in, RevealPk wasn't considered when calculating fees - it is overestimating though so will work as is (though is incorrect). We still need to determine how fees can be calculated when we have a batch with of Txs, and (currently) at most 2 Tx types in that batch (RevealPK + the type of other Txs). I'm up for taking a look, and we don't necessarily need to address that in this PR!

jurevans (approved):
LGTM! Left a comment about fee estimation - that can be handled separately, as it needs a bit of investigation and discussion!

pedrorezende:
Closing it for now

Issue #906, opened 2024-07-03
Namadillo: Sign Arbitrary message feature


pedrorezende:
We need to add a way to sign arbitrary messages using Namadillo. Talking internally, we thought about adding an icon in the top bar navigation, visible when the user enables this feature in the settings

PR #907, opened 2024-07-03
Namadillo: Sign Arbitrary Messages

pedrorezende:
This PR enables arbitrary messages in Namadillo interface. To sign messages, you must enable this feature in the settings panel.

Fixes: #906


emccorson (approved):
All looks good to me! There will be some conflicts with main now but it looks like the changes are basically the same.

PR #908, opened 2024-07-04
fix: fix AmountInput bugs

### LINKED ISSUE #905: AmountInput is buggy
The AmountInput has a few problems:

- It shows the input string after it has been converted to and from a BigNumber, resulting in the user's input changing. I think it should always just show whatever string the user typed.
- The error message stays after the input has been cleared.

emccorson:
- Don't overwrite user's input string
- Clear error message when bad input is cleared
- Throw error on too many zeroes after decimal point
- Make re-delegate clear button clear input instead of setting it to 0

Fixes #905.


mateuszjasiuk (approved):
Tested and it works nicely! :)

pedrorezende:
LGTM! Thanks for fixing these errors, they were really annoying!

pedrorezende (approved):

jurevans (approved):
LGTM!

PR #913, opened 2024-07-08
Small fixes 2024-07-08

### LINKED ISSUE #802: Namadillo: Search component icon overlaps placeholder in Firefox
There was a UI bug for me on Firefox with this component where the placeholder overlapped with the search icon. Chrome seems fine.

You can verify this by opening the interface in Firefox and looking at the search bar in the All Proposals table.

### LINKED ISSUE #911: Namadillo: Missing space character in unable to connect to indexer message
2024-07-08-091445_783x257_scrot

### LINKED ISSUE #912: Namadillo: App is always slightly taller than viewport height
No matter how much you zoom out on Namadillo, the app is always slightly taller than the viewport height meaning there is always a small scroll on the page. I think this might be to do with the side menu height but I have not investigated.

### LINKED ISSUE #914: Namadillo: Side menu opens and closes when switching to small screen view
When switching from the desktop layout (which has the side menu always visible on the left) to the mobile layout (which has the side menu hidden and a burger button to open it), the menu opens and closes very quickly but it shouldn't open at all.

emccorson:
Several small bug fixes for Namadillo.

With the fix about reducing the menu height to stop the scroll, I was only able to reproduce this on Firefox and not Chrome, and I'm not entirely sure why the heights are different. I think we could have a better solution for this by calculating the panel heights based on the viewport height, but it will potentially be tricky to do this nicely, so for now I've just reduced the pixel height.

---

### Fixed
- Add missing space in indexer error message. Fixes #911.
- Fix padding in search input on Firefox. Fixes #802.
- Reduce menu height to stop small scroll. Fixes #912.
- ~Don't open menu when changing to small screen. Fixes #914.~ Dropped.
- Make side menu viewport height on mobile.

### DIFF
--- a/apps/namadillo/src/App/Common/Container.tsx
+++ b/apps/namadillo/src/App/Common/Container.tsx
@@ -46,7 +46,8 @@ export const Container = ({
           className={clsx(
             "transition-transform duration-500 ease-out-expo",
             "pt-10 bg-black rounded-sm fixed top-0 z-[9999] w-[240px]",
-            "h-[calc(100svh-90px)] left-0 xl:z-0 xl:transition-none xl:pt-0 xl:w-auto xl:relative",
+            "h-svh xl:h-[calc(100svh-90px)] left-0 xl:z-0 xl:transition-none",
+            "xl:pt-0 xl:w-auto xl:relative",
             { "-translate-x-full xl:translate-x-0": !displayNavigation }
           )}
         >
--- a/apps/namadillo/src/App/Common/Container.tsx
+++ b/apps/namadillo/src/App/Common/Container.tsx
@@ -46,7 +46,7 @@ export const Container = ({
           className={clsx(
             "transition-transform duration-500 ease-out-expo",
             "pt-10 bg-black rounded-sm fixed top-0 z-[9999] w-[240px]",
-            "h-[calc(100svh-85px)] left-0 xl:z-0 xl:transition-none xl:pt-0 xl:w-auto xl:relative",
+            "h-[calc(100svh-90px)] left-0 xl:z-0 xl:transition-none xl:pt-0 xl:w-auto xl:relative",
             { "-translate-x-full xl:translate-x-0": !displayNavigation }
           )}
         >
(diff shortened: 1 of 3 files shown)


jurevans (approved):
LGTM!

apps/namadillo/src/App/Common/Container.tsx
```diff
@@ -46,8 +46,9 @@ export const Container = ({
           className={clsx(
             "transition-transform duration-500 ease-out-expo",
             "pt-10 bg-black rounded-sm fixed top-0 z-[9999] w-[240px]",
-            "h-[calc(100svh-85px)] left-0 xl:z-0 xl:transition-none xl:pt-0 xl:w-auto xl:relative",
-            { "-translate-x-full xl:translate-x-0": !displayNavigation }
+            "h-[calc(100svh-90px)] left-[-240px] xl:left-0 xl:z-0",
```
pedrorezende, line 49: I see that we have a few errors here: 
- On mobile, the menu height should be the size of the screen, if we're testing in a vertical tablet, for example, we're gonna have some space below the panel
- I think it's better to keep the `-translate-x-full` instead of using `left-[-240px]`. If we wanna tweak the width in the future, we just need to change the `w-[240px]` .

pedrorezende (changes requested):
Hello hello! Just made a few small requests o/

apps/namadillo/src/App/Common/Container.tsx
pedrorezende: Example of the first item: 
image

emccorson: With the second point, the conditional `-translate-x-full` is what is causing the menu to quickly open and close when switching to the mobile view. I've added two new commits, one fixing the height and another to remove the duplication of the width variable. Could you have a look and let me know if you think this solution is OK or if you would prefer a different way?

pedrorezende: Now I see this behaviour! I don't think this is a problem, tbh, but can't see any other solution simpler than the one you've made!

pedrorezende (approved):

apps/namadillo/src/App/Common/Container.tsx
emccorson: Ok, I think that makes sense so I've dropped that commit for now.

PR #915, opened 2024-07-08
fix: mainnet phases link

euharrison:
Fix the Mainnet Phases link using `<a>` instead of `<button>`

Also, update the default value for `as` prop on `<ActionButton>` to observe the presence of the `href` prop

Screenshot 2024-07-08 at 17 19 32


pedrorezende (approved):
LGTM!

PR #919, opened 2024-07-10
feat/891: Adding info to broadcast Tx response

### LINKED ISSUE #891: SDK package: Broadcast Tx should return actual gas used
In https://github.com/anoma/namada-interface/pull/854, I've updated the `process_tx` call to return "applied" status for each Tx commitment in the response. There is a temporary fake gas used, this should be updated to show actual gas used.

jurevans:
Resolves #891 

This adds the remaining details in the `ProcessTxResponse::Applied(resp)`, so the result of `rpc.broadcastTx()` can return this data to the client.

- [x] Invoking `rpc.broadcastTx` should return an object containing details of the `TxResponse`, as well as applied status of each commitment

### Testing

- In `apps/namadillo/src/lib/query.ts`, when invoking `rpc.broadcastTx()`, assign response to a variable and log it (the interface currently doesn't do anything with this data, but we will likely want to, at least to display `gasUsed` & `height`)


pedrorezende (approved):
LGTM!

PR #920, opened 2024-07-10
chore: update references of namada-interface to namadillo

euharrison:
Rename `apps/namada-interface` to `apps/namadillo`


pedrorezende (approved):
@euharrison I'm approving these changes, but this README should be rewritten. We're not using .env file anymore, for example

euharrison:
@pedrorezende instructions updated :)

PR #921, opened 2024-07-11
Refactoring ActionButton

pedrorezende:
This PR removes unused components and fix a very old problem with hairlines appearing on button hover corners. 

Before this PR, the ActionButton component was also using "primary" and "secondary" theme colors, however, these colors are not part of the Tailwind theme. This PR refactors this and also adds more properties to make the ActionButton component even more flexible. 

Related Storybook stories are also updated


packages/components/src/utils.ts
```diff
@@ -0,0 +1,20 @@
...
+
+export const getDefaultColorString = (color: Color): string => {
+  if (!colors.hasOwnProperty(color)) {
+    throw `Unable to find color`;
+  }
+
+  const colorValue = colors[color];
+
+  if (typeof colorValue === "string") {
+    return colors[color] as string;
```
emccorson, line 12: You could return `colorValue` here to avoid casting.

apps/namadillo/src/App/SignMessages/SignMessages.tsx
```diff
@@ -73,7 +72,7 @@ export const SignMessages = (): JSX.Element => {
                 value={data?.signature}
               />
             </Stack>
-            <ActionButton size="lg" borderRadius="sm" onClick={onCloseModal}>
+            <ActionButton borderRadius="sm" onClick={onCloseModal}>
```
emccorson, line 75: It's probably not worth fixing, but I am still getting the hairline at one particular zoom level on Chrome, both laptop screen and monitor. Not on Firefox though strangely.
2024-07-15-134232_2880x2880_scrot

emccorson (approved):
Nice refactor and the buttons are looking good!

packages/components/package.json
```diff
@@ -7,6 +7,7 @@
   "repository": "https://github.com/anoma/namada-interface",
   "author": "Heliax Dev <info@heliax.dev>",
   "license": "MIT",
+  "type": "module",
```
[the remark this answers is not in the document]
pedrorezende: Ops, I think that was the first attempt to fix the error I was receiving. Shouldn't be in the commit. Just removed that.

packages/components/src/ActionButton.tsx
```diff
@@ -1,140 +1,128 @@
...
+  backgroundColor?: Color;
+  backgroundHoverColor?: Color;
+  outlineColor?: Color;
+  textColor?: Color;
+  textHoverColor?: Color;
 } & React.ComponentPropsWithoutRef<HtmlTag> &
-  ActionButtonBaseProps &
-  ActionButtonHoverBaseProps;
+  VariantProps<typeof actionButtonShape> &
+  VariantProps<typeof actionButtonBackground>;
```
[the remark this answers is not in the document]
pedrorezende: Good catch! Fixed

packages/components/src/utils.ts
pedrorezende: Fixed!

apps/namadillo/src/App/SignMessages/SignMessages.tsx
pedrorezende: I think this one might be really complex to fix :/ Maybe we should open an issue for that, but using so much time to fix this hairline that appears only when you zoom in the interface might be a bit too much

pedrorezende:
@emccorson I've made the changes you've pointed out, but since they're only small fixes, I'm going ahead and merging this one. Please let me know if you see something that is not quite right and I can address in a later PR

apps/namadillo/src/App/SignMessages/SignMessages.tsx
emccorson: Yeah, I think that is totally fine to leave :+1:

Issue #922, opened 2024-07-11
Extension: Remove wasm checksums from persisted storage, update client accordingly


jurevans:
In https://github.com/anoma/namada-interface/pull/897, we added the ability to deserialize Tx bytes in the extension. However, since the extension is currently responsible for persisting these once they are fetched, it is possible that a site to provide invalid hashes for a particular chain ID, which would prevent Tx details from displaying.

- Look into a better way of providing these to the extension. It could be added to `connect()`, or simply invoke `addTxWasmHashes(...)` _after_ a successful connection by an interface, but we would remove the `getTxWasmHashes()` call from the extension, meaning the interface would persist these instead
- Update namadillo atom service for checksums to remove the check inside `fetchWasmHashes`, and persist the response from the SDK (soon to be the indexer API)

PR #924, opened 2024-07-11
chore: bump Namada to 0.40.0

jurevans:
This PR bumps the version of `Namada` SDK to `v0.40.0`, and makes the necessary changes

- [x] Bump `@namada/shared` to `v0.40.0` of Namada
- [x] Fix types and `Transfer` args (these have changed!) in `shared`
- [x] Update our SDK package to reflect changes in `Transfer` args, regenerate docs
- [x] Update our schema to properly deserialize `Transfer` type to reflect Namada changes
- [x] Small change to `fetchWasmHashes` to ensure that the check works in the event of an RPC error (interface was adding hashes as an empty array when it shouldn't have been invoked at all, due to a local CORS error)
- [x] Regenerate `packages/types` docs

### Testing

I attempted to test this against Luminare, which is still on `v0.39.0`, and it did not work.

- Set up a local chain based on `v0.40.0` with balances, and point a local indexer to it (I reset the DB as well)
- Set up the extension with an account containing balances
- Issue Tx, and they should succeed!


apps/extension/src/Approvals/ApproveSignTx/ApproveSignTx.tsx
```diff
@@ -134,11 +134,17 @@ export const ApproveSignTx: React.FC<Props> = ({ details, setDetails }) => {
...
-                case TxType.TransparentTransfer:
-                  const transfer = tx as TransparentTransferProps;
+                case TxType.Transfer:
+                  const transfer = tx as TransferProps;
                   return (
                     <div key={i}>
-                      Transparent Transfer: {transfer.target} {transfer.amount}
+                      Transparent Transfer:
+                      <br />
+                      {transfer.data.map(({ source, target, amount }, j) => (
```
jurevans, line 143: This is just a temporary measure to reflect what is in the data! We shouldn't use this UI

apps/namadillo/src/atoms/checksums/services.ts
```diff
@@ -7,11 +7,12 @@ export const fetchWasmHashes = async (
   integration: Namada
 ): Promise<WasmHash[]> => {
   const wasmHashes = await integration.getTxWasmHashes(chainId);
```
jurevans, line 9: Note: In another PR I'm removing these functions from the integration - this atom should only query for these when there is a new chain ID, and store locally, then provide as an (optional) argument to the extension when signing. As there is potential for a client to provide a bad set of wasm hashes, we shouldn't track it in the extension as it may break the Tx details! See https://github.com/anoma/namada-interface/issues/922

packages/types/src/tx/schema/transfer.ts
```diff
@@ -17,7 +17,16 @@ export class TransparentTransferMsgValue {
...
 
-  constructor(data: TransparentTransferProps) {
+  constructor(data: TransferDataProps) {
+    Object.assign(this, data);
+  }
+}
+
+export class TransferMsgValue {
+  @field({ type: vec(TransferDataMsgValue) })
+  data!: TransferDataMsgValue[];
```
emccorson, line 27: Another question :smile: What does it do to pass an array of multiple values here, and how is that different from a batch tx?

packages/types/src/tx/schema/transfer.ts
```diff
@@ -17,7 +17,16 @@ export class TransparentTransferMsgValue {
...
+
+  constructor(data: TransferProps) {
     Object.assign(this, data);
```
emccorson, line 30: The serialization of this is not working for me when I call `buildTransparentTransfer`. It looks like the problem is the data needs to be mapped with `new TransferDataMsgValue` to add the borsh methods to the plain data objects.

packages/shared/lib/src/sdk/transaction.rs
```diff
@@ -127,23 +128,29 @@ impl TransactionKind {
...
+                let Transfer {
+                    sources,
+                    targets,
+                    shielded_section_hash: _,
+                } = transfer;
+
+                let mut data: Vec<TransferDataMsg> = vec![];
+
+                for (i, (source, amount)) in sources.into_iter().enumerate() {
+                    let (target, token) = targets.iter().nth(i).unwrap();
```
emccorson, line 142: It looks like `token` here is the amount, which is putting the amount in the token field when viewing the data.

emccorson (approved):
Looks good to me and works nicely once the serialization is fixed!

packages/types/src/tx/schema/transfer.ts
jurevans: the `Transfer` Tx now contains an array of props, like `{ source, target, amount, token }`, so technically you could have a batch of transfers that each have multiple transfers within them. I'm just going according to the changes to the Tx, I have no idea why this would be needed to be honest!

packages/types/src/tx/schema/transfer.ts
jurevans: Interesting, I'll take a look! I did not test the whole process, just tried to get types updated so it will build, but should be easy enough to fix!

packages/shared/lib/src/sdk/transaction.rs
jurevans: I think you're right, I must have messed it up at some point :(

packages/shared/lib/src/sdk/args.rs
```diff
@@ -335,21 +333,16 @@ pub fn vote_proposal_tx_args(
 
 #[derive(BorshSerialize, BorshDeserialize)]
 #[borsh(crate = "namada::core::borsh")]
-pub struct TransparentTransferMsg {
+pub struct TransferDataMsg {
```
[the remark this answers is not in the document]
jurevans: I'll take a look - I tried to remove `Transparent` where it could refer to another type of Transfer since it's more generic now, but when it comes to args you're right - the args will be different shielded. I think I got a little overzealous when renaming things :D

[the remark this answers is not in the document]
jurevans: I think for the file name in `types` I may keep everything in `types/src/tx/schema/transfer.ts` - we can just define multiple schemas there, including the first to be defined which should be `TransparentTransferMsgValue`

packages/shared/lib/src/sdk/transaction.rs
```diff
@@ -32,9 +34,8 @@ pub enum TransactionKind {
 impl TransactionKind {
     pub fn from(tx_type: TxType, data: &[u8]) -> Self {
         match tx_type {
-            TxType::TransparentTransfer => TransactionKind::TransparentTransfer(
-                TransparentTransfer::try_from_slice(data)
-                    .expect("Cannot deserialize TransparentTransfer"),
+            TxType::Transfer => TransactionKind::Transfer(
+                Transfer::try_from_slice(data).expect("Cannot deserialize TransparentTransfer"),
```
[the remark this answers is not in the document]
jurevans: So this block of code in `shared/lib/src/sdk/tx.rs` gets the "code" section from the Tx and parses the bytes into a hash - this hash refers to the hash of the wasm used in the Tx that determines its behavior:

https://github.com/anoma/namada-interface/blob/a3754f477e7f9230d186d5e6259d9112ddd1c45a/packages/shared/lib/src/sdk/tx.rs#L170-L176 

We match these hashes to their wasm path via a query, since this can only be known after build, and can be queried from the chain storage (these checksum hashes will be different when the wasms change - they get their checksums at the end of `make build-wasm-scripts`):

https://github.com/anoma/namada-interface/blob/a3754f477e7f9230d186d5e6259d9112ddd1c45a/packages/shared/lib/src/query.rs#L776-L789 

Code path is need to issue this query, where we get a storage key from it then make a call over RPC: 

```rust
let hash_key = Key::wasm_hash(tx_code_path);
let (tx_code_res, _) = rpc::query_storage_value_bytes(&self.client, &hash_key, None, false)....
```

Here I just map the wasm "code path" to our `TxType`: https://github.com/anoma/namada-interface/blob/a3754f477e7f9230d186d5e6259d9112ddd1c45a/packages/shared/lib/src/sdk/tx.rs#L97-L123

Not sure of any documentation you could read up on Tx, but I can try to explain here: The core structure of a Tx is simple, it's just a `header` and `sections`, and the sections contain different types of data, signatures (`Authorization`), code (transaction wasm code), inner tx data, etc.:

Tx: https://github.com/anoma/namada/blob/22a4839d84b8e7ff6a909b94d50165d98434ee15/crates/tx/src/types.rs#L1032-L1049

Sections enum: https://github.com/anoma/namada/blob/22a4839d84b8e7ff6a909b94d50165d98434ee15/crates/tx/src/types.rs#L768-L788

`Header` contains things like `chain_id`, `batch`, `expiration`, and `tx_type`. `tx_type` here is different than ours, it is an enum that refers to `Raw`, `Wrapper` and `Protocol`

That's about the extent of what I understand! Still learning myself.

packages/shared/lib/src/sdk/args.rs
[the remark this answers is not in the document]
jurevans: Updated - and I think our `TxType` enum should still indicate `TransparentTransfer` as that determines what args are acceptable, so I'm changing that back as well!

packages/types/src/tx/schema/transfer.ts
jurevans: Fixed!

packages/shared/lib/src/sdk/args.rs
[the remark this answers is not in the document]
jurevans: Done!

packages/shared/lib/src/sdk/transaction.rs
jurevans: Updated, thanks!

emccorson: I am still seeing this error with the latest changes. I think it's that `sources` is actually a map of account →
 amount, not account → token. There is a `token` field on account that seems to work though, although I'm not sure if it should be the source account or the target account.

PR #925, opened 2024-07-11
feat: add url validation

euharrison:
Add the URL validation for Indexer URL and RPC URL, also checking the `/health` endpoint to confirm they are valid endpoints

It's important to notice that we are now validating the RPC using the `/health` endpoint as well, in case we have other services that doesn't support this endpoint

Screenshot 2024-07-11 at 16 30 18


apps/namadillo/src/App/AppSetup.tsx
```diff
@@ -57,15 +57,19 @@ export const AppSetup = ({ children }: AppSetupProps): JSX.Element => {
...
+            </p>
+            <p className="mt-1">
+              If the problem persists, you can{" "}
+              <button
+                className="text-yellow hover:text-cyan"
+                onClick={() => setChangeIndexerSettings(true)}
+              >
+                update the indexer settings.
+              </button>
+            </p>
```
euharrison, line 72: Update the copy/style of the button

Screenshot 2024-07-11 at 16 27 17

apps/namadillo/src/hooks/useValidateApiUrl.ts
```diff
@@ -0,0 +1,25 @@
+import { isUrlValid } from "@namada/utils";
+import { isApiAlive } from "atoms/settings";
+import { useMemo } from "react";
+
+export const useValidateApiUrl = (): ((url: string) => Promise<string>) => {
+  return useMemo(() => {
+    return async (url) => {
+      const trimmedUrl = url.trim();
```
pedrorezende, line 8: Because we're now reusing this, maybe it would be good to separate the concerns and move the sanitize url lines into a new function. Also, maybe `useValidateApiUrl` should return a boolean or another type rather than a string (it's a bit unexpected that this hooks returns the sanitized string, for example)

apps/namadillo/src/hooks/useValidateApiUrl.ts
```diff
@@ -0,0 +1,25 @@
+import { isUrlValid } from "@namada/utils";
+import { isApiAlive } from "atoms/settings";
+import { useMemo } from "react";
+
+export const useValidateApiUrl = (): ((url: string) => Promise<string>) => {
+  return useMemo(() => {
```
pedrorezende, line 6: Should we use useCallback here?

apps/namadillo/src/atoms/settings/services.ts
```diff
@@ -2,7 +2,7 @@ import { DefaultApi } from "@anomaorg/namada-indexer-client";
 import toml from "toml";
 import { SettingsTomlOptions } from "types";
 
-export const isIndexerAlive = async (url: string): Promise<boolean> => {
+export const isApiAlive = async (url: string): Promise<boolean> => {
```
pedrorezende, line 5: I understand what you've done here :) and I think it's a very clean idea, however I'm concern that it might not be future-proof. The RPC and the indexer have two completely different APIs and purposes, although the /health endpoint coincides. However, DefaultApi it's an external library that only queries things in the indexer, and it's susceptible to changes. So I think, for this specific case, we should add some replication to ensure everything will be always fine

apps/namadillo/src/App/Settings/Advanced.tsx
```diff
@@ -1,23 +1,45 @@
...
+  const [indexerError, setIndexerError] = useState("");
+  const [validatingUrl, setValidatingUrl] = useState(false);
 
-  const onSubmit = (e: React.FormEvent): void => {
+  const onSubmit = async (e: React.FormEvent): Promise<void> => {
     e.preventDefault();
-    setCurrentIndexer(indexer);
-    setCurrentRpc(rpc);
-    navigate(SettingsRoute.index(), { replace: true, state: location.state });
+    setValidatingUrl(true);
```
pedrorezende, line 23: I wonder if we should create query atoms to validate the indexer / rpc urls 🤔 so we don't need to recreate all these flags. wdyt?

pedrorezende (changes requested):
Thank you, @euharrison o/ I've opened some questions and made some comments for us to discuss before approving this one

apps/namadillo/src/hooks/useValidateApiUrl.ts
euharrison: yes! sorry, I was moving code around and forgot this

apps/namadillo/src/hooks/useValidateApiUrl.ts
euharrison: I understood your concern, but we need to return the string because we trim and remove the last slash.
Maybe this is a naming issue. I'll propose something different soon

apps/namadillo/src/atoms/settings/services.ts
euharrison: Yeah, makes sense, I'll improve it 👍

apps/namadillo/src/App/Settings/Advanced.tsx
euharrison: do you mean like the current implementation of `indexerHeartbeatAtom`?

apps/namadillo/src/hooks/useValidateApiUrl.ts
euharrison: I moved the logic to the write function of the atom and created the helper `changeSettingsUrl` to handle both cases. With this change, we don't have the hook for validation anymore. What do you think?

apps/namadillo/src/atoms/settings/services.ts
euharrison: We need to make a decision here as don't have a query library on namadillo yet. I can install `axios` to resolve this, what to do you think?

Another alternative is to create a new instance of the sdk and add a method there to check the health endpoint, but I believe it's a bazooka for a simple task

euharrison: done!

apps/namadillo/src/hooks/useValidateApiUrl.ts
euharrison: moved this logic to `sanitizeUrl`

apps/namadillo/src/App/Settings/Advanced.tsx
euharrison: @pedrorezende I moved everything to the mutation solution. Let me know your impressions :)

pedrorezende (approved):
LGTM! Approving with some comments :)

PR #927, opened 2024-07-12
Namadillo: Creating Docker setup

### LINKED ISSUE #926: Namadillo: Create Dockerfile
We should create a Dockerfile to run Namadillo production build

pedrorezende:
This PR introduces the first version of a Dockerfile for Namadillo

Closes #926


apps/namadillo/package.json
```diff
@@ -22,6 +22,7 @@
     "framer-motion": "^11.0.28",
     "idb-keyval": "^6.2.1",
     "invariant": "^2.2.4",
+    "io-ts": "^2.2.21",
```
jurevans, line 25: @pedrorezende Seems that all of the CI is failing due to this. Maybe we just need to update the `yarn.lock` file?

pedrorezende: Thanks @jurevans I missed the yarn.lock on my commit!

Fraccaman (approved):

Fraccaman:
@pedrorezende thanks, lgtm!

emccorson (approved):
Looks good! The only question I have is about how this will be used. If we get bug reports from the GitHub releases built through the CI, it should be easy to track down which version is the problem, but I'm not sure whether that will be the case for Docker builds from main. But I don't know, maybe it's not worth worrying about.

pedrorezende:
I think the idea is to have a different option to host Namadillo, once Docker is widely used. Also, the CI is generating the build, so we can do whatever we want with that package. On the other hand, the Docker image configures everything.

PR #928, opened 2024-07-12
Namadillo: Adding a cute logo

pedrorezende:
We now have a very nice and cute logo for Namadillo, that also close their eyes when you hide the balance :D


emccorson (approved):
Looks good to me!

PR #931, opened 2024-07-15
feat/922: Update namadillo for wasm checksums

### LINKED ISSUE #922: Extension: Remove wasm checksums from persisted storage, update client accordingly

### LINKED ISSUE #923: SDK/Indexer: Move fetch for wasm checksum hashes into the indexer
In https://github.com/anoma/namada-interface/pull/897, we added an SDK rpc query for wasm checksum hashes. This should also be added to the indexer:

- Introduce a query to fetch wasm code paths & the checksums associated with them from the indexer @ https://github.com/anoma/namada-indexer
- Update `namadillo` to use this instead of the SDK RPC query

jurevans:
Resolves #922 
Resolves #923 (Indexer changes are merged!)

This PR updates the wasm checksum hashes for details to allow a client to optionally pass these to the signer in order to display details. Previously, these were added to the extension and persisted, however, this would allow a site to potentially update invalid hashes which would prevent details from displaying. As it's already the responsibility of the interface to query and provide these, I've moved them to the `.sign()` call invoked on the integration as an optional argument. If they are not provided, then the commitments within a batch will not deserialize, but signing will still work!

- [x] Get checksums from indexer API in chain parameters
- [x] Remove `addTxWasmHashes` & `getTxWasmHashes` from integration, along with all handling in extension
- [x] Add `checksums` to `PendingTx` as optional prop, update messages accordingly
- [x] Update namadillo to provide these checksums when signing - this means that the methods which invoke sign will need to pass along the checksums from the React components where the checksums value is accessed

__NOTE__ the `sign()` API is also changed here: https://github.com/anoma/namada-interface/pull/815, this removed `tx_type` & `wrapper_tx_msg`, the the max num of args ultimately will be 3


packages/types/src/namada.ts
```diff
@@ -8,10 +8,13 @@ export type SignArbitraryProps = {
 };
 
 export type SignProps = {
+  // TODO: Simplify these props!
+  // Remove txType & wrapperTxMsg!
```
jurevans, line 12: Ignore, this is done in https://github.com/anoma/namada-interface/pull/815

packages/integrations/src/Namada.ts
```diff
@@ -56,17 +55,4 @@ export default class Namada implements Integration<Account, Signer> {
   public signer(): Signer | undefined {
     return this._namada?.getSigner();
   }
-
-  public async addTxWasmHashes(
```
emccorson, line 60: Very minor thing, but these are still present in the `types` docs.

emccorson (approved):
Looks good to me and works nicely!

mateuszjasiuk (approved):
Tested and works nicely!

apps/namadillo/src/lib/query.ts
```diff
@@ -117,6 +119,10 @@ export const signTx = async <T>(
   const integration = getIntegration(chain.id);
   const signingClient = integration.signer() as Signer;
 
+  const store = createStore();
```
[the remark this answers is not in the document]
jurevans: Got it, thanks!

packages/integrations/src/Namada.ts
jurevans: Thanks! Updated

PR #935, opened 2024-07-19
Extension: General fixes

### LINKED ISSUE #782: AccountChanged event emitted when clicking on the already default key
In the extension, if you click the checkbox on an account to set it as the default for an account that is _already_ the default, the `AccountChange` event is still emitted, even though the default account did not change.

### LINKED ISSUE #831: No scroll on the signer module: Increase height 600px
Screenshot 2024-06-02 at 10 07 45

Can we make the singer module height 600px which is an industry standard for accessibility :)

### LINKED ISSUE #857: Extension: Move Setup app to React 18 style root
Currently, the `Setup` is using the old React ~17 mode of rendering with ReactDOM - Like `Approvals` & `App`, this should be using `createRoot`. However, doing this under `Setup` causes accounts to be added twice and will result in errors.

- Move `index.tsx` in `Setup` over to React 18 style root
- Fix issue with the `Completion` process in account creation where accounts are being added twice

pedrorezende:
This PR fixes the following issues:
- Closes #831 
- Closes #782 
- Closes #857

### DIFF
--- a/apps/extension/src/App/index.tsx
+++ b/apps/extension/src/App/index.tsx
@@ -1,6 +1,6 @@
 import { AccountContextWrapper, VaultContextWrapper } from "context";
 import React from "react";
-import ReactDOM from "react-dom";
+import { createRoot } from "react-dom/client";
 import { HashRouter } from "react-router-dom";
 import { RequesterProvider } from "services";
 import { App } from "./App";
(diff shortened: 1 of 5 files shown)


jurevans (approved):
LGTM!

PR #938, opened 2024-07-22
fix: stop passing ActionButton props to DOM element

emccorson:
This was throwing a React error in the console.


euharrison (approved):

pedrorezende (approved):
Ooops :) Thanks for the fix! LGTM

PR #941, opened 2024-07-24
fix: serialize bigints to fix vote proposal signing

emccorson:
This fixes the extension being unable to sign vote proposal txs because the background script can't serialize the tx details object since it contains a bigint.


pedrorezende (approved):
Good one! LGTM!

jurevans (approved):
Nice!

PR #947, opened 2024-07-25
Show proposal activation time, remove cached proposals

emccorson:
### Changed
- Show proposal activation time instead of activation epoch

### Removed
- Remove unused IndexedDB proposal caching
  - This could be added again later if we need it, but it was complicating the proposal code a bit so I removed it. The ability to cache queries in IndexedDB is still present though.


pedrorezende (approved):
Great one! o/

PR #948, opened 2024-07-25
chore - Bump shared to namada 0.41.0

jurevans:
Update to support Namada `v0.41.0` (mainnet release)

- Bump Namada to `v0.41.0`
- Updated args for IBC Transfer
- some changes to shielded sync - left comments as we can't implement shielded sync the same way now, so this could be a new issue to look into
- Regenerate docs for `packages/types` to catch latest changes

### DIFF
--- a/packages/shared/lib/src/sdk/args.rs
+++ b/packages/shared/lib/src/sdk/args.rs
@@ -2,6 +2,7 @@ use std::{path::PathBuf, str::FromStr};
 
 use namada::core::borsh::{BorshDeserialize, BorshSerialize};
 use namada::core::ibc::core::host::types::identifiers::{ChannelId, PortId};
+use namada::ibc::IbcShieldingData;
 use namada::tendermint_rpc;
 use namada::tx::data::GasLimit;
 use namada::{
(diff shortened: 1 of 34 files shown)


packages/shared/lib/src/sdk/args.rs
```diff
@@ -534,11 +531,17 @@ pub fn ibc_transfer_tx_args(
     let amount = InputAmount::Unvalidated(denom_amount);
     let port_id = PortId::from_str(&port_id).expect("Port id to be valid");
     let channel_id = ChannelId::from_str(&channel_id).expect("Channel id to be valid");
+    let ibc_shielding_data = match shielding_data {
+        Some(v) => Some(IbcShieldingData::try_from_slice(&v)?),
+        None => None,
+    };
+
```
jurevans, line 538: See https://github.com/anoma/namada/blob/9b8252e6c3dd73678c00495b474bb7795a77b6c6/crates/ibc/src/msg.rs#L130-L137

We will need to implement something for `MaspTransaction` for shielding over IBC. Our schema can just handle the Borsh-serialized bytes, but that will need to be constructed in the interface, this just gets or schema up-to-date for IBC

mateuszjasiuk (approved):
LGTM! Tested We need to merge [this](https://github.com/anoma/namada-indexer/pull/86#pullrequestreview-2201846434) before. Otherwise txs will get rejected

PR #955, opened 2024-07-30
Namadillo: fixing some errors on re-delegate flow

pedrorezende:
This PR contains the following fixes:
- When Public Key was not revealed, the server was returning 404 and the app was throwing an unhandled exception
- After submitting re-delegate first form, nothing was happening. Now it goes to the next form
- Fixed a reference issue after deleting an object property on re-delegate


jurevans (approved):
LGTM!

PR #958, opened 2024-08-02
fix: fix notifications were not being removed

euharrison:
Fix the notifications were not being removed when deploying the Namadillo

The issue was caused because the transaction event listeners weren't initialized due to the hook `useEffectSkipFirstRender`, which skips the listener creation on first render.

I'm not aware why it was implemented like this in the past, maybe it was to fix another issue of a concurrency listener creation.

Anyway, I created the individual hook listener `useTransactionEventListener` that handle the subscribe and also the unsubscribe with `useEffect`, which allow us to reinvoke the method anytime we want and it's more a "react oriented" solution

### DIFF
--- a/apps/namadillo/src/utils/index.ts
+++ b/apps/namadillo/src/utils/index.ts
@@ -2,7 +2,6 @@ import { ProposalStatus, ProposalTypeString } from "@namada/types";
 import * as fns from "date-fns";
 import { DateTime } from "luxon";
 import { useEffect } from "react";
-import { EventData, TransactionEvent } from "types/events";
 
 export const proposalStatusToString = (status: ProposalStatus): string => {
   const statusText: Record<ProposalStatus, string> = {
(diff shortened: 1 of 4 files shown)


apps/namadillo/src/utils/index.ts
```diff
@@ -33,11 +34,16 @@ export const epochToString = (epoch: bigint): string =>
 export const proposalIdToString = (proposalId: bigint): string =>
   `#${proposalId.toString()}`;
 
-export const addTransactionEvent = <T>(
+export const useTransactionEventListener = <T>(
```
emccorson, line 37: There is also `useEventListener` in `@namada/hooks` which does almost the same as this function, but I think either way is fine.

emccorson (approved):
Looks good to me and works well!

I was able to reproduce this locally by running Namadillo in production mode. I guess the behaviour you and @jurevans were seeing with the deployed version was that Namadillo was running in production mode so the  `App` component was only mounted once (and therefore not adding the event listeners), but locally it was mounted twice because of development mode.

apps/namadillo/src/utils/index.ts
euharrison: Thanks @emccorson ! I didn't know it, good to know!

I tried to use the hook, but the typescript change become a big lift. So I'll merge this PR as it's and prepare it in another PR, so we have a better visibility over the changes and the refactor proposal

PR #959, opened 2024-08-04
CI yarn install timeout fix + various CI improvements

emccorson:
Various CI improvements, mainly to speed up CI runs and hopefully fix the yarn install timeouts.

Currently the CI is running for PRs at around 20 minutes, but once this is merged into main PRs should take less than 10 minutes to run in the CI.

Example yarn install timeout:
https://github.com/anoma/namada-interface/actions/runs/10198828123/job/28214547488

I'm not 100% sure why the yarn install is sometimes timing out. I think it could be to do with how puppeteer is downloaded to the home directory and then linked by yarn, so I've added puppeteer to the yarn cache. If the problem comes up again after this PR, I'll look into it further.

---

### Changed
- Cache puppeteer to speed up yarn install and hopefully fix timeouts
- Use a different Rust cache action (seems faster)
- Run yarn without hardened mode and separately check lockfile poisoning
  - See https://yarnpkg.com/features/security#hardened-mode
- Disable e2e job as no tests are being run
- Run main workflow on push to main regardless of files changed
- Run PR workflow regardless of files changed
- Name PR jobs and rename PR workflow
- Bump checkout action to get newer Node version

### Added
- Add Rust multicore cache to speed up Chrome extension build

### Removed
- Remove slack notifications
- Remove wasm-pack install (already handled by yarn)
- Remove deploy to Netlify for PRs
- Remove build artifacts for PRs

### DIFF
--- a/.github/workflows/deploy-wallet-at-pr.yml
+++ b/.github/workflows/deploy-wallet-at-pr.yml
@@ -1,20 +1,18 @@
-name: Deploy wallet preview to netlify at PR and pushes to it
+name: PR checks
 on:
   pull_request:
-    paths:
-      - "apps/namadillo/**"
-      - "apps/extension/**"
-      - "packages/**"
-      - ".github/workflows/**"
 
 env:
   CI: false
+  YARN_ENABLE_HARDENED_MODE: 0
+
 jobs:
   lint:
+    name: Lint
     runs-on: ubuntu-latest
     steps:
       - name: Checkout repository
-        uses: actions/checkout@v3
+        uses: actions/checkout@v4
 
       - name: Install yarn dependencies
         uses: ./.github/actions/yarn-cache
@@ -23,10 +21,11 @@ jobs:
         run: yarn lint:ci
 
   unit-tests-js:
+    name: JS unit tests
     runs-on: ubuntu-latest
     steps:
       - name: Checkout repository
-        uses: actions/checkout@v3
+        uses: actions/checkout@v4
 
       - name: Install yarn dependencies
         uses: ./.github/actions/yarn-cache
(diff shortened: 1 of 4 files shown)


.github/workflows/deploy-wallet-at-pr.yml
```diff
@@ -125,55 +88,28 @@ jobs:
...
-        run: rustup target add wasm32-unknown-unknown
-
       - name: build the site
         working-directory: ./apps/namadillo
         run: yarn build
         env:
           NAMADA_INTERFACE_NAMADA_ALIAS: "Namada Devnet"
           NAMADA_INTERFACE_NAMADA_CHAIN_ID: "internal-devnet-6be.86067e06a5"
           NAMADA_INTERFACE_NAMADA_URL: "https://proxy.heliax.click/internal-devnet-6be.86067e06a5"
-      - name: Deploy to Netlify
```
euharrison, line 143: out of curiosity, is there a reason to not deploy on this case?

.github/workflows/deploy-wallet-at-pr.yml
```diff
@@ -186,25 +122,12 @@ jobs:
         working-directory: ./apps/extension
         run: yarn build:chrome
 
-      # GitHub actions zips artifacts so we need to unzip the built extension to
-      # avoid a double zip
-      - name: Prepare artifact
```
euharrison, line 191: If I understood correctly, @jurevans use this to build the extension and share with the team

emccorson: It is actually this one that is used for the extension build:
https://github.com/anoma/namada-interface/actions/workflows/release-wallet.yml

As far as I know the artifacts are not used from this workflow, so I removed them.

.github/workflows/deploy-wallet-at-pr.yml
emccorson: It's basically just not being used right now. I wasn't sure but I checked with Justin a couple of weeks ago.

PR #961, opened 2024-08-05
fix: fix broken "Back Up My Wallet" extension link

### LINKED ISSUE #960: Extension: "Back Up My Wallet" link when deleting keys doesn't work
On the Delete Keys page, the link "Back Up My Wallet" sends the user to an invalid URL and the page is not found.

emccorson:
Fixes #960.


euharrison (approved):

euharrison:
Thinking loud, we could explore an alternative of exporting only the classname of `<LinkButton>` then use it directly on the `<Link>` component, on the `DeleteAccount`

Or, we could accept the idea of importing `react-router` on the root of the monorepo, then the `<Link>` component could be used by `<LinkButton>`

PR #981, opened 2024-08-07
Fix gas price and gas limit in tx details

### LINKED ISSUE #972: Extension: feeAmount and gasLimit are always 1 in transaction details
2024-08-07-113813_776x727_scrot

emccorson:
Closes #972.

### Fixed
- Show tx detail gas limit in gas units, not NAM
- Show gas price, not total fees, in tx details
  - This field is still named `feeAmount`, but I was wondering about changing it to `gasPrice`, which would match what Namada CLI calls it and I think is a less confusing name. I can add it to this PR or do it later if we decide to change the name.

---

Example transaction with gas price 0.00001 NAM and gas limit 1000000:

2024-08-07-163707_484x784_scrot


mateuszjasiuk (approved):
Nice! LGTM I like `gasPrice` more than `feeAmount` :)

PR #982, opened 2024-08-07
feat: improve canWithdraw logic

mateuszjasiuk:
From slack:
> Small update regarding withdraw button. So I changed code a bit, now we compare current epoch to withdraw epoch to determine if withdraw is possible(before we were comparing time left to current time). Also if time left is obviously incorrect(this might happen if the genesis parameters are wrong) we will display "N/A". But note that  this is only a fallback. For live version we should never see "N/A" as genesis parameters have to be correct  :smile: So this is only for testing in case someone gets confused because withdraws are not working.

Use branch from this [PR](https://github.com/anoma/namada-indexer/pull/95) or wait for merge :) 🙏

### DIFF
--- a/apps/namadillo/src/atoms/validators/functions.ts
+++ b/apps/namadillo/src/atoms/validators/functions.ts
@@ -89,9 +89,9 @@ export const toUnbondingValidators = (
 
     const canWithdraw = indexerUnbond.canWithdraw;
     const timeLeft =
-      // If can't withdraw but estimation is incorrect display n/a
-      !canWithdraw && withdrawTime < timeNow ? "N/A"
-      : canWithdraw ? ""
+      canWithdraw ? ""
+        // If can't withdraw but estimation is incorrect display withdraw epoch
+      : withdrawTime < timeNow ? `Epoch ${indexerUnbond.withdrawEpoch}`
       : singleUnitDurationFromInterval(timeNow, withdrawTime);
 
     const amountValue = BigNumber(indexerUnbond.amount);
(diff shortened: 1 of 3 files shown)


euharrison (approved):

apps/namadillo/src/atoms/validators/functions.ts
```diff
@@ -86,13 +86,13 @@ export const toUnbondingValidators = (
...
-    const currentTime = Number(indexerUnbond.currentTime);
-    const secondsLeft = withdrawTime - currentTime;
 
-    // TODO: later return from the backend
-    const canWithdraw = secondsLeft <= 0;
+    const canWithdraw = indexerUnbond.canWithdraw;
     const timeLeft =
-      canWithdraw ? "" : singleUnitDurationFromInterval(timeNow, withdrawTime);
+      // If can't withdraw but estimation is incorrect display n/a
+      !canWithdraw && withdrawTime < timeNow ? "N/A"
```
euharrison, line 93: What do you think about this?

Screenshot 2024-08-07 at 11 11 11

```ts
const timeLeft =
  canWithdraw ? ""
  : withdrawTime < timeNow ? `Epoch ${indexerUnbond.withdrawEpoch}`
  : singleUnitDurationFromInterval(timeNow, withdrawTime);
```

emccorson (approved):
Looks good to me and should be fine for testing!

apps/namadillo/src/atoms/validators/functions.ts
mateuszjasiuk: Looks good to me. I thought about it, but I wasn't sure how useful this is because we do not display current epoch anywhere in the interface :D

euharrison: Yeah, I was searching for a place with the current epoch as well 😅 
But at least it give some information, even though it'll be hard to understand

Issue #970, opened 2024-08-07
Namadillo: Indexer URL can't be fixed if you accidentally use RPC URL as indexer URL


emccorson:
In the setup step, if you enter an RPC URL as the indexer URL, Namadillo will accept the URL, but then Namadillo will get stuck trying to load and there is no option to fix the URL (see screenshot).

We may need a better check to see if a URL is a valid indexer URL. I think currently it pings an endpoint that exists on both indexer and RPC URLs, which is why an RPC URL is allowed as an indexer URL by Namadillo.

We should maybe also allow fixing the URL in this case instead of showing the "Try again" link only.

![2024-08-07-110346_1021x371_scrot](https://github.com/user-attachments/assets/75223c0d-cadd-465c-8e82-41b31d149a99)

PR #983, opened 2024-08-07
feat: allow to change the indexer url on error

euharrison:
Resolves https://github.com/anoma/namada-interface/issues/970 adding the link to update the Indexer Url

Screenshot 2024-08-07 at 11 42 21

PR #984, opened 2024-08-07
fix: enable refetch on try again button

euharrison:
Relates to https://github.com/anoma/namada-interface/issues/970

Enable the `Try Again` button to refetch the chain atom

https://github.com/user-attachments/assets/96d1d49f-eecc-4dfb-9acc-1ae6eb048d68

### DIFF
--- a/apps/namadillo/src/atoms/chain/atoms.ts
+++ b/apps/namadillo/src/atoms/chain/atoms.ts
@@ -2,36 +2,26 @@ import namada from "@namada/chains/chains/namada";
 import { integrations } from "@namada/integrations";
 import { indexerApiAtom } from "atoms/api";
 import {
-  defaultServerConfigAtom,
   indexerUrlAtom,
   namadaExtensionConnectedAtom,
   rpcUrlAtom,
 } from "atoms/settings";
 import { queryDependentFn } from "atoms/utils";
+import { atom } from "jotai";
 import { atomWithQuery } from "jotai-tanstack-query";
 import { ChainParameters, ChainSettings } from "types";
 import { fetchChainParameters, fetchRpcUrlFromIndexer } from "./services";
 
 export const chainAtom = atomWithQuery<ChainSettings>((get) => {
   const chainParameters = get(chainParametersAtom);
-  const indexerRpcUrl = get(indexerRpcUrlAtom);
   const rpcUrl = get(rpcUrlAtom);
-  const indexerUrl = get(indexerUrlAtom);
-  const tomlConfig = get(defaultServerConfigAtom);
   const extensionConnected = get(namadaExtensionConnectedAtom);
 
   return {
-    queryKey: [
-      "chain",
-      rpcUrl,
-      chainParameters,
-      indexerRpcUrl,
-      extensionConnected,
-    ],
+    queryKey: ["chain", rpcUrl, chainParameters.data, extensionConnected],
     staleTime: Infinity,
     retry: false,
     ...queryDependentFn(async () => {
-      const rpcUrl = get(rpcUrlAtom);
       const extensionChainId =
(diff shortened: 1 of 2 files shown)


apps/namadillo/src/atoms/chain/atoms.ts
```diff
@@ -48,7 +38,19 @@ export const chainAtom = atomWithQuery<ChainSettings>((get) => {
         nativeTokenAddress: chainParameters.data!.nativeTokenAddress,
         checksums: chainParameters.data!.checksums,
       };
-    }, [!!indexerUrl, indexerRpcUrl, chainParameters, tomlConfig]),
```
euharrison, line 51: Please, review here carefully

I removed the `indexerRpcUrl` and `tomlConfig` because they don't have direct relation with the `queryFn`, meaning they are useless for it

If it's used by a child request, we should add the `queryDependentFn` on this child request to handle this case

Or if it's another edge case, we should add a comment for future developers

apps/namadillo/src/atoms/chain/atoms.ts
```diff
@@ -2,36 +2,26 @@ import namada from "@namada/chains/chains/namada";
...
 
   return {
-    queryKey: [
-      "chain",
-      rpcUrl,
-      chainParameters,
-      indexerRpcUrl,
-      extensionConnected,
-    ],
+    queryKey: ["chain", rpcUrl, chainParameters.data, extensionConnected],
```
euharrison, line 21: Passing only the `.data` because this is what we really use by the `queryFn`. Also, removed the other values as the comment below

emccorson (commented):
Firstly, I actually hadn't noticed that the "Try again" link doesn't work, so I was confused by your comment in #983, but it makes sense now :smile: 

`chainAtom` already has a refetch function since it is an `AtomWithQueryResult`, but the refetch function isn't working because of the way `queryDependentFn` is written. If there are errors in a dependency (in this case, `chainParametersAtom`), it will never try to refetch the dependency.

I think we need a general way of dealing with this, since it will also be a problem in other atoms, but I'm not sure of a good way yet. One quick fix would be to add the dependencies to the `result` prop of `AtomErroryBoundary` (haven't tested but I _think_ that should work). It's not ideal though and maybe there's a way of changing `queryDependentFn` to refetch dependencies if they have errors.

euharrison:
That's a tough problem!

In my past experience we used custom hooks with react-query instead of the atom, which give us more freedom.
Maybe this is an alternative to simplify. Something like this:

```tsx
export const useChain = () => {
  const chainParameters = useAtomValue(chainParametersAtom);
  const rpcUrl = useAtomValue(rpcUrlAtom);
  const extensionConnected = useAtomValue(namadaExtensionConnectedAtom);

  const chain = useQuery({
    // similar implementation
  });

  return {
    data: chain.data,
    isPending: chainParameters.isPending || chain.isPending,
    isError: chainParameters.isError || chain.isError,
    error: chainParameters.error || chain.error,
    refetch: async () => {
      await chainParameters.refetch();
      return chain.refetch();
    },
  };
};
```

We can create some helper functions to minimize the code boilerplate like what the `queryDependentFn` does.

Another alternative is using the suspense, but we'll need to confirm that our app is prepared to use it with the `<Suspense>` handler

There are some examples in the docs: https://tanstack.com/query/latest/docs/framework/react/guides/request-waterfalls

pedrorezende:
Hello, hello. Maybe I'm missing something, but can't we fix this issue by just adding  `result={[chainParameters, chain]}` to the related `<AtomErrorBoundary>` ?

euharrison:
https://github.com/anoma/namada-interface/pull/1014 will resolve this issue

PR #993, opened 2024-08-13
feat(namadillo): avoid retrying for 4xx errors

pedrorezende:
We don't need to refetch if the server responds with a client / request error (HTTP 4xx).

This PR adds this implementation, based on this section of jotai-react-query docs https://github.com/jotaijs/jotai-tanstack-query?tab=readme-ov-file#referencing-the-same-instance-of-query-client-in-your-project .


jurevans (approved):
Looks good!

PR #994, opened 2024-08-13
feat(namadillo): adding validation to unstake amounts

pedrorezende:
This PR won't let users to unbond more than what they have in a validator, causing an error. Displays a simple "Invalid amount" message and changes the color to red for the invalid amounts.


euharrison (approved):

jurevans (approved):
LGTM!

PR #998, opened 2024-08-14
Fix vote tx details on Firefox + uncaught error

emccorson:
### Fixed
- Namadillo: fix uncaught error in proposal time remaining
- Extension: show proposal id and vote tx details in Firefox
  - This is another case of a Firefox bug due to the difference in how Firefox and Chrome serialize messages passed between the background and content scripts. Firefox uses the structured clone algorithm but Chrome uses JSON serialization.


pedrorezende (approved):
LGTM!

jurevans (approved):
Looks good!

PR #1002, opened 2024-08-15
feat: render the version number

euharrison:
Render the version number on the Settings page

Screenshot 2024-08-15 at 11 58 34


jurevans (approved):
LGTM!

emccorson:
Sorry, just seeing this PR now.

I think this will not work for the version of Namadillo built by the CI (the one I think we're planning to distribute) because the package.json file is not included.

When we were using webpack, we used to be able to get the version via `process.env.npm_package_version` and this would get replaced by the actual version string at build time. It's still configured in `apps/namadillo/webpack.config.js` if you want to see how it's set up. I don't know if we have that in the Vite config yet though.

PR #1005, opened 2024-08-16
feat: start the checker ui minimized

euharrison:
Start the checker UI minimized, to make it less intrusive during development

Screenshot 2024-08-16 at 12 08 20


jurevans (approved):
LGTM!

PR #1007, opened 2024-08-16
Namadillo: fixing error on loading token

pedrorezende:
We've introduced some different default settings for the store in a preview commit, however `getDefaultStore` in some other places started to refer to a store different than the one used by jotai-tanstack-query, leading to some unexpected behaviours. This PR simply adds this reference to the query provider.


jurevans (approved):
LGTM! Works perfectly

euharrison (approved):

PR #1010, opened 2024-08-19
feat/1009 - Namadillo: Remove hide balances feature

### LINKED ISSUE #1009: Namadillo: Remove "Hide Balances" feature
It has been requested that we remove the "Hide Balances" feature from Namadillo. The rationale being that it may be confused with shielding features.

- Remove `Hide Balances` toggle
- Remove logic for swapping balances with asterisks

jurevans:
Resolves #1009 

- [x] Remove `Hide Balances` toggle
- [x] Update settings and Namadillo components

### DIFF
--- a/apps/namadillo/src/App/Common/AppContainer.tsx
+++ b/apps/namadillo/src/App/Common/AppContainer.tsx
@@ -1,6 +1,4 @@
-import { hideBalancesAtom } from "atoms/settings";
 import clsx from "clsx";
-import { useAtomValue } from "jotai";
 import { useState } from "react";
 import { Link } from "react-router-dom";
 import { BurgerButton } from "./BurgerButton";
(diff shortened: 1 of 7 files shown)


apps/namadillo/src/App/Common/AppContainer.tsx
```diff
@@ -37,9 +33,6 @@ export const AppContainer = ({
               "flex items-center gap-3 text-yellow text-xl not-italic uppercase"
             )}
           >
-            <i className="w-[40px]">
-              <Logo eyeOpen={!hideBalance} />
```
pedrorezende, line 41: I think we should keep the logo, just setting the eyeOpen property as true

pedrorezende (changes requested):

apps/namadillo/src/App/Common/AppContainer.tsx
jurevans: Done!

pedrorezende (approved):
LGTM!

PR #1014, opened 2024-08-19
feat: move chain error to page scope

euharrison:
Move chain error to page scope

To achieve this, the `AppSetup` was splited into multiple individual `Loader`s: `TomlConfigLoader`, `IndexerLoader`, `ExtensionLoader`, and `ChainLoader`. Each of this component will be gate and responsible to load and handle errors for its own scope

Additionally, the `try again` and when the user change the indexer/rpc url, the app execute a full browser reaload. This is important because we are not listening for `api.configuraion` (or `api.basePath`) in the atoms queries. Then, to avoid, a huge refactoring, it's easier to execute a full reload

And the bonus is the `update settings` button can now redirect the user to `/settings/advaced` instead of having a "black screen of death"

The Sync Indicator was also updated to handle the heartbeat atom, adding a `refetchInterval` of 10 seconds

Screenshot 2024-08-19 at 18 07 04


apps/namadillo/src/App/Setup/ChainLoader.tsx
```diff
@@ -0,0 +1,79 @@
...
+
+  if (!chain.data?.rpcUrl) {
+    return (
+      <ErrorBox
+        containerProps={errorContainerProps}
+        niceError={
+          <>
+            <p>RPC Url was not provided</p>
+            <p className="mt-1">
+              Do you like to <UpdateSettingsButton />?
```
jurevans, line 70: We should probably change this to `Would you like to...`

jurevans (approved):
Nice! Works great for me! Left one minor comment

PR #1016, opened 2024-08-20
Extension - address feedback

### LINKED ISSUE #1015: Extension: UX/UI changes per feedback
- Update address shortening on Key Import confirmation to show as many characters as possible
- Remove Chain query for Extension (Namadillo should _only_ acquire Chain ID from indexer, and user can update extension setting if need be)

jurevans:
Resolves #1015 

- [x] Namadillo should use chain ID from indexer only
- [x] Lengthen "shortened" address in set up (Key import) - Note: I updated the component to accept a trim value for `shortenAddress()`, otherwise it will default to 16, and tried to find values in Setup and View Keys that would work well with different keys through trial and error. It should be close, but I imagine with some keys it may be _possible_ that part of a character clips, if so, we can reduce the trim in that case.

### Screenshots

Lengthen keys in Setup view:
Screen Shot 2024-08-20 at 11 30 47 AM

Lengthen keys in `View Keys` view:
Screen Shot 2024-08-20 at 11 32 59 AM

### DIFF
--- a/apps/extension/src/App/Accounts/ViewAccount.tsx
+++ b/apps/extension/src/App/Accounts/ViewAccount.tsx
@@ -61,6 +61,7 @@ export const ViewAccount = (): JSX.Element => {
               publicKeyAddress={parentAccount.publicKey ?? ""}
               transparentAccountAddress={transparentAddress}
               shieldedAccountAddress={shieldedAddress}
+              trimCharacters={21}
             />
           </Stack>
           <ActionButton size="md" onClick={() => navigate(-1)}>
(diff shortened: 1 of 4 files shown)


euharrison:
@jurevans my impression was that we should increase the box width to fit the whole text without the ellipsis, no?

euharrison (approved):

jurevans:
> @jurevans my impression was that we should increase the box width to fit the whole text without the ellipsis, no?

@euharrison I believe we were to use as much of the space so we can to see more of the address. It was configured to work in the `View Keys` view, which left a lot of blank space when that same setting was used in Setup

PR #1033, opened 2024-08-21
feat/1012: Improve Tx API in SDK

### LINKED ISSUE #1012: SDK: Improve result of build functions
Currently, all of the `build_` functions (including `build_batch`) in the `@heliax/namada-sdk` package return a Rust struct instance (wasm pointer). It would be better to return a deserialized value from the SDK calls so they can be more easily inspected and destructured. We could Borsh-serialize/deserialize the `BuiltTx` type, and deserialize when returning from the SDK

Example of what this value would look like:

```typescript
import { WrapperTxProps } from "@namada/types";

// We'll use borsh schemas instead of type defs, but essentially what the user
// will see is this:

type AccountPublicKeysMap = {
  pkToIdx: [string, number];
  idxToPk: [number, string];
}

type SigningTxData = {
  publicKeys: string[];
  threshold: number;
  feePayer: string;
  accountPublicKeysMap?: AccountPublicKeysMap[];
  owner?: string;
}

type Tx = {
  // tx hash
  hash: string;
  // tx bytes
  bytes: Uint8Array;
  // args used to create args::Tx in namada-sdk
  args: WrapperTxProps;
  // serialized signing data bytes
  signingDataBytes: SigningTxData[];
}
```

jurevans:
Resolves #1012 

This PR refactors the response from the various `build_` functions, including `build_batch`, to be able to provide via the SDK package a deserialized object that can be inspected and destructured. I think in the future, the `@heliax/namada-sdk` should never return a reference to a Rust struct, as it's clunky and breaks TS convention to use. This PR returns an normal TS type where values can be inspected and destructured.

- [x] New Tx type and related schema with all values required in client
- [x] Hook up Tx to existing build functions
- [x] Update SDK package to deserialize these values from Rust before returning to client
- [x] Regenerate docs once types & API are finalized: See new Tx type in https://github.com/anoma/namada-interface/blob/33c6485fc044f5212c5bb2501c306f8d5b6db170/packages/types/docs/classes/TxMsgValue.md

### NOTES

Instead of returning an instance of a Rust struct `BuiltTx` to the client, return a deserialized tx value:

```typescript
import { WrapperTxProps } from "@namada/types";

// We'll use borsh schemas instead of type defs, but essentially what the user
// will see is this:

type SigningData = {
  publicKeys: string[];
  threshold: number;
  feePayer: string;
  accountPublicKeysMap?: Uint8Array;
  owner?: string;
}

type Tx = {
  // tx hash
  hash: string;
  // tx bytes
  bytes: Uint8Array;
  // args used to create args::Tx in namada-sdk
  args: WrapperTxProps;
  // deserialized signing data 
  signingData: SigningData[];
}
```
This makes the result of any build function something that can be inspected or destructured.

### Testing

- Simply build this branch and test existing Tx functionality

### Example Tx

```json
{
    "args": {
        "token": "tnam1q87wtaqqtlwkw927gaff34hgda36huk0kgry692a",
        "feeAmount": "0.000001",
        "gasLimit": "300000",
        "chainId": "internal-devnet-43.35fd77a4447",
        "publicKey": "tpknam1qr4w8d26738k7zmhxhsp9987puunenrjt4kxp5f5k8wlgey3zzuyqg9tgqh",
        "memo": ""
    },
    "hash": "0CEBA68C9BB9391B5481C46069B32EFE45758E77C5114D676455F471A711A2F5",
    "bytes": {
        "type": "Buffer",
        "data": [
          ...
        ]
    },
    "signingData": [
        {
            "owner": "tnam1qqql82l8d8xcgu8q598gl29vecmelc9v6vfyzwn3",
            "publicKeys": [
                "tpknam1qr4w8d26738k7zmhxhsp9987puunenrjt4kxp5f5k8wlgey3zzuyqg9tgqh"
            ],
            "threshold": 1,
            "accountPublicKeysMap": {
                "type": "Buffer",
                "data": [
                   ...
                ]
            },
            "feePayer": "tpknam1qr4w8d26738k7zmhxhsp9987puunenrjt4kxp5f5k8wlgey3zzuyqg9tgqh"
        }
    ]
}
```


euharrison (approved):
LGTM ✅

apps/extension/src/Setup/Ledger/LedgerConfirmation.tsx
```diff
@@ -22,6 +22,7 @@ export const LedgerConfirmation = (): JSX.Element => {
       <ViewKeys
         publicKeyAddress={account.publicKey}
         transparentAccountAddress={account.address}
+        trimCharacters={35}
```
jurevans, line 25: Missed this in a previous PR

packages/shared/lib/src/sdk/mod.rs
```diff
@@ -296,42 +223,54 @@ impl Sdk {
...
+
+        for bytes in built_txs_bytes.iter() {
+            let tx: tx::Tx = borsh::from_slice(&bytes)?;
+            built_txs.push(tx);
+        }
+
+        // Get wrapper args
+        let first_tx = built_txs
+            .iter()
+            .nth(0)
```
mateuszjasiuk, line 247: You can also just `.get(0)` without converting vec to iter :)

mateuszjasiuk (approved):
Tested everything except withdraw as I have to wait 8 hours xD LGTM!

packages/shared/lib/src/sdk/mod.rs
jurevans: Thanks! I updated it in a couple places. Just re-tested `Withdraw`, working fine!

PR #1034, opened 2024-08-21
feat: update unbonding period calculation

### LINKED ISSUE #1020: Resolve the word TODO on Staking
Resolve the word TODO on Staking

Screenshot 2024-08-20 at 13 20 30

euharrison:
Update Unbonding Period calculation

- Use an unified calculation and save it on `chain.unbondingPeriod` on the atom layer, removing it from different components and any reference for `unbondingPeriodInEpochs` which leads to distributed and confused calculations
- Remove the data transformation from the `service`, than it'll be used only to execute the indexer request, and all data transformation will happen on the atom layer
- Update the unbondoing period on the staking page `Staking/IncrementBonding.tsx`

Closes #1020 

Screenshot 2024-08-21 at 13 51 09

Screenshot 2024-08-21 at 13 51 25

Screenshot 2024-08-21 at 13 51 28

Screenshot 2024-08-21 at 13 52 14


apps/namadillo/src/atoms/chain/atoms.ts
```diff
@@ -83,6 +84,33 @@ export const chainParametersAtom = atomWithQuery<ChainParameters>((get) => {
     queryKey: ["chain-parameters", indexerUrl],
     staleTime: Infinity,
     enabled: !!indexerUrl,
-    queryFn: async () => fetchChainParameters(api),
+    queryFn: async () => {
+      const parameters = await fetchChainParameters(api);
+
+      const unbondingPeriodInEpochs =
```
mateuszjasiuk, line 90: I think this is good! :) But maybe it's worth moving it to some common/utility space(component?) if it's used in multiple places?

euharrison: Thanks! I'll move to a dedicated function

euharrison: Thanks! I'll move it to a dedicated function. But the idea here is to keep it attached to the atom, then anytime we needed, we pull the value directly from the atom and we don't need to care about the math/parameters anymore

emccorson (commented):
I've not reviewed the code, but just on the TODO replacement: I think "a 13 hours unbonding schedule" doesn't read right. It could either be "a 13-hour unbonding schedule" or "an unbonding schedule of 13 hours". Maybe the second is easier since the string formatting doesn't need to change.

mateuszjasiuk (approved):

PR #1035, opened 2024-08-21
feat: add account address tooltip

euharrison:
Add account address tooltip

Screenshot 2024-08-21 at 14 42 01


pedrorezende (approved):
LGTM! 

I've opened the #1038 because I just noticed that the Storybook docs needs to be improved for this component

PR #1036, opened 2024-08-21
feat: resolve forceBalanceDisplay warning

euharrison:
Resolve `forceBalanceDisplay` warning

Also, enable the `ignoreRestSiblings: true` which is great for removing one prop from `props` to pass to children and could help on future issues similar to this one:

```tsx
({ forceBalanceDisplay, ...props })  => <div {...props} />
```

Screenshot 2024-08-21 at 14 55 51


pedrorezende (approved):
LGTM!

PR #1043, opened 2024-08-23
fix: fix fees and redelegating to same validator

### LINKED ISSUE #975: Namadillo: transaction fee is incorrect
When submitting a bonding transaction on the latest devnet, I'm getting the transaction fee printed as 0.000002 NAM, but in the extension it shows 1 NAM, and what seems to actually be charged is 1 NAM. Probably related to #972.

emccorson:
- Add default gas config and use it to display correct transaction fees

- Disallow redelegating when it would result in no transactions (e.g. redelegating to the same validator)

- Refactor redelegate validation and display clearer error message in submit button

Fixes #975.


jurevans (approved):
Nice! LGTM!

PR #1044, opened 2024-08-23
bug: governance and unbonding fixes

mateuszjasiuk:
- fixed epoch estimations
- fix for vote button - now disabled if address did not have active bonds before proposal stared
- added polling for proposal view after sending vote tx
- small changes due to using new indexer client generation library

Test it with 0.43.0 and [this ](https://github.com/anoma/namada-indexer/pull/94) indexer

### DIFF
--- a/apps/namadillo/src/atoms/proposals/atoms.ts
+++ b/apps/namadillo/src/atoms/proposals/atoms.ts
@@ -21,15 +21,16 @@ import {
   fetchVotedProposalIds,
 } from "./functions";
 
-import {
-  Bond as NamadaIndexerBond,
-  BondStatusEnum as NamadaIndexerBondStatusEnum,
-} from "@anomaorg/namada-indexer-client";
+import { Bond as NamadaIndexerBond } from "@anomaorg/namada-indexer-client";
+import { shouldUpdateProposalAtom } from "atoms/etc";
 export const proposalFamily = atomFamily((id: bigint) =>
   atomWithQuery((get) => {
     const api = get(indexerApiAtom);
+    const enablePolling = get(shouldUpdateProposalAtom);
 
     return {
+      // TODO: subscribe to indexer events when it's done
+      refetchInterval: enablePolling ? 1000 : false,
       queryKey: ["proposal", id.toString()],
       queryFn: () => fetchProposalById(api, id),
     };
@@ -96,7 +97,11 @@ export const paginatedProposalsFamily = atomFamily(
 export const votedProposalIdsAtom = atomWithQuery((get) => {
   const account = get(defaultAccountAtom);
   const api = get(indexerApiAtom);
+  const enablePolling = get(shouldUpdateProposalAtom);
+
   return {
+    // TODO: subscribe to indexer events when it's done
+    refetchInterval: enablePolling ? 1000 : false,
     queryKey: ["voted-proposal-ids", account.data],
     ...queryDependentFn(async () => {
       if (typeof account.data === "undefined") {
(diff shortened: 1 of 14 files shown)


jurevans (approved):
Code LGTM! I'm having a hard time testing atm, but everything looks in order!

apps/namadillo/src/atoms/proposals/atoms.ts
```diff
@@ -107,25 +112,29 @@ export const votedProposalIdsAtom = atomWithQuery((get) => {
...
+    const account = get(defaultAccountAtom);
+    const api = get(indexerApiAtom);
 
-  return {
-    queryKey: ["can-vote"],
-    enabled: account.isSuccess,
-    queryFn: async () => {
-      const all_bonds = await api.apiV1PosBondAddressGet(account.data!.address);
+    return {
+      queryKey: ["can-vote"],
```
emccorson, line 121: Not introduced in this PR, but `queryKey` is missing the dependencies (definitely `account` and I think possibly `api` too), so this doesn't react when the default account is changed in the extension.

apps/namadillo/src/hooks/useTransactionCallbacks.tsx
```diff
@@ -19,4 +21,27 @@ export const useTransactionCallback = (): void => {
   useTransactionEventListener("Bond.Success", onBalanceUpdate);
   useTransactionEventListener("Unbond.Success", onBalanceUpdate);
   useTransactionEventListener("Withdraw.Success", onBalanceUpdate);
+
+  const store = createStore();
```
emccorson, line 25: It might be better to use `getDefaultStore` here in case you get the same problems as in #1007.

apps/namadillo/src/hooks/useTransactionCallbacks.tsx
```diff
@@ -19,4 +21,27 @@ export const useTransactionCallback = (): void => {
...
+  useTransactionEventListener("VoteProposal.Success", (args) => {
+    // TODO: refactor this after event subscription is enabled on indexer
+    shouldUpdateProposal(true);
+
+    // [0] is fine for time being as we can only vote for one proposal at a time
+    const proposalId = args.detail.data[0].proposalId;
+    const { refetch: refetchProposalFamily } = store.get(
+      proposalFamily(proposalId)
+    );
+    refetchProposalFamily();
```
emccorson, line 36: Just to double check: Is this needed because setting `refetchInterval` on the atom will mean we need to wait 1 second before the first refetch?

emccorson (approved):
Looks good to me! I am just waiting for my proposal to become active on the testnet in order to test this.

emccorson (commented):
I just finished testing now that my proposal is live. The voting seems to work well and the times look good, but Namadillo seems to get stuck polling for the proposal data forever even after I refresh. I'm not sure exactly where that's happening though.

apps/namadillo/src/atoms/proposals/functions.ts
```diff
@@ -257,16 +258,16 @@ const fromIndexerStatus = (
 
 const toIndexerStatus = (
   proposalStatus: ProposalStatus
-): IndexerProposalStatusEnum => {
+): ApiIndexerProposalStatusEnum => {
```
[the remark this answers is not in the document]
mateuszjasiuk: So there are two enums specified in swagger file. One is for status used in query params and other is used as a part of proposal response type. I think they should always be the same(kind of), but in rust we have two, because query params and response are two separate things.
FIrst:
```
        status:
          type: string
          enum: [pending, voting, passed, rejected]
```
Second:
```
          schema:
            type: string
            enum: [pending, votingPeriod, passed, rejected]
```
It's a bit confusing :D We can potentially leave two enums in rust but have only one described in swagger 🤔

apps/namadillo/src/atoms/proposals/atoms.ts
mateuszjasiuk: Cool! I will fix that

apps/namadillo/src/hooks/useTransactionCallbacks.tsx
mateuszjasiuk: Hmm I kind of copied it mindlessly from the example above :D but I think you are right. I think in this case we can just remove it, I think there is 0% chance we get proposal status right away. I will double check this

emccorson (approved):

PR #1050, opened 2024-08-27
fix: don't flash lock screen when opening extension

### LINKED ISSUE #968: Extension: Saved login prompt intermittently appears in Firefox when extension unlocked
When opening the extension in Firefox with the extension unlocked, a popup appears with the saved logins for the extension, but there is no need to input a password because the extension is already unlocked.

There is a flash of the lock screen when opening the extension, so I'm guessing the problem is that the password input of the lock screen is briefly rendered causing Firefox to show the saved logins.

This is an intermittent problem and probably does not affect that many users, but is very annoying if you are affected :smile: 

2024-08-07-103514_2880x1800_scrot

emccorson:
This fixes a Firefox bug where the saved password prompt pops up when opening the unlocked extension.

Fixes #968.

### DIFF
--- a/apps/extension/src/App/App.tsx
+++ b/apps/extension/src/App/App.tsx
@@ -9,36 +9,42 @@ import routes from "./routes";
 export const App: React.FC = () => {
   const location = useLocation();
 
-  const { isLocked, unlock, passwordInitialized } = useVaultContext();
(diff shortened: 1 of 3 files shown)


euharrison (approved):

jurevans (approved):
Nice, LGTM!

PR #1051, opened 2024-08-27
feat: add the switch account button

euharrison:
Add the switch account button


apps/extension/src/background/keyring/keyring.ts
```diff
@@ -408,6 +408,22 @@ export class KeyRing {
...
+  public async updateDefaultAccount(address: string): Promise<void> {
+    const accounts = await this.queryAllAccounts();
+    const account = accounts.find((acc) => acc.address === address);
+    if (!account) {
+      throw new Error(`Account with address ${address} not found.`);
+    }
+    if (
+      account.type === AccountType.PrivateKey ||
+      account.type === AccountType.ShieldedKeys
+    ) {
```
jurevans, line 420: Not an issue with your PR, but maybe we need to rethink what a "default" account can be. I feel like if a user imports a private key (or in the future derives a new private key), it should be allowed to be selected. The only difference with a private key is that we can't derive additional keys from it without a mnemonic or seed, so it can't be a parent. I can take a look at this later!

jurevans (approved):
Code looks great! Everything worked perfectly

apps/extension/src/background/keyring/keyring.ts
```diff
@@ -408,6 +408,22 @@ export class KeyRing {
...
   }
 
+  public async updateDefaultAccount(address: string): Promise<void> {
+    const accounts = await this.queryAllAccounts();
+    const account = accounts.find((acc) => acc.address === address);
+    if (!account) {
+      throw new Error(`Account with address ${address} not found.`);
+    }
+    if (
+      account.type === AccountType.PrivateKey ||
```
jurevans, line 418: I think we can remove this - I imported a Private key, so no available mnemonic, and I was able to use this as my "default" account, however, I was unable to later "switch" to it after setting something else.

PR #1056, opened 2024-08-29
feat/1055 - Faucet settings

### LINKED ISSUE #1055: Faucet app: Discuss/Implement configurable API settings
It would be very useful to update the faucet app (currently at `https://faucet.heliax.click`) to have a configurable faucet API url setting. We could also add a token limit (defaults to `1000`, which is generally the case, but both of these are currently configurable at build time)

- Add settings form to the Faucet app for build-time env vars
- Deploy new version after merge, default to public testnet

jurevans:
Resolves #1055 

- Added settings modal to configure Faucet API url with basic persistence (localStorage)
- Restructured a bit
- Faucet Limit is now determined by settings (from API)
- Updated components so that forms submit on Enter key

### Testing

- Build and load the faucet app as per usual, in the top right, click the gear icon
- Enter a new valid URL, then submit
- If this new URL cannot load settings, you will see an error. Setting this to a valid API url will reload the settings and clear the error
- Reload page to ensure this was persisted, or check the local storage key for `baseUrl`


mateuszjasiuk (approved):
Cool! Tested and it works for me :)

PR #1057, opened 2024-08-30
feat: add vote type to voted label

### LINKED ISSUE #1054: Voting signifier change
Currently when a user votes we have a generic blue VOTED symbol

Screenshot 2024-08-29 at 10 59 07

----

----

We would like to change this to be specific to what a user votes (See below text change and color system
Screenshot 2024-08-29 at 10 56 13

----

----

Also needs adjusting here in the main governance overview
Screenshot 2024-08-29 at 10 58 05
Screenshot 2024-08-29 at 10 58 08

emccorson:
Closes #1054.

Example of the different labels:
2024-08-30-113451_2508x1069_scrot

### DIFF
--- a/apps/namadillo/src/App/Governance/ProposalLabels.tsx
+++ b/apps/namadillo/src/App/Governance/ProposalLabels.tsx
@@ -1,7 +1,7 @@
 import { InsetLabel, RoundedLabel } from "@namada/components";
-import { ProposalStatus, ProposalType } from "@namada/types";
+import { ProposalStatus, ProposalType, VoteType } from "@namada/types";
 import { assertNever } from "@namada/utils";
-import { GoCheckCircleFill } from "react-icons/go";
+import { GoCheckCircleFill, GoSkipFill, GoXCircleFill } from "react-icons/go";
 import { twMerge } from "tailwind-merge";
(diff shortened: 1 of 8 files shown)


apps/namadillo/src/App/Governance/ProposalLabels.tsx
```diff
@@ -24,22 +24,35 @@ export const StatusLabel: React.FC<
...
   ...props
 }) => {
+  const { icon, label, color } =
+    vote === "yay" ?
+      { icon: <GoCheckCircleFill />, label: "yes", color: "text-yay" }
+    : vote === "nay" ?
+      { icon: <GoXCircleFill />, label: "no", color: "text-nay" }
+    : vote === "abstain" ?
+      {
+        icon: <GoSkipFill className="rotate-45" />,
```
emccorson, line 43: I couldn't see a suitable icon for this so I used rotate, but if there is a better way that would be good to know.

mateuszjasiuk (approved):
LGTM! Tested and it works nicely!
There is one thing not connected to this PR.
Could you enable polling for `proposalVoteFamily`? I've overlooked it :| 
This way button will change after vote

jurevans (approved):
LGTM!

PR #1064, opened 2024-09-04
fix: Extension - Remove unused Chrome permission

jurevans:
Removes the `offscreen` permission, this caused our extension to be rejected.


mateuszjasiuk (approved):
LGTM! I don't think it will be needed anymore anyways :)

PR #1065, opened 2024-09-04
feat/1061: Firefox - Move durability warning out of main App view

### LINKED ISSUE #1061: Extension: Firefox extension should show "durability" warning elsewhere
We need to move the warning users see when the IndexedDB experimental mode is not enabled

- Move this warning to an internal view, perhaps an `About` view
- We may want to go ahead and display the extension's version here, for Chrome & Firefox

jurevans:
Resolves #1061 

This PR moves the indexedDB durability warning out of the main view, and fixes the width. What I did is add a alert icon to the Settings menu, which will appear in Firefox if the DB is determined to not be durable. This navigates to a general `Warnings` page that shows any warnings (in this case, just the indexed DB warning)

**NOTE** I also applied some similar styles to fix width issues in certain views. Checked these again in Chrome to ensure that it didn't introduce any issues.

- [x] Add warnings view to display durability warning if browser IndexedDB storage durability cannot be guaranteed
- [x] Fix styles issues for Firefox
- [x] Fix bug on Setup tab launch, where empty popup is displayed
- [x] Remove `Connect Hardware Wallet` if USB transport is unavailable (Firefox) 

### Testing

- Install add-on in Firefox, then navigate to `about:config`
- Set `dom.indexedDB.experimental` to `false` (if it isn't already), and the alert icon should be available in the Settings menu
- Set `dom.indexedDB.experimental` to `true`, then this alert will go away

### Screenshots

Menu when store is _not_ durable (Note the triangle with `!` in the top left)
Screen Shot 2024-09-04 at 11 08 28 AM

Warnings page, with fixed width of warning message (preserves normal layout):
Screen Shot 2024-09-04 at 10 42 42 AM

Menu when store is durable:
Screen Shot 2024-09-04 at 10 43 03 AM

### DIFF
--- a/apps/extension/src/App/Accounts/DeleteAccount.tsx
+++ b/apps/extension/src/App/Accounts/DeleteAccount.tsx
@@ -94,7 +94,7 @@ export const DeleteAccount = (): JSX.Element => {
       >
         <Stack className="flex-1 justify-center" gap={4} full>
           <Stack as="header" gap={4}>
-            <Alert type="warning" title="Alert!">
+            <Alert type="warning" title="Alert!" className="max-w-84">
               <p className="mb-4">
                 Make sure that you&apos;ve backed up your seed phrase and
                 private key.
@@ -108,7 +108,7 @@ export const DeleteAccount = (): JSX.Element => {
               </LinkButton>
             </Alert>
           </Stack>
-          <p className="text-white leading-5 text-sm px-2 font-medium mb-10">
+          <p className="text-white leading-5 text-sm px-2 font-medium mb-10 max-w-84">
             After deletion, you will be required to import your seed phrase to
             restore your access to it
           </p>
--- a/apps/extension/src/App/App.tsx
+++ b/apps/extension/src/App/App.tsx
@@ -1,13 +1,36 @@
 import { Container } from "@namada/components";
 import { useVaultContext } from "context/VaultContext";
+import { useRequester } from "hooks/useRequester";
+import { CheckDurabilityMsg } from "provider";
+import { useEffect, useState } from "react";
 import { matchPath, useLocation } from "react-router-dom";
+import { Ports } from "router";
 import { AppContent } from "./AppContent";
 import { AppHeader } from "./Common/AppHeader";
 import { Login } from "./Login";
 import routes from "./routes";
 
+const STORE_DURABILITY_INFO =
+  'Store is not durable. This might cause problems when persisting data on disk.\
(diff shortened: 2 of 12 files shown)


apps/extension/src/App/Accounts/DeleteAccount.tsx
```diff
@@ -94,7 +94,7 @@ export const DeleteAccount = (): JSX.Element => {
       >
         <Stack className="flex-1 justify-center" gap={4} full>
           <Stack as="header" gap={4}>
-            <Alert type="warning" title="Alert!">
+            <Alert type="warning" title="Alert!" className="max-w-84">
```
pedrorezende, line 97: @jurevans could you explain why this `max-w-84` was necessary in these updates?

apps/extension/src/App/App.tsx
```diff
@@ -1,14 +1,37 @@
...
 import routes from "./routes";
 
+const STORE_DURABILITY_INFO =
+  'Store is not durable. This might cause problems when persisting data on disk.\
+ To fix this issue, please navigate to "about:config" and set "dom.indexedDB.experimental" to true.';
+
 export const App: React.FC = () => {
   const location = useLocation();
 
+  const [warnings, setWarnings] = useState<string[]>([]);
```
pedrorezende, line 20: I think we could leave it here for now, but maybe in the future it would be good if we move this to the store, so other parts of the extension could add a warning to the list (but let's not make this premature refactor :P)

pedrorezende (approved):
LGTM! Just added a small comment with a question o/

apps/extension/src/App/Accounts/DeleteAccount.tsx
jurevans: Yes! It was causing the layout to completely break for these specific components. I'm not sure why it doesn't affect Chrome, but the popup would get very wide without it. This value works for both Chrome & FF

apps/extension/src/App/App.tsx
jurevans: I think that makes sense! I was thinking it would be a good idea to approach all of the apps in a similar way, we could just use a Jotai store here. I definitely didn't want to start a refactor for this release

apps/extension/src/App/Accounts/DeleteAccount.tsx
pedrorezende: Oh gotcha! Thanks :)

PR #1066, opened 2024-09-05
feat/1060 - Faucet: Move computePowSolution into web worker

### LINKED ISSUE #1060: Faucet App: Delegate PoW computation to web worker
Currently, when the faucet API `difficulty` setting is set higher than 1, it causes the tab to hang for potentially minutes while it performs this computation.

- Delegate this to a web-worker to unblock UI
- Update UI to listen for completion
  - Tab should not hang, but user should not be allowed to perform any action while this is taking place

jurevans:
Resolves #1060 

When the Faucet API returns a high `difficulty` (e.g., 3) the `computePowSolution` method causes the tab to hang, which can issue warnings to the user. This PR moves that computation into a web-worker.

### Testing

- Load the faucet app and connect to extension
- Ensure that on submit, it displays `Computing POW Solution`, followed by `Processing Faucet Transfer` (Not sure if there's a better way to inspect what is going on while computing

**NOTE**

Testing against our API should finish very quickly, as `difficulty` is set at zero: https://faucet.public.heliax.work/internal-devnet-431.4e30980eee

Testing against the Housefire API will likely take several minutes, with the `difficulty` set at 3 (hence this PR): https://api.faucet.knowable.run

### DIFF
--- a/apps/faucet/.env.sample
+++ b/apps/faucet/.env.sample
@@ -1,9 +1,8 @@
-# Faucet API Endpoint override
+# Faucet API URL Default
 NAMADA_INTERFACE_FAUCET_API_URL=http://127.0.0.1:5000
-NAMADA_INTERFACE_FAUCET_API_ENDPOINT=/api/v1/faucet
 
-# Faucet limit, as defined in genesis toml
-NAMADA_INTERFACE_FAUCET_LIMIT=1000
+# Override the endpoint
+NAMADA_INTERFACE_FAUCET_API_ENDPOINT=/api/v1/faucet
 
 # Specify a proxy port, if proxy is needed (defaults to 9000)
 NAMADA_INTERFACE_PROXY_PORT=1234
--- a/.github/workflows/release-faucet.yml
+++ b/.github/workflows/release-faucet.yml
@@ -14,10 +14,6 @@ on:
         required: true
         type: string
         default: "/api/v1/faucet"
-      NAMADA_INTERFACE_FAUCET_LIMIT:
-        required: true
-        type: string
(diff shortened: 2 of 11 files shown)


apps/faucet/.env.sample
```diff
@@ -1,9 +1,8 @@
-# Faucet API Endpoint override
+# Faucet API URL Default
 NAMADA_INTERFACE_FAUCET_API_URL=http://127.0.0.1:5000
-NAMADA_INTERFACE_FAUCET_API_ENDPOINT=/api/v1/faucet
 
-# Faucet limit, as defined in genesis toml
-NAMADA_INTERFACE_FAUCET_LIMIT=1000
```
euharrison, line 6: can you remove this env var also from `apps/faucet/README.md` and `release-faucet.yml`

euharrison (approved):
Nice!

Personally I feel more complex to read code with a lot of `useEffect` that could be async operations

How about if we implement something similar to this to have all executions on the `async handleSubmit` function removing the `useEffect`s?

```tsx
const asyncPow = (args: PowChallenge) =>
  new Promise((resolve) => {
    powSolver.onmessage = ({ data }) => {
      resolve(data);
      powSolver.onmessage = null
    };
    powSolver.postMessage(args);
  });
```

apps/faucet/.env.sample
jurevans: Thanks for catching that!

jurevans:
> How about if we implement something similar to this to have all executions on the `async handleSubmit` function removing the `useEffect`s?

I've updated this. Thanks!

PR #1067, opened 2024-09-07
Chore: bumping packages and setting up React-Testing-Library correctly for Namadillo

pedrorezende:
This PR initially aimed to implement changes in the Staking module and introduce automated integration tests for several components. However, the existing Jest + React Testing Library configuration had issues, particularly with module mapping. To address this, I not only began setting up the tests but also upgraded several libraries to their latest versions.

The following list shows the most critical upgrades on Namadillo / Extension:

- React: 18.00 -> 18.3.0
- React-Router-Dom: 6.0.0 -> 6.26.1
- Typescript: 5.1.0 -> 5.5.4
- Vite: 5.2.11 -> 5.4.3
- Framer-motion (Extension only): 6.2.4 -> 11.3.28
- Webpack (Namadillo only): removed

A few changes were required to some components to address some new TypeScript errors fired due to these updates.

Also some devDependencies used by different apps/packages were moved to the root of the repository in order to make it easier to upgrade them in the future.


apps/extension/src/storage/VaultStorage.ts
```diff
@@ -12,7 +12,7 @@ const Uint8ArrayCodec = new t.Type<Uint8Array, Uint8Array, unknown>(
     try {
       const a = Array.from(u as Uint8Array);
       return t.success(Uint8Array.from(a));
-    } catch (e) {
+    } catch (_e) {
```
jurevans, line 15: Could we just leave this empty?

pedrorezende: Will do that for other occurrences as well

jurevans (approved):
This looks great! I did a round of testing and all is well :)

PR #1068, opened 2024-09-08
Namadillo: fixing message appearing due to wrong string cast

pedrorezende:
This PR fixes an error introduced in #1067 where Advanced panel starts to display a wrong "null" string.


emccorson (approved):
I think this looks good! My only concern would be if an error gets thrown that isn't an instance of Error, it wouldn't be shown, but I'm not sure if that can ever happen.

PR #1074, opened 2024-09-09
fix: Faucet - Use relative path for assets in Webpack config

jurevans:
Fixes an issue where the asset URL was set to `https://faucet.heliax.click`. Note that this has introduced a 404 on the `apple-touch-icon.png`, I will see if I can resolve that error, but it's not critical

### TODO

Update this app to Vite!


mateuszjasiuk (approved):
LGTM! :)

PR #1083, opened 2024-09-09
fix: Faucet - Prevent default submit behavior

jurevans:
Resolves issue that occurs on some Firefox browsers (in this case Linux), where the default behavior of the submit button reloads the page.

### Testing

- Either run locally, or run the deployed faucet at https://faucet.heliax.click
- Perform faucet transfer


emccorson (approved):
Nice!

PR #1084, opened 2024-09-09
feat./1081 - Namadillo: Check Tx applied status when broadcasting

### LINKED ISSUE #1081: Namadillo/Indexer: Notifications should indicate Tx applied
Currently, we display a Success notification when the SDK successfully adds a Tx to mempool (valid signature, valid gas)

We should be more accurate in the notifications:

- Pending Notification on _broadcast_
- Mempool validation notification once _accepted_
- Final Success notification once _applied_

jurevans:
Resolves #1081 

Using the response from `rpc.broadcastTx()`, check the `isApplied` value for each commitment, raise error if any are not applied.

- [x] Update SDK to fix issue where we were not getting accurate `isApplied` values if any of the innerTx were not applied
- [x] Export SDK function to return the computed inner Tx hashes of all commitments (this is used later for mapping to inner Tx props)
- [x] Update Toast & Notifications in Namadillo to collect successful/failed Tx and display a new notification to users, `PartialSuccess`, where both successful and failed inner-Tx can be displayed

This makes use of the `ProcessTxResponse` from the SDK's `rpc.broadcastTx()`. Example:
```json
{
    "code": "0",
    "commitments": [
        {
            "hash": "DE979B02CC9014068E09DEFCB26592FB2A992D43B35B57C1CC9643BAE4F392CE",
            "isApplied": true
        },
        {
            "hash": "C91EED4782248D482517AE5C37AEA9895E1469439C42C059789C334FB9F8031C",
            "isApplied": true
        }
    ],
    "gasUsed": "210",
    "hash": "BC99E787272BDD15468E3F5C86EEC2FBAE752C568983086805EDB5286F5BDB95",
    "height": "16602",
    "info": "Check batch for result.",
    "log": ""
}
```

Each commitment in this response has a boolean indicating whether that commitment has been accepted by all VPs and applied on chain.

### Testing

Testing this PR is a bit weird, as you need to trigger a VP rejection _after_ SDK validation passes. An easy way to do this is to hijack the `gasLimit` value. In Namadillo, in `src/lib/query.ts` -> `getTxProps`:
- use (for example)`gasLimit: BigNumber(100)`, then submit 3 staking Tx - all should fail
- Set this value to `BigNumber(285)` and only one will fail.

**NOTE** You can test the results (check the hashes) in the error msg against what is in the console

### Screenshot

Partial success notification, indicates both successful and unsuccessful Tx:
Screen Shot 2024-09-10 at 3 36 34 PM

Failed state when SDK validation passes, but the Tx response indicates that commitments were rejected by VPs:
Screen Shot 2024-09-10 at 4 57 21 PM

Success notification is unchanged.

### DIFF
--- a/apps/namadillo/src/hooks/useTransactionNotifications.tsx
+++ b/apps/namadillo/src/hooks/useTransactionNotifications.tsx
@@ -111,7 +111,7 @@ export const useTransactionNotifications = (): void => {
     clearPendingNotifications(id);
     dispatchNotification({
       id,
-      title: "Some staking tx failed",
+      title: "Some staking transactions failed",
       description: (
         <>
           Your staking transaction of <NamCurrency amount={total} /> has
@@ -141,6 +141,27 @@ export const useTransactionNotifications = (): void => {
     });
   });
 
+  useTransactionEventListener("Unbond.PartialSuccess", (e) => {
+    const { id, total } = parseTxsData(e.detail.tx, e.detail.data);
+    clearPendingNotifications(id);
+    dispatchNotification({
+      id,
+      title: "Some Unstake transactions failed",
+      description: (
+        <>
+          Your unbonding of <NamCurrency amount={total} /> has succeeded
+        </>
+      ),
+      details: getAmountByValidatorList(e.detail.data),
+      failedDetails:
+        e.detail.failedData ?
+          getAmountByValidatorList(e.detail.failedData)
+        : undefined,
+      type: "partialSuccess",
+      forceDetailsOpen: true,
+    });
+  });
+
   useTransactionEventListener("Unbond.Error", (e) => {
     const { id, total } = parseTxsData(e.detail.tx, e.detail.data);
     clearPendingNotifications(id);
@@ -163,7 +184,7 @@ export const useTransactionNotifications = (): void => {
     dispatchNotification({
       id,
       title: "Withdrawal Success",
-      description: `Your withdrawal transaction has succeeded`,
+      description: <>Your withdrawal transaction has succeeded</>,
       type: "success",
     });
   });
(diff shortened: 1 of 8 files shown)


apps/namadillo/src/hooks/useTransactionNotifications.tsx
```diff
@@ -106,6 +107,28 @@ export const useTransactionNotifications = (): void => {
...
+          succeeded
+        </>
+      ),
+      failedDescription: (
+        <>The following staking transactions were not applied:</>
+      ),
+      details: getAmountByValidatorList(e.detail.data),
+      failedDetails: getAmountByValidatorList(e.detail.failedData || []),
+      type: "partialSuccess",
+      forceDetailsOpen: true,
```
euharrison, line 128: an alternative for this is using the `description` field with a component, then removing the `failedDescription`, `failedDetails` and `forceDetailsOpen` ✂️ 

something like this:

```tsx
dispatchNotification({
  ...
  description: <PartialSuccessDescription detail={e.detail} />
})

const PartialSuccessDescription = (detail) => {
  return (
    <>
      <div className="w-full text-xs text-white block">
        <div>{detail.details}</div>
        <div>{detail.failedDescription}</div>
        <div>{detail.failedDetails}</div>
      </div>
    </>
  )
}

**Review by euharrison — APPROVED** [review]
I could not test with the hijack, but the implementation looks good ✅

File: apps/namadillo/src/lib/query.ts [hunk 2]
```diff
@@ -193,14 +206,63 @@ export const broadcastTx = async <T>(
...
+          return commitmentErrors.includes(data.hash);
+        });
+
+        if (successData?.length) {
+          eventType &&
+            window.dispatchEvent(
+              new CustomEvent(`${eventType}.PartialSuccess`, {
+                detail: {
+                  tx,
+                  data: successData,
```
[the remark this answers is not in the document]
jurevans: you're right! I will do that

apps/namadillo/src/hooks/useTransactionNotifications.tsx
jurevans: Nice, I will do this! Thanks!

jurevans: Done!

pedrorezende (approved):

PR #1085, opened 2024-09-09
Namadillo: Adding Staking filters and different states

### LINKED ISSUE #1040: Additions to the stake flow to include different validator states
Alterations needed See figma very top right [here](https://www.figma.com/design/NFyHbLZXBSl3aUsMxtffvV/Namada-End-User-Interface-V1.0-%2F-Phase-1%3A-Block-Party?node-id=6760-511280&t=ftlBgMfGVF7MUcXt-1)

- ALL button is now a dropdown filter / ALL / Active / Jailed / Inactive (ALL is default)
- Remove randomise button
- Add in Validator consensus rank as an extra column. Validators in consensus will rank 1-257. Validators outside of consensus can be display as a –. This includes both below threshold and below capacity
- Staking with a validator outside of consensus should bring up a text prompt advising no staking rewards
- Jailed validators red bars over the number + A text saying jailed validator where the stake input box is - Info is greyed out
- Jailed validator that you have stake with the staked icon turns from cyan to red also. Text prompt -  info is greyed out
- Inactive validator - elements greyed out + text prompt "inactive validator
- Inactive validator you have stake with - greyed out + prompt to unbond

pedrorezende:
This PR adds validator status to staking tables, removes the "randomize" button, and adds different warnings and elements depending on validator status.

Fixes #1040 .

### DIFF
--- a/apps/namadillo/src/App/Common/NamInput.tsx
+++ b/apps/namadillo/src/App/Common/NamInput.tsx
@@ -1,7 +1,7 @@
 import { AmountInput } from "@namada/components";
 import { Tokens } from "@namada/types";
 
-type NamInputProps = Omit<
+export type NamInputProps = Omit<
   React.ComponentProps<typeof AmountInput>,
   "maxDecimalPlaces"
 >;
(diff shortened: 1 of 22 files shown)


jurevans (approved):
LGTM! Works nicely, excellent tests

PR #1089, opened 2024-09-10
feat: Faucet - Make extension connection optional

jurevans:
Users of the faucet app have had issues using it when the accounts they need to fund are generate in the CLI.

This update removes the need to connect the extension to access your accounts, but it is still an option.


euharrison (approved):
LGTM! ✅

euharrison:
Thinking loud... we should increase the contrast of the fields

Screenshot 2024-09-13 at 10 34 00

mateuszjasiuk:
Not sure if it's me only but I found hidden message :)
image

mateuszjasiuk (approved):
LGTM! Works nicely!

jurevans:
> Not sure if it's me only but I found hidden message :)

Good catch! I can fix that

PR #1091, opened 2024-09-11
WIP: feat/1078 - Extension - Enable chain approval

### LINKED ISSUE #1078: Extension: Allow interface to specify chain ID when connecting, update when approved
Right now, we have a simple chain ID field in the Extension settings.

When Namadillo connects to the extension, it could provide a chain ID, which the user should review in the extension for approval, _if_ it would require that the extension needs to sign for a different chain. As we already track chain ID, we can keep this in storage, but allow approvals to change it.

- Extension is released with a "default" chain ID
- When interface connects, it provides the chain ID returned from the indexer
- If this chain Id is different than what is in the extension, the extension approval will notify the user to review
- Upon approval, the chain ID in the extension is updated, and the user won't see the notification until the chain is requested to be changed.

jurevans:
Resolves #1078 

- [x] When approving connection, if interface is attempting to use a different chain ID, alert user, and if approved, save this in extension
- [ ] When the interface is later connected to an indexer that returns a different chain ID, display a new approval to allow updating the extension

### DIFF
--- a/apps/extension/src/Approvals/ApproveConnection.tsx
+++ b/apps/extension/src/Approvals/ApproveConnection.tsx
@@ -1,8 +1,11 @@
 import { ActionButton, Alert, GapPatterns, Stack } from "@namada/components";
+import { Chain } from "@namada/types";
 import { PageHeader } from "App/Common";
 import { ConnectInterfaceResponseMsg } from "background/approvals";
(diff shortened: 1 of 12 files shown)


jurevans:
Closing in favor of https://github.com/anoma/namada-interface/pull/1142

PR #1092, opened 2024-09-11
feat: add namadillo disconnect button

euharrison:
Add the disconnect button on top right that removes the current origin of the Namadillo from the "Connected Sites" on the extensions

https://github.com/user-attachments/assets/941d90af-cbe9-454b-95ea-9c9cf2bfc5f1


jurevans (approved):
LGTM!

PR #1101, opened 2024-09-12
feat: add disconnect account approval step

euharrison:
Add the disconnection account approval step

This PR also combine e create some methods for the Approvals service to manage the popup easier, the main one is `launchApprovalPopup` which takes a typed route and custom params. And it'll remove the popup resolver from the map once they are resolved or rejected, or when the popup is removed/closed by the user

Screenshot 2024-09-12 at 19 38 04

https://github.com/user-attachments/assets/285a3a6f-aef3-4e8a-8c08-ea47e1cdc988

### DIFF
--- a/apps/extension/src/Approvals/ApproveConnection.tsx
+++ b/apps/extension/src/Approvals/ApproveConnection.tsx
@@ -10,17 +10,12 @@ export const ApproveConnection: React.FC = () => {
   const requester = useRequester();
   const params = useQuery();
   const interfaceOrigin = params.get("interfaceOrigin");
-  const interfaceTabId = params.get("interfaceTabId");
 
   const handleResponse = async (allowConnection: boolean): Promise<void> => {
-    if (interfaceTabId && interfaceOrigin) {
+    if (interfaceOrigin) {
       await requester.sendMessage(
         Ports.Background,
-        new ConnectInterfaceResponseMsg(
-          parseInt(interfaceTabId),
-          interfaceOrigin,
-          allowConnection
-        )
+        new ConnectInterfaceResponseMsg(interfaceOrigin, allowConnection)
       );
       await closeCurrentTab();
     }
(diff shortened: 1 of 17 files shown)


apps/extension/src/Approvals/ApproveConnection.tsx
```diff
@@ -10,17 +10,12 @@ export const ApproveConnection: React.FC = () => {
   const requester = useRequester();
   const params = useQuery();
   const interfaceOrigin = params.get("interfaceOrigin");
-  const interfaceTabId = params.get("interfaceTabId");
```
euharrison, line 13: `interfaceTabId` wasn't used anymore, it was replaced by `popupTabId` directly on the service

apps/extension/src/background/approvals/service.ts
```diff
@@ -34,7 +37,15 @@ export class ApprovalsService {
     protected readonly vaultService: VaultService,
     protected readonly chainService: ChainsService,
     protected readonly broadcaster: ExtensionBroadcaster
-  ) {}
+  ) {
+    browser.tabs.onRemoved.addListener((tabId) => {
+      const resolver = this.getResolver(tabId);
+      if (resolver) {
+        resolver.reject(new Error("Window closed"));
+        this.removeResolver(tabId);
```
euharrison, line 45: If the user closes the popup, it'll dispatch a message and clean the `resolverMap`

Screenshot 2024-09-12 at 19 56 37

apps/extension/src/background/approvals/service.ts
```diff
@@ -338,11 +307,54 @@ export class ApprovalsService {
...
+    }
+
+    if (popupTabId in this.resolverMap) {
+      throw new Error(`tab ID ${popupTabId} already exists in promise map`);
+    }
+
     return popupTabId;
   };
+
+  private launchApprovalPopup = async <T>(
```
euharrison, line 325: `launchApprovalPopup` now combines the same pattern we have in all functions: 
- Create the url
- Open the popup
- Return the resolver promise
- Plus: when the promise is resolved or rejected, it's also removed from the `resolverMap`

jurevans (approved):
Very nice!

PR #1102, opened 2024-09-13
fix: trim long strings on extension list items

euharrison:
Trim long strings on extension list items and add the `title` attribute for it as a last user resource

Screenshot 2024-09-13 at 10 54 27

Screenshot 2024-09-13 at 10 59 33


pedrorezende (approved):
LGTM!

PR #1112, opened 2024-09-14
fix: docker nginx config

Fraccaman:


Fraccaman:
internally we let the CDN gzip the request, thats why I switched it off

PR #1099, opened 2024-09-12
feat/1098 - SDK: Add remaining Transfer types

jurevans:
Resolves #1098

Mainly we want to add support to MASP to the SDK & the Interface. Short of implementing the Transfer UI, we should have the following:

Need args and Borsh schema for:
- [x] Shielded Transfer (Shielded<->Shielded)
- [x] Shielding Transfer (Transparent->Shielded)
- [x] Unshielding Transfer (Shielding->Shielded)

Need SDK (`@heliax/namada-sdk`) functions for:

- [x] `build_shielded_transfer`
- [x] `build_shielding_transfer`
- [x] `build_unshielding_transfer`

Docs:
- [x] Regenerate docs for `@namada/types` `@heliax/namada-sdk`

TODO:
- [ ] Load MASP Params into SDK wasm (fetch & store, provide to SDK)

Issue #1098, opened 2024-09-12
SDK package: Hook up remaining Transfer types


jurevans:
Since the build functions for transfer and the transfer types themselves have been split from a single type, we need to add definitions to our SDK package to support these.

- Hook up Shielding transfer function
- Hook up Unshielding transfer function
- Ensure we can serialize/deserialize these (there may be some things to do in the `Transfer` type, but looks like it's complete as is)
- MASP Params will need to be loaded to fully test
- Generate updated documentation

PR #1114, opened 2024-09-14
POC/WIP: feat/1098 - Fetch and store MASP params in background

jurevans:
Relates to #1098  & #1099 

Namadillo will need to fetch and store MASP params once. This PR is to experiment with fetching these within a service worker and storing these in IndexedDB.

To move this PR out of POC status, we need a smoother development & build integration. Currently, packages cannot be imported into this context (specifically, @namada/storage which is the IndexedDB interface we use). For this branch, I am just using a copied version for now. We will potentially be able to improve the Vite pipeline & dev process for working with SW.

Also, to avoid CORS issues, I've had to proxy while working locally. This is not ideal, but is an issue regardless of whether we request in the App vs SW.


jurevans:
Close, as not needed

PR #1115, opened 2024-09-14
Namadillo: Optimistically updating staking rewards balance

pedrorezende:
When you claim rewards, the indexer will need to wait for the next block to sync the new amount. This process might take a few seconds, depending on network configuration. This PR sets the available rewards balance to zero as soon as the claim rewards success event gets fired.

### DIFF
--- a/apps/namadillo/src/atoms/staking/services.ts
+++ b/apps/namadillo/src/atoms/staking/services.ts
@@ -12,6 +12,7 @@ import {
   WithdrawProps,
   WrapperTxProps,
 } from "@namada/types";
+import { queryClient } from "App/Common/QueryProvider";
 import { getSdkInstance } from "hooks";
 import { TransactionPair, buildTxPair } from "lib/query";
 import { Address, AddressBalance, ChainSettings, GasConfig } from "types";
@@ -158,4 +159,8 @@ export const createClaimAndStakeTx = async (
     buildClaimRewardsAndStake,
     account.address
   );
+};
+
+export const clearClaimRewards = (accountAddress: string): void => {
+  queryClient.setQueryData(["claim-rewards", accountAddress], () => ({}));
 };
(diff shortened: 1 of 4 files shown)


apps/namadillo/src/hooks/useTransactionCallbacks.tsx
```diff
@@ -1,29 +1,36 @@
...
+    }
   };
 
   useTransactionEventListener("Bond.Success", onBalanceUpdate);
   useTransactionEventListener("Unbond.Success", onBalanceUpdate);
   useTransactionEventListener("Withdraw.Success", onBalanceUpdate);
   useTransactionEventListener("Redelegate.Success", onBalanceUpdate);
-  useTransactionEventListener("ClaimRewards.Success", onBalanceUpdate);
+  useTransactionEventListener("ClaimRewards.Success", onBalanceUpdate, [
+    account,
```
euharrison, line 32: nice idea! I'm asking myself if this this object changes for whatever reason and if we should use `account?.address` instead

apps/namadillo/src/atoms/staking/services.ts
```diff
@@ -159,3 +160,7 @@ export const createClaimAndStakeTx = async (
     account.address
   );
 };
+
+export const clearClaimRewards = (accountAddress: string): void => {
+  queryClient.setQueryData(["claim-rewards", accountAddress], () => ({}));
```
euharrison, line 165: Did you try this as `invalidateQueries` https://tanstack.com/query/latest/docs/framework/react/guides/query-invalidation ? I think this is the "correct way" to clear queries for react-query

euharrison (approved):
Nice!

apps/namadillo/src/atoms/staking/services.ts
pedrorezende: Actually not. I'm overriding the result of the claim-rewards query, so no rewards are available :) . But maybe I can be more explicit on this. Will make a small change before merging

PR #1119, opened 2024-09-16
Namadillo: first sketch on Transfer module

pedrorezende:
This PR contains the skeleton of the first components required by the transfer module. This is still a work in progress, but I decided to open this PR for review to discuss if I'm going in the right direction, and to not let it become so big. The next steps will contain the implementation of these components, integrating them with the actual api.


apps/namadillo/src/App/Transfer/TransferDestination.tsx
```diff
@@ -0,0 +1,106 @@
...
+  onToggleCustomAddress,
+  address,
+  onChangeAddress,
+  memo,
+  onChangeMemo,
+}: TransferDestinationProps): JSX.Element => {
+  return (
+    <div
+      className={twMerge(
+        clsx("relative bg-neutral-800 rounded-lg px-4 py-5", {
```
euharrison, line 57: should we use `twMerge`?

apps/namadillo/src/App/Transfer/SelectAssetsModal.tsx
euharrison, line 1: is this an unused file?

apps/namadillo/src/App/Transfer/IBCToNamadaModule.tsx
euharrison, line 1: only to confirm that you plan to fill this file up with some code :) 

maybe you could create all the options into a big page so we can see and test them all together, only while they are under development, wdyt?

euharrison (approved):
LGTM

apps/namadillo/src/App/Transfer/IBCToNamadaModule.tsx
pedrorezende: The problem is that we're gonna have different configs to customize the modal behavior. But I'll start with IBC in a different page. Removing this file for now!

apps/namadillo/src/App/Transfer/SelectAssetsModal.tsx
pedrorezende: Not for too long :) I'm keeping this one because I'll be implementing its behavior in a next PR this week

PR #1120, opened 2024-09-17
MASP indexer integration

mateuszjasiuk:
I don't think there is an easy way to test this for now :/ But I think as long as this does not break other features we can merge it.

Okay!
**Make sure you test against 0.44.0**
So to test shielded sync w/o masp indexer you can:
- shield some nam using CLI(localnet example)
```sh
target/debug/namada wallet gen --shielded --alias albert-shielded --base-dir .namada/validator-0
target/debug/namada wallet gen-payment-addr --alias albert-pa1 --key albert-shielded --base-dir .namada/validator-0
target/debug/namada client shield --source albert --target albert-pa1 --token nam --amount 1000  --base-dir .namada/validator-0
# To get the viewing key
target/debug/namada wallet find --alias albert-shielded --base-dir .namada/validator-0
```

- Add this to `App.tsx` in namadillo(with missing imports)
```ts
  useEffect(() => {
    const shieldedSync = async (): Promise<void> => {
      const vk =
        "YOUR_VK";
      const sdk = await getSdkInstance();
      await sdk.rpc.shieldedSync([vk]);
    };

    // eslint-disable-next-line @typescript-eslint/no-explicit-any
    (window as any).shieldedSync = shieldedSync;
  });
```
- run `await window.shieldedSync()` in namadillo console, you should see progress logged
- :eyes: 

image

To test with masp indexer you can do the same setup but also run masp_indexer locally.
- running [instructions ](https://github.com/anoma/namada-masp-indexer?tab=readme-ov-file#how-to-run)
- if you want to run masp indexer alongsinde indexer(it's a bit of a headache for now) you need to change ports :/ I will change ports in regular indexer soon so there are no conflicts during development
```yml
    command: ["postgres", "-c", "listen_addresses=0.0.0.0", "-c", "max_connections=200", "-p", "5433"]
    expose:
      - "5433" # Publishes 5432 to other containers but NOT to host machine
    ports:
      - "5433:5433"
# AND
    ports:
      - 5000:5001
 # AND ALL
      DATABASE_URL: postgres://postgres:password@postgres:5433/masp_indexer_local
```
- paste MASP indexer url
image
- wait for commitment tree to get indexed
- retest

❗**CLEAR CONTEXT MANUALLY BETWEEN TESTS, JUST REMOVE IT FROM IDB** ❗
ALSO currently I;m getting this :(
```
Caused by:
    Could not deserialize the notes map JSON response at height 3017: error decoding response body: missing field `batch_index` at line 1 column 89
```

### DIFF
--- a/packages/crypto/lib/src/crypto/zip32.rs
+++ b/packages/crypto/lib/src/crypto/zip32.rs
@@ -124,11 +124,11 @@ mod tests {
             .expect("Deriving from ExtendedKeys should not fail");
 
         let payment_address: PaymentAddress =
-            borsh::BorshDeserialize::try_from_slice(&payment_address)
+            borsh::BorshDeserialize::try_from_slice(payment_address)
                 .expect("Should be able to deserialize payment address!");
-        let xsk: ExtendedSpendingKey = borsh::BorshDeserialize::try_from_slice(&xsk)
+        let xsk: ExtendedSpendingKey = borsh::BorshDeserialize::try_from_slice(xsk)
             .expect("Should be able to deserialize extended spending key!");
-        let xfvk: ExtendedFullViewingKey = borsh::BorshDeserialize::try_from_slice(&xfvk)
+        let xfvk: ExtendedFullViewingKey = borsh::BorshDeserialize::try_from_slice(xfvk)
             .expect("Should be able to deserialize full viewing key!");
 
         assert_eq!(payment_address.to_bytes().len(), 43);
--- a/packages/shared/lib/src/sdk/args.rs
+++ b/packages/shared/lib/src/sdk/args.rs
@@ -3,7 +3,6 @@ use std::{path::PathBuf, str::FromStr};
 use namada_sdk::borsh::{BorshDeserialize, BorshSerialize};
 use namada_sdk::ibc::core::host::types::identifiers::{ChannelId, PortId};
 use namada_sdk::ibc::IbcShieldingData;
-use namada_sdk::masp::{ExtendedSpendingKey, PaymentAddress};
 use namada_sdk::tendermint_rpc;
 use namada_sdk::tx::data::GasLimit;
 use namada_sdk::{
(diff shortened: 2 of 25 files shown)


packages/shared/lib/src/sdk/mod.rs
```diff
@@ -252,13 +251,13 @@ impl Sdk {
...
         for bytes in built_txs_bytes.iter() {
-            let tx: tx::Tx = borsh::from_slice(&bytes)?;
+            let tx: tx::Tx = borsh::from_slice(bytes)?;
             built_txs.push(tx);
         }
 
         // Get wrapper args
         let first_tx = built_txs
-            .get(0)
+            .first()
```
jurevans, line 260: thanks, I thought I had updated all of these :D

packages/shared/lib/src/sdk/args.rs
```diff
@@ -761,6 +765,8 @@ pub fn ibc_transfer_tx_args(
         channel_id,
         timeout_height,
         timeout_sec_offset,
+        // TODO: false for now
+        disposable_signing_key: false,
```
jurevans, line 769: What do you think about mapping this to the types/args before we release? I could do a follow-up PR that adds any potential args we may want soon to our types package, that way we wouldn't have to worry about any breaking APIs after we publish. Unless there's not a use-case we'll need for this. I can look into that!

mateuszjasiuk: I think we can add it when we start testing transfers

packages/crypto/lib/src/crypto/zip32.rs
```diff
@@ -124,11 +124,11 @@ mod tests {
             .expect("Deriving from ExtendedKeys should not fail");
```
jurevans, line 125: Weird, I'm now getting a build warning on this file, though not sure why it's happening on this branch:
```
warning: field `0` is never read
  --> src/crypto/zip32.rs:15:27
   |
15 | pub struct ExtSpendingKey(pub(crate) ExtendedSpendingKey);
   |            -------------- ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
   |            |
   |            field in this struct
```
Not sure that we're using that type any more. Not a big deal on this PR, just noting!

jurevans (approved):
Code LGTM! Was able to test this locally, works great! I think it's fine to point to that `rev` for now, we can later point to the next tag that has the fixes we need.

PR #1122, opened 2024-09-18
fix: fix vote button being incorrectly disabled

emccorson:
This was caused by returning undefined from react query to signal that the user has no existing vote, but returning undefined is disallowed by react query.


jurevans (approved):
LGTM! Tested locally, was able to vote!

PR #1124, opened 2024-09-18
WIP: feat/1098 - Fetch masp params in Namadillo

### LINKED ISSUE #1098: SDK package: Hook up remaining Transfer types

jurevans:
Resolves #1098 

To support building shielded Tx, we need to provide MASP params. This PR hooks up fetching and storing of params in Namadillo.

### Notes

- I move `SessionKVStore` into `extension/src/extension` as it uses the `webextension-polyfill`, meaning that package couldn't be used outside of the extension. Moving it here allows us to re-use `IndexedDBKVStore` in Namadillo
- In [1114](https://github.com/anoma/namada-interface/pull/1114) I did some experimenting with fetching these params in a service worker, however, to do this the right way, I'll need to set up a better dev/build pipeline with `rollup.js`/`vite` so that we can use modules during development. For now, I will incorporate this in Namadillo, then follow-up with delegating this to a background worker


jurevans:
Closing, this was implemented in another PR

PR #1127, opened 2024-09-18
Namadillo: removing unused .env injection on Vite config

pedrorezende:
Environment variables are not used anymore in Namadillo. Recently, due to some package updates we've introduced a very unlikely error entitled as `Uncaught TypeError: process.version is undefined` . This PR removes this unused configuration and also fixes this error.

### DIFF
--- a/apps/namadillo/vite.config.mjs
+++ b/apps/namadillo/vite.config.mjs
@@ -1,22 +1,11 @@
 /* eslint-disable */
 import react from "@vitejs/plugin-react";
-import { defineConfig, loadEnv } from "vite";
+import { defineConfig } from "vite";
 import checker from "vite-plugin-checker";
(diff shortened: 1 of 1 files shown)


apps/namadillo/vite.config.mjs
```diff
@@ -30,9 +19,6 @@ export default defineConfig(({ mode }) => {
         overlay: { initialIsOpen: false },
       }),
     ],
-    define: {
-      "process.env": filteredEnv,
```
emccorson, line 34: I had no idea this line overrides the whole `process` object :smile:

emccorson (approved):
Looks good to me!

The bug should already have been fixed by #1121, but I think it still makes sense to merge this.

`process.env` appears to still be present in the default chain configs in the chains package, but that's not being used anywhere as far as I can see.

jurevans (approved):

PR #1129, opened 2024-09-19
packages: Additional tests around shielded keys

jurevans:
Someone on Discord has reported that when testing import of the same mnemonic in the extension, sometimes their Payment Address is different. I'm not sure how this could be possible, and we have existing tests that ensure that the keys must match for any given mnemonic.

So, I propose:
- Add additional tests to the `@namada/crypto` lib (Rust & Jest)
  - Imported mnemonic generates the same seed
  - Payment Address & Viewing Key match expected when deriving from seed
- Add additional tests to `@heliax/namada-sdk` package (we currently have one test mnemonic, will add more)
- Ensure these tests are being run in CI each time

### Testing

- `packages/crypto`
  - `yarn test-wasm:ci`
  - `yarn test`
- `packages/sdk`
  - `yarn test`

### DIFF
--- a/packages/crypto/lib/src/crypto/bip39.rs
+++ b/packages/crypto/lib/src/crypto/bip39.rs
@@ -115,6 +115,24 @@ mod tests {
     }
 
     #[wasm_bindgen_test]
+    fn can_restore_seed_from_phrase() {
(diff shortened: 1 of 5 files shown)


mateuszjasiuk (approved):
LGTM!

PR #1131, opened 2024-09-19
Transfer Module - Part 2

pedrorezende:
This PR contains another set of improvements for the TransferModule component. Now it should be able to support all the UI required to implement transfers from IBC chains to Namada

This component is still in development, but it now can be used to start to integrate with the actual IBC transfer flow.

A rough example on a mock implementation can be found in App/Transfer/Example.tsx component, which is accessible by `/transfer/example` route.


euharrison (approved):
Niiiice!

PR #1133, opened 2024-09-20
feat: add rough form for IBC withdrawals

emccorson:
This PR adds another very rough IBC prototype, this time for IBC transfers from Namada to Cosmos.

Note: The tx details are not dislaying properly in the extension, but as far as I know we need to get the IBC WASM hash from the indexer to do this, which it is currently not sending. anoma/namada-indexer#120

## Testing

1. Set up hermes as in https://hackmd.io/a95Q2m6TRM-ntjFSSPzKTA (make sure to use `theta-testnet-001` as the Cosmos testnet)
2. Send some NAM to using the form
3. Check you received the NAM on the Cosmos side with `gaiad query bank balances <cosmos address> --node https://rpc-t.cosmos.nodestake.top/`

### DIFF
--- a/apps/namadillo/src/App/Ibc/IbcWithdraw.tsx
+++ b/apps/namadillo/src/App/Ibc/IbcWithdraw.tsx
@@ -1,9 +1,160 @@
+import { Key as KeplrKey, Window as KeplrWindow } from "@keplr-wallet/types";
 import { Panel } from "@namada/components";
+import { WindowWithNamada } from "@namada/types";
+import BigNumber from "bignumber.js";
+import { getSdkInstance } from "hooks";
+import { useAtomValue } from "jotai";
+import { useEffect, useState } from "react";
+
+import { NamCurrency } from "App/Common/NamCurrency";
+import { NamInput } from "App/Common/NamInput";
+import { accountBalanceAtom, defaultAccountAtom } from "atoms/accounts";
+import { chainAtom, chainParametersAtom } from "atoms/chain";
+import { getFirstError } from "atoms/utils";
+
+const keplr = (window as KeplrWindow).keplr!;
+const namada = (window as WindowWithNamada).namada!;
+
+const keplrChain = "theta-testnet-001";
+
+const buttonStyles = "bg-white my-2 p-2 block";
 
 export const IbcWithdraw: React.FC = () => {
+  const [error, setError] = useState("");
+  const [keplrAccount, setKeplrAccount] = useState<KeplrKey>();
+  const [amount, setAmount] = useState<BigNumber>();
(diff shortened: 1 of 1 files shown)


mateuszjasiuk:
As a side note. I've just [added ](https://github.com/anoma/namada-indexer/pull/121)tx_ibc checksums to the indexer.

apps/namadillo/src/App/Ibc/IbcFromNamada.tsx
```diff
@@ -0,0 +1,162 @@
...
+import { useEffect, useState } from "react";
+
+import { NamCurrency } from "App/Common/NamCurrency";
+import { NamInput } from "App/Common/NamInput";
+import { accountBalanceAtom, defaultAccountAtom } from "atoms/accounts";
+import { chainAtom, chainParametersAtom } from "atoms/chain";
+import { getFirstError } from "atoms/utils";
+
+const keplr = (window as KeplrWindow).keplr!;
+const namada = (window as WindowWithNamada).namada!;
```
mateuszjasiuk, line 16: For testing I had to move this inside `submitIbcTransfer` otherwise it was sometimes undefined :P

mateuszjasiuk (approved):
I did not test it until the end because of the "Per-epoch throughput limit", but I'm pretty sure it works fine. I also got an error when channel became inactive so I think all is good 👍

apps/namadillo/src/App/Ibc/IbcFromNamada.tsx
```diff
@@ -0,0 +1,162 @@
...
+    const tx = await sdk.tx.buildIbcTransfer(wrapperTxProps, {
+      source: namadaAccount.data!.address,
+      receiver: keplrAccount!.bech32Address,
+      token: namadaChain.data!.nativeTokenAddress,
+      amount: amount!,
+      portId: "transfer",
+      channelId: channel,
+      timeoutHeight: undefined,
+      timeoutSecOffset: undefined,
+      memo: `Namada->${keplrChain}`, // how is this different from the wrapper memo?
```
jurevans, line 86: The wrapper memo in this case will be applied with the Tx on Namada, but the IBC Transfer memo gets passed on via the IBC Msg sent to destination, AFAIK. I think in this case, we could omit the wrapper memo if you like. It is kind of weird that we have memos at different levels, the Tx wrapper args, then individual commitments all have their own memo, but in this case, the IbcTransfer memo would be useful to see on Keplr!

emccorson: Ah, I see! I've just gone ahead and removed the memos for now, since they're not used yet, but I can add them in again later.

PR #1136, opened 2024-09-24
feat/1098: Namadillo: Fetch, Store, and Load MASP params in SDK

jurevans:
Relates to #1098 (for building shielded Tx)

- [x] Update proxy script to proxy the MASP Params URL (to get around the CORS errors on local dev)
- [x] Update SDK to accept an alternate URL (proxied URL) to fetch MASP params
- [x] Issue check for masp params: If we don't have them, fetch and store params. Once stored, load them into SDK

### Testing

- In Namadillo: `yarn dev:proxy`
  - **NOTE** In local dev mode, you only need to run the proxy once when to fetch the params. Once it's downloaded, the `masp.hasMaspParams()` check should always return true, as it is now persisted in IndexedDB!
- You'll probably need to just monitor the network requests the first time you load it, ensure the MASP param requests runs the first time, then on future runs it should not request anything. I didn't want to leave any `console.log` :D

### DIFF
--- a/apps/namadillo/src/hooks/useSdk.tsx
+++ b/apps/namadillo/src/hooks/useSdk.tsx
@@ -11,8 +11,14 @@ import {
   useEffect,
   useState,
 } from "react";
+import Proxies from "../../scripts/proxies.json";
 
 export const SdkContext = createContext<Sdk | undefined>(undefined);
+
+const { VITE_PROXY: isProxied } = import.meta.env;
+
+const paramsUrl =
+  isProxied ? `http://localhost:${Proxies[0].proxyPort}/proxy/` : undefined;
 
 const initializeSdk = async (): Promise<Sdk> => {
   const { cryptoMemory } = await initSdk();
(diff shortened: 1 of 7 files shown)


apps/namadillo/src/hooks/useSdk.tsx
```diff
@@ -47,7 +53,19 @@ export const SdkProvider: FunctionComponent<PropsWithChildren> = ({
 
   useEffect(() => {
     if (nativeToken.data) {
-      getSdkInstance().then((sdk) => setSdk(sdk));
+      getSdkInstance().then((sdk) => {
+        setSdk(sdk);
+        const { masp } = sdk;
+        masp.hasMaspParams().then((hasMaspParams) => {
+          if (!hasMaspParams) {
```
mateuszjasiuk, line 60: I've tried to test it and I think we have to remove this `!` otherwise params will never fetch 🤔

jurevans: yep, sorry, I was doing a test and forgot to remove this. Thanks!

jurevans: Fixed!

mateuszjasiuk (approved):
LGTM!

PR #1143, opened 2024-09-27
feat/592: Validate length and checksum of fetched masp param

### LINKED ISSUE #592: Add mechanism for checking correctness of masp-params
```rs
pub const MASP_SPEND_BYTES: u64 = 49848572;
pub const MASP_CONVERT_BYTES: u64 = 22570940;
pub const MASP_OUTPUT_BYTES: u64 = 16398620;
```
We should verify if params stored in IDB have correct sizes, and refetch/restore if needed.

jurevans:
Resolves #592 

This PR validates the fetched MASP param before storing it. This will reject the promise and prevent an invalid param from being stored should length or checksum not match.

## TESTING

- In `packages/shared/lib/src/sdk/mod.ts` https://github.com/anoma/namada-interface/blob/e160a31a23a259e9e7a4ed205b7d15fcbfa596e9/packages/shared/lib/src/sdk/mod.ts#L21-L40 modify these values to invalidate them (either length or hash, length is checked first!)
- Rebuild the shared package: `yarn wasm:build` (Even though these changes are 100% TS, they are a part of `shared` lib and will not auto-reload in dev mode)
- Start Namadillo with `yarn dev:proxy` (bypasses CORS issue with localhost)
- If you have already successfully downloaded MASP params, just comment out the lines below in `useSdk` to force a download on next refresh: https://github.com/anoma/namada-interface/blob/e160a31a23a259e9e7a4ed205b7d15fcbfa596e9/apps/namadillo/src/hooks/useSdk.tsx#L60-L62
- You should see errors for each param that has issues. Currently if any one of these fails, `hasMaspParams()` will return false, so it will retry all of them on next refresh.
- Also, you can run `yarn dev` locally (without the proxy) which will trigger the `retry` behavior, it is currently configured to retry a total of 3 times with 3 second intervals in between before failing

## TODO

- How should we respond to a MASP param validation failure? One option would be to simply prompt the user to try again, perhaps we could have a warning modal with a `Retry` button that invokes the download again.

## NOTES

- What could cause an invalid MASP param download?
  - If a download is interrupted due to connectivity issues, the promise will likely just hang
  - If the user is running this locally and overrides the trusted setup params URL, they could be loading invalid params
    - If the length is valid, it will then check the checksum, which must match our constants
  - This feature is for piece of mind, so that the user can be sure they are using the trusted params. Only the trusted params will successfully be stored in IndexedDB

### DIFF
--- a/apps/namadillo/src/hooks/useSdk.tsx
+++ b/apps/namadillo/src/hooks/useSdk.tsx
@@ -63,7 +63,7 @@ export const SdkProvider: FunctionComponent<PropsWithChildren> = ({
           masp
             .fetchAndStoreMaspParams(paramsUrl)
             .then(() => masp.loadMaspParams(""))
-            .catch((e) => console.error(`${e}`));
+            .catch((e) => console.error(`Error downloading MASP params ${e}`));
         });
       });
     }
--- a/apps/namadillo/src/hooks/useSdk.tsx
+++ b/apps/namadillo/src/hooks/useSdk.tsx
@@ -13,7 +13,21 @@ import {
 } from "react";
 import Proxies from "../../scripts/proxies.json";
 
-export const SdkContext = createContext<Sdk | undefined>(undefined);
+export enum MaspParamsStatus {
+  Fetching = "fetching",
+  Loaded = "loaded",
+  Error = "error",
+}
+
+type SdkContext = {
+  sdk?: Sdk;
+  maspParamsStatus?: MaspParamsStatus;
+};
+
+export const SdkContext = createContext<SdkContext>({
+  sdk: undefined,
+  maspParamsStatus: undefined,
+});
 
 const { VITE_PROXY: isProxied } = import.meta.env;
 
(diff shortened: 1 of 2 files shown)


apps/namadillo/src/hooks/useSdk.tsx
```diff
@@ -13,7 +13,21 @@ import {
 } from "react";
 import Proxies from "../../scripts/proxies.json";
 
-export const SdkContext = createContext<Sdk | undefined>(undefined);
+export enum MaspParamsStatus {
```
pedrorezende, line 16: Nitpick, but more an idea. What do you think of using QueryStatus type from `@tanstack/react-query` instead? So we can make this loading state type standard across the app

packages/shared/lib/src/sdk/mod.ts
```diff
@@ -1,46 +1,121 @@
...
+  Convert = "masp-convert.params",
+  Spend = "masp-spend.params",
+}
+
+type MaspParamBytes = {
+  param: MaspParam;
+  bytes: Uint8Array;
+};
+
+const MASP_PARAM_ATTR: Record<
```
pedrorezende, line 23: Can we add a comment telling from where these numbers / hashes come from?

pedrorezende (approved):
LGTM!

apps/namadillo/src/hooks/useSdk.tsx
jurevans: Done!

pedrorezende (approved):

PR #1157, opened 2024-10-03
bug/592: Provide MASP Params as local asset

jurevans:
Relates to #592 

This PR allows us to fetch the MASP params directly from Namadillo assets, which would eliminate the need to handle potential CORS issues.

Adding this here as an option to dealing with the CORS issues seen on `https://interface.knowable.run`. There are a couple alternatives though:

- Keep the Github URL, but ensure the all hosts are configured with the correct CORS allow origins header
- Host these params elsewhere


mateuszjasiuk (approved):
Tested! Works great!

PR #1162, opened 2024-10-04
feat: begin integration of IBC with transfer module

emccorson:
This is the first attempt at integration the prototype IBC form with the transfer module.

Some caveats:
- Only transparent transfers are working
- We still need to set the channel ID manually for now
- Assets will take some time to load after the chain has been selected

Testing
----------

### Namadillo settings
Indexer: https://indexer.public.heliax.work/internal-devnet-44a.1bd3e6ca62
RPC: https://proxy.public.heliax.work/internal-devnet-44a.1bd3e6ca62

### IBC transfer to Namada settings
Source channel: channel-4353
Destination channel: channel-0
(Note: destination channel is only needed for transfers to *shielded* accounts)

### Instructions
1. Go to Namadillo settings -> advanced, and make sure "Enable Testnets" is checked
2. Go to <Namadillo URL>/settings/features and check "IBC Transfers"
3. Make sure you have ATOM for the Cosmos testnet (theta-testnet-001) in Keplr
4. Go to the Namadillo IBC Transfer page
   a. Enter the source and destination channel IDs as above
   b. Select the Cosmos testnet (the one called "[DEPRECATED] Cosmos Hub...")
   c. Wait for assets to load and then select "Cosmos"
   d. Enter an amount
   e. Select "Transparent" for the Namada account
   f. Click "Submit" and approve the Keplr popup
5. After a few seconds, check your ATOM was received in Namada with:
    
```
namadac balance \
    --owner <your address> \
    --token transfer/channel-0/uatom \
    --node https://proxy.public.heliax.work/internal-devnet-44a.1bd3e6ca62
```


apps/namadillo/src/App/Ibc/IbcTransfer.tsx
```diff
@@ -1,172 +1,227 @@
...
+          isLoadingAssets: assetBalanceQuery.isFetching,
+          availableAssets:
+            Object.values(assetsBalances).map((el) => el.asset) || [],
+          selectedAsset,
+          onChangeSelectedAsset: setSelectedAsset,
+          availableAmount:
+            selectedAsset && selectedAsset.base in assetsBalances ?
+              assetsBalances[selectedAsset.base].balance
+            : undefined,
+          availableChains: Object.values(knownChains).map(
```
emccorson, line 196: I think we will need to deal with the situation where one or more of the chains we want to list is not known by the user's Keplr wallet. For example, at the moment I am getting the error `There is no chain info for mocha-4` because I am missing that chain in Keplr.

We can get the chains in Keplr using `keplr.getChainInfosWithEndpoints`. I'm not sure how we should handle this situation though. We could just not display unknown chains, or disable them, or even prompt them to be added to Keplr.

apps/namadillo/src/atoms/integrations/services.ts
```diff
@@ -0,0 +1,54 @@
...
+
+    const fee = {
+      amount: coins("0", token),
+      gas: "222000",
+    };
+
+    const response = await client.sendIbcTokens(
+      sourceAddress,
+      destinationAddress,
+      coin(amount.toString(), token),
```
emccorson, line 42: This will be a pain, but somewhere we're going to need to convert the amount the user input to a different denomination for sending (I think). For example, the user inputs 2 ATOM, and we will need to convert it to 2000000 uatom.

apps/namadillo/src/atoms/registry/services.ts
```diff
@@ -0,0 +1,37 @@
...
+
+export const mapCoinsToAssets = (
+  coins: Coin[],
+  assetList: AssetList
+): Record<string, AssetWithBalance> => {
+  return coins.reduce((prev, current) => {
+    const asset = assetList.assets.find(
+      (asset) => asset.base === current.denom
+    );
+    if (!asset) return prev;
```
emccorson, line 28: I think we'll still want to display an asset even if we don't have information for it in the asset list. For example, NAM isn't going to show up in the asset list (although actually maybe we could add it), but we still want to support transferring NAM over IBC.

The easiest thing to do here might be to change TransferModule to specify the properties it needs for assets e.g. instead of `Asset`, use `Pick<Asset, "name" | "base">` or something like that. Then we can more easily construct what the TransferModule needs in these cases.

apps/namadillo/src/atoms/registry/functions.ts
```diff
@@ -0,0 +1,46 @@
...
+    availableRpcs[Math.floor(Math.random() * availableRpcs.length)];
+
+  return randomRpc.address;
+};
+
+// eslint-disable-next-line
+
+type QueryFn<T> = (rpc: string) => Promise<T>;
+
+export const queryAndStoreRpc = async <T>(
```
emccorson, line 21: This is cool! I think you're right about using the RPCs sequentially instead of randomly. Just from brief testing, it seems hit or miss whether an RPC will actually work (and I'm hoping they are ordered roughly by reliability).

apps/namadillo/src/atoms/registry/services.ts
pedrorezende: I think this might be a bigger problem, because if the asset is not known in the asset list, maybe we will need to query some of its info from somewhere. Can we separate this into a new issue?

apps/namadillo/src/App/Ibc/IbcTransfer.tsx
jurevans: That's a good point. I wonder if we should even show any chains not currently enabled by the user's wallet, since we wouldn't be able to get those addresses or sign unless we prompt them to first enable that chain. Hmmm

pedrorezende: Just made a few changes giving the possibility of adding an unknown chain to the Keplr wallet. This will be useful mainly for the testnet chains.

euharrison (approved):
I'm able to do an IBC transfer 🚀 
https://explorer.polypore.xyz/theta-testnet-001/tx/495021E251E23B3B9C619F91383114DFBE654185106FBDF1142F1AEB93BDF416

I believe we can do more upgrades on future PRs

PR #1163, opened 2024-10-04
feat: add support for IBC deposits to shielded

emccorson:
This PR adds support for IBC deposits to shielded Namada addresses.

Testing instructions are the same as #1162 but running `namadac shielded-sync` is necessary before checking that balance was received.

---

### Changed
- Re-add source channel input element (since I think we will need to change channels relatively frequently before mainnet)
- Rename `IBCTransferParams` to `IbcTransferParams`

### Added
- Support IBC deposits to Namada shielded addresses
- Add SDK methods to get MASP address and to generate IBC shielding memo

### Fixed
- Fix Long value being passed instead of BigInt in `timeoutTimestampNanoseconds`
  - I don't think this is what was causing the timeouts over IBC but I'm not 100% sure

### DIFF
--- a/packages/shared/lib/src/sdk/mod.rs
+++ b/packages/shared/lib/src/sdk/mod.rs
@@ -13,7 +13,7 @@ use crate::utils::set_panic_hook;
 use crate::utils::to_bytes;
 use crate::utils::to_js_result;
 use gloo_utils::format::JsValueSerdeExt;
-use namada_sdk::address::Address;
+use namada_sdk::address::{Address, MASP};
 use namada_sdk::borsh::{self, BorshDeserialize};
 use namada_sdk::eth_bridge::bridge_pool::build_bridge_pool_tx;
 use namada_sdk::hash::Hash;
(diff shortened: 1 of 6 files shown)


packages/shared/lib/src/sdk/mod.rs
```diff
@@ -13,31 +13,38 @@ use crate::utils::set_panic_hook;
...
 use gloo_utils::format::JsValueSerdeExt;
-use namada_sdk::address::Address;
+use namada_sdk::address::{Address, MASP};
 use namada_sdk::borsh::{self, BorshDeserialize};
 use namada_sdk::eth_bridge::bridge_pool::build_bridge_pool_tx;
 use namada_sdk::hash::Hash;
 use namada_sdk::io::NamadaIo;
 use namada_sdk::key::{common, ed25519, SigScheme};
-use namada_sdk::masp::ShieldedContext;
+use namada_sdk::masp::{ShieldedContext};
```
jurevans, line 22: should probably remove these `{}`

jurevans (approved):
This looks really good! Only very minor comments

PR #1168, opened 2024-10-09
PoC: Adding offscreen & web worker to the extension

jurevans:
Restoring offscreen/web-workers as of: https://github.com/anoma/namada-interface/tree/5eaf361df780a660dd73b2c462a03c057c47b241

- Updated naming for generating proofs (instead of Transfers)
- Use new version of SDK package
- Updated `@types/chrome` to add `chrome.offscreen.Reason.WORKERS` to support newer browsers

I think we can pass the Tx hash of the shielded Tx as the `msgId` to link it up in the interface, but if we need to change anything in the msg, we can do that. This PR is mainly to get the pieces in place

### DIFF
--- a/apps/extension/src/background/keyring/keyring.ts
+++ b/apps/extension/src/background/keyring/keyring.ts
@@ -5,6 +5,7 @@ import {
   Bip44Path,
   DerivedAccount,
   SignArbitraryResponse,
+  TransferMsgValue,
   TxProps,
 } from "@namada/types";
 import { Result, assertNever, truncateInMiddle } from "@namada/utils";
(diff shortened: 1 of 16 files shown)


apps/extension/src/background/keyring/keyring.ts
```diff
@@ -592,4 +595,36 @@ export class KeyRing {
...
   }
+
+  async generateProof(
+    transferMsg: Uint8Array,
+    generate: (spendingKey: SpendingKey) => Promise<void>
+  ): Promise<void> {
+    await this.vaultService.assertIsUnlocked();
+
+    // We need to get the source address to find the spending key
+    const { sources } = deserialize(Buffer.from(transferMsg), TransferMsgValue);
```
jurevans, line 606: Maybe we could use `@heliax/namada-sdk` `deserializeTx` for this, though I think this should work, unless we start passing a batch Tx or multiparty transfer to this

jurevans:
Closing, as we will not be doing this!

PR #1174, opened 2024-10-11
feat: support decoding ibc tx details in sdk

emccorson:


jurevans (approved):
This looks good! I think (if you like) this could go ahead and go in. I think we can chat more about how to handle the `timeout_sec_offset` & `shielding_data` then one of us can follow up in another PR, but we're pretty well covered for transparent IBC with this!

emccorson:
I will go ahead and merge this as suggested, and I've created an issue #1196 to follow up.

PR #1188, opened 2024-10-14
Components: create TabContainer component

pedrorezende:
This PR creates a simple <TabContainer> component in order to manage tabs and their contents.

### DIFF
--- a/packages/components/src/TabContainer.tsx
+++ b/packages/components/src/TabContainer.tsx
@@ -1,0 +1,66 @@
+import { ActionButton, ActionButtonProps } from "@namada/components";
+import clsx from "clsx";
+import { useState } from "react";
+import { twMerge } from "tailwind-merge";
+
+type TabContents = {
+  title: React.ReactNode;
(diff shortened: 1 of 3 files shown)


packages/components/src/TabContainer.tsx
```diff
@@ -0,0 +1,65 @@
...
+    <div>
+      <div role="tablist" aria-label={title} className="flex">
+        {tabs.map((tab, index) => (
+          <ActionButton
+            key={index}
+            id={`tab-${id}-${index}`}
+            role="tab"
+            aria-selected={activeTabIndex === index}
+            aria-controls={`tabpanel-${id}-${index}`}
+            tabIndex={activeTabIndex === index ? 0 : -1} // Only the active tab is focusable
```
euharrison, line 33: personally, I prefer to unset the tabIndex directly on the component because it can easily conflict with other input elements on the page and mess it around, then usually I define it on the page level. But we can test it as well and observe if it's necessary or not :)

PR #1191, opened 2024-10-17
feat: unshield + shielded transfers + disposable gas payer

mateuszjasiuk:
Resolves [#1208](https://github.com/anoma/namada-interface/issues/1208)
This adds support for unshielding and shielded transfers.
You should be able to test it using `MASP` tab in  the UI.
Transfers using `Transfer` tab are not yet supported.
Additionally you can add `<WorkerTest />` component somewhere and test via the terminal.
This should work with disposable gas payer, meaning each unshielding/shielded transfer should be signed by randomly generated keypair.


euharrison (approved):
Awesome! ✅

apps/namadillo/src/workers/ShieldWorker.ts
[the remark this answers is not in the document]
mateuszjasiuk: RIght I totally missed the Unshield worker. I will delete one

packages/sdk/src/signing.ts
```diff
@@ -17,19 +17,23 @@ export class Signing {
...
    * @param signingKey - private key
+   * @param xsks - spending keys
    * @param [chainId] - optional chain ID, will enforce validation if present
    * @returns signed tx bytes - Promise resolving to Uint8Array
    */
   async sign(
     txProps: TxProps,
     signingKey: string,
+    xsks: string[],
     chainId?: string
```
jurevans, line 28: Quick question on `xsks: string[]` - would it be better to treat this as an optional parameter since this is specific to MASP?

packages/shared/lib/Cargo.toml
```diff
@@ -18,7 +18,7 @@ nodejs = []
 web = []
 
 [build-dependencies]
-namada_tx = { git = "https://github.com/anoma/namada", tag = "v0.46.0" }
+namada_tx = { git = "https://github.com/anoma/namada", rev = "49a4a5d3260423df19ead14df82d18a51fa9b157" }
```
jurevans, line 21: Can this be pointed to `v0.46.1`, or are we waiting on the next release?

jurevans (approved):
LGTM! Just a couple small comments, but nothing necessarily needs to change!

packages/shared/lib/Cargo.toml
mateuszjasiuk: I think this one is later probably like 1.0.1 or sth :)

mateuszjasiuk: But it should be safe to merge I think.

PR #1193, opened 2024-10-21
Update README.md

chimmykk:
Syntax Error in chainId Declaration: Should be ;


jurevans (approved):
Thanks for catching this!

Issue #1190, opened 2024-10-16
Namadillo: IBC -> Determine amount for transfer fees


emccorson:
Following on from #1194, we need to decide how much transfer fees should be.

We can probably do whatever Keplr is doing to match their transfer fees. I looked into using the values from the chain regsitry configs we are currently using, but those fees seemed to be much higher than Keplr.

Currently, the fee is hardcoded as `1` for every chain, so we should determine the fees per chain and then remove this hardcoding.

euharrison:
Will be handled at https://github.com/anoma/namada-interface/issues/1486

PR #1194, opened 2024-10-21
IBC transfer fee support

### LINKED ISSUE #1135: IBC: Figure out how to calculate fees for Shield All
When performing a Shield All IBC transfer, as far as I understand we want to send all a user's funds for a given token to Namada, minus transaction fees. However, even though we specify the amount to send up front, the transaction fee is set by the user in the Keplr popup. We need to figure out a way of sending all funds to Namada accounting for this.

emccorson:
This PR adds support for things related to IBC transfer fees. It makes Namadillo set the transfer fee instead of Keplr, and makes available amount take into account transfer fee. These changes are both needed to make the max button work (which should now be working), and are also prerequisites for Shield All (to follow in a later PR). There are also some other minor fixes included in the list below.

Closes #1135.

---

### Changed
- Set fees in Namadillo instead of Keplr
- Make available amount account for fees to make the max button work
- Set TransferModule fee in base unit and convert for display (e.g. transfer fee of "1" displays as "0.000001 NAM")

### Removed
- Remove TransferModule examples to avoid maintenance

### Fixed
- Show the correct fee token based on the selected chain


apps/namadillo/src/App/Ibc/IbcTransfer.tsx
```diff
@@ -35,6 +35,29 @@ export const IbcTransfer: React.FC = () => {
   const performIbcTransfer = useAtomValue(ibcTransferAtom);
   const defaultAccounts = useAtomValue(allDefaultAccountsAtom);
 
+  const transactionFee = useMemo(() => {
+    if (typeof registry !== "undefined") {
+      // TODO: can we get a better type for registry to avoid optional chaining?
+      // TODO: some chains support multiple fee tokens - what should we do?
```
jurevans, line 41: Good point - maybe at some point we could let user's choose a fee token that has balance, otherwise prevent them from submitting? I think that can be done later! Defaulting to native token is probably a good idea for now

apps/namadillo/src/App/Ibc/IbcTransfer.tsx
```diff
@@ -35,6 +35,29 @@ export const IbcTransfer: React.FC = () => {
...
+      const feeToken = registry.chain?.fees?.fee_tokens?.[0];
+
+      if (typeof feeToken !== "undefined") {
+        const asset = registry.assets.assets.find(
+          (asset) => asset.base === feeToken.denom
+        );
+
+        if (typeof asset !== "undefined") {
+          return {
+            amount: BigNumber(1), // TODO: remove hardcoding
```
jurevans, line 51: We probably just need a follow-up issue to handle fees when submitting to Keplr right?

jurevans (approved):
LGTM! Just left a couple comments - we can follow-up to address the `todo` comments

apps/namadillo/src/App/Ibc/IbcTransfer.tsx
emccorson: I think so! I've adapted #1190 to track this.

apps/namadillo/src/App/Ibc/IbcTransfer.tsx
emccorson: I've added #1203 to track this.

emccorson:
I had to move around a few things when rebasing to main, but the functionality is the same, so I'll go ahead and merge this and if anything is broken I can fix it later.

PR #1195, opened 2024-10-21
Namadillo: IBC improvements + Withdrawal

pedrorezende:
This PR refactors and reorganizes some parts of the IBC transfer codebase, and also adds the transfer module for withdrawals. It also displays IBC transfer fields for TransferModule, as a temporary measure until we have our channels well defined.

This is one more step forward for IBC transfers, but still requires a few elements, like the list of Namada assets, and a proper handling of the transfer confirmation / status.


apps/namadillo/src/App/Ibc/assets/token.png
euharrison, line 1: should we use these images as svg?

euharrison (approved):
I'm able to execute the deposit 🚀 
https://www.mintscan.io/cosmos-testnet/tx/9DF1D22E78B6E9B90E2D623EAD1629ACA5146B4CBB288C4AE5E8345225D97EB0

apps/namadillo/src/atoms/integrations/atoms.ts
```diff
@@ -53,31 +54,38 @@ export const ibcTransferAtom = atomWithMutation(() => {
...
-});
+export const knownChainsAtom = atom<Record<string, ChainRegistryEntry>>(
+  (get) => {
+    const settings = get(settingsAtom);
+    const knownChains = getKnownChains(settings.enableTestnets);
+    const map: Record<string, ChainRegistryEntry> = {};
+    knownChains.forEach((chain) => {
+      map[chain.chain.chain_id] = chain;
+    });
+    return map;
```
euharrison, line 89: We are executing `Object.value(knownChains` four times that actaully reverts this map

Screenshot 2024-10-22 at 17 10 08

I believe it'll be better if we return the original array and create a new atom for the map, that derivates from this atom

Or we can use a helper function like `_.keyBy` https://lodash.com/docs/4.17.15#keyBy directly in the component

pedrorezende: I've renamed a bit the atoms. Left the registry as the map structure with more info, and the availableChains as a list of chains

apps/namadillo/src/App/Ibc/assets/token.png
pedrorezende: Good catch! I've reduced the sizes of png images and converted token.png to svg.

emccorson (approved):
Looks good to me! I will follow up with a PR to switch back on IBC withdrawals.

apps/namadillo/src/integrations/Keplr.ts
```diff
@@ -0,0 +1,28 @@
...
+  OfflineDirectSigner,
+} from "@keplr-wallet/types";
+import { ChainRegistryEntry } from "types";
+import { WalletInterface } from "./types";
+import { basicConvertToKeplrChain } from "./utils";
+
+type Signer = OfflineAminoSigner & OfflineDirectSigner;
+const keplr = (window as KeplrWindow).keplr!;
+
+export class KeplrWalletManager implements WalletInterface {
```
emccorson, line 13: One other thing I just realised when rebasing my PR to this one: if it all possible, I would really like not to abstract here with an interface until we're forced to by supporting other wallets. The reason is that we have to do some things that are specific to Keplr (in the case I'm thinking of, set and unset a Keplr option before and after submitting a transfer). If we use an interface for this, we'll need to add before/after hooks and it will get quite complicated. We had something like this before with the previous `Integration` interface, and ended up adding dummy methods in places where we needed to do things specific to one extension but not another. I think the future when we support a second extension would be a better time to add abstractions, since at that time we will know the exact requirements (and also that time may never come depending on requirements). What do you think?

jurevans: If we keep using a wallet abstraction, we could at least only require implementations for things that all wallets contain (where the interfaces can easily match), everything else could be added on to the specific wallet. Otherwise, it might be easier to just have 2 different types of wallets (eventually 3 with Metamask). Just a thought, I'm fine either way, so long as we can avoid those dummy implementations

jurevans (approved):
LGTM! 🥇

apps/namadillo/src/integrations/Keplr.ts
pedrorezende: I'll keep this abstraction because I think it will be easier to remove it in the future than add other wallets if we don't have this implemented. But I renamed the interface to `WalletConnector`, to be something more specific.

PR #1201, opened 2024-10-24
Namadillo: handling errors on IBC transfer form

pedrorezende:
This PR adds some nice ways to handle errors on IBC Transfer and a warning on TransferModule showing why the submit button is disabled.

### DIFF
--- a/apps/namadillo/src/App/Ibc/IbcTransfer.tsx
+++ b/apps/namadillo/src/App/Ibc/IbcTransfer.tsx
@@ -94,6 +94,7 @@ export const IbcTransfer: React.FC = () => {
     ibcOptions,
   }: OnSubmitTransferParams): Promise<void> => {
     try {
+      setGeneralErrorMessage("");
       setCurrentStep(1);
 
       if (typeof sourceAddress === "undefined") {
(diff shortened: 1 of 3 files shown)


apps/namadillo/src/App/Transfer/TransferModule.tsx
```diff
@@ -23,6 +24,7 @@ export type TransferModuleConfig = {
   chain?: Chain;
   onChangeChain?: (chain: Chain) => void;
   isShielded?: boolean;
+  errorMessage?: string;
```
emccorson, line 27: It looks like this `errorMessage` is not used anywhere, just the one in `TransferModuleProps`. I think you could also remove the `export` from this type since it isn't used anywhere else.

apps/namadillo/src/App/Ibc/IbcTransfer.tsx
```diff
@@ -115,7 +116,8 @@ export const IbcTransfer: React.FC = () => {
         },
       });
       setCurrentStep(2);
-    } catch {
+    } catch (err) {
+      setGeneralErrorMessage(err + "");
```
emccorson, line 120: I'm not sure if this is a valid case, but do we also need to unset `generalErrorMessage` when the submission is successful? I think we display a different page on success, so maybe not, but I'm not sure.

apps/namadillo/src/App/Transfer/TransferModule.tsx
```diff
@@ -161,6 +165,33 @@ export const TransferModule = ({
...
+    if (!source.selectedAsset && source.onChangeSelectedAsset) {
+      return "Select Asset";
+    }
+
+    // TODO: this should be updated for nfts
+    if (!amount || amount.eq(0)) {
+      return "Define an amount to transfer";
+    }
+
+    // TODO: amount + fee < available amount
```
emccorson, line 190: It should be possible to do this now that #1194 is merged.

emccorson (approved):
Looks great! I just left a few minor comments, but I think this can go in any time.

PR #1209, opened 2024-10-24
feat: render dollar values on masp overview

### LINKED ISSUE #1171: Connect the Shielded Overview Table with real data
We need to connect the Shielded Overview Table with real data

We can skip the NFT for now, focused only on the Fungible Table

Screenshot 2024-10-09 at 18 49 11

euharrison:
Render dollar values on masp overview

We are using the CoinGecko api for the value converstion.

To enable this api, you'll need to create the `.env` file on namadillo folder, you can copy the sample from `.env.sample`.  And fill the environment value of `COIN_GECKO_API_KEY`. 

In case there is a coin that doesn't exist on the CoinGecko (like `nam`), or it's missing the api key, it'll not be able to calculate the value, and it'll be reflected on the UI

Closes #1171 

Screenshot 2024-10-24 at 19 18 53

Screenshot 2024-10-24 at 19 19 12


apps/namadillo/src/atoms/masp/functions.ts
```diff
@@ -0,0 +1,36 @@
...
+import { TokenBalance } from "./atoms";
+
+/**
+ * Find the expoent of a denom (`uatom`, `atom`, etc).
+ * Useful to calculate the token amount with correct decimal part
+ * @param asset
+ * @param denom
+ * @returns The expoent, or zero if not found
+ */
+export const findExpoent = (asset: Asset, denom: string): number =>
```
pedrorezende, line 12: Nitpick: Maybe these functions could be moved to integrations or other place more generic than MASP

apps/namadillo/src/hooks/useBalances.ts
```diff
@@ -35,30 +35,32 @@ export const useBalances = (): useBalancesOutput => {
...
   );
-  const shieldedAmount = new BigNumber(shieldedBalance || 0);
+  const shieldedAmount = shieldedDollars === undefined ? null : shieldedDollars;
   const totalAmount = BigNumber.sum(
     availableAmount,
     bondedAmount,
     unbondedAmount,
     withdrawableAmount,
-    shieldedAmount
+    shieldedDollars || 0
```
pedrorezende, line 56: Hmm, do you think we should make this sum using dollars? Maybe I'm missing something, but I'm not sure it looks correct

apps/namadillo/src/atoms/masp/atoms.ts
```diff
@@ -1,18 +1,36 @@
...
+export type TokenBalance = {
+  address: string;
+  amount: string;
+  denom: string;
+  asset: Asset;
+  balance: BigNumber;
+  dollar?: BigNumber;
+};
+
+const NAM_DENOM = "nam";
```
pedrorezende, line 26: Can we add a TODO comment here, so we can replace it by the namada-chain-registry entry in the future?

apps/namadillo/src/atoms/masp/atoms.ts
```diff
@@ -1,18 +1,36 @@
...
 import { atomWithQuery } from "jotai-tanstack-query";
+import { namadaAsset } from "registry/namadaAsset";
+import { unknownAsset } from "registry/unknownAsset";
 import { getSdkInstance } from "utils/sdk";
+import { findExpoent, sumDollars } from "./functions";
+import { fetchCoinPrices } from "./services";
+
+export type TokenBalance = {
+  address: string;
+  amount: string;
```
pedrorezende, line 19: I know this type is coming from somewhere else, but is there a reason why amount is a `string` instead of a `BigNumber` here? Also, maybe we should use the denom from Asset

pedrorezende (changes requested):
Good one! I've opened a few comments to discuss, but I think my main concern is related to where the CoinGecko API key should lie. I'm not super familiar with the MASP stuff, so maybe it would be great if we could have the review from @emccorson or @jurevans too, mainly on the masp/atoms section

apps/namadillo/.env.sample
```diff
@@ -0,0 +1 @@
+COIN_GECKO_API_KEY=
```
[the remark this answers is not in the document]
euharrison: Agreed. This is only the simplest solution to enable the fiat feature. It's still not clear the definition of using the CoinGecko api, maybe we go with an alternative api. I'm planning to discuss this on the afternoon meeting

apps/namadillo/src/hooks/useBalances.ts
euharrison: This is how the Figma layout looks like 🤷 . I'm refactoring this on the next PR that will upgrade the AccountOverview page, but still with the sum of dollars. I believe this is like this because it's doesn't make sense to sum different tokens as they have different values

apps/namadillo/src/atoms/masp/atoms.ts
euharrison: The `address` and `amount` are the original values returned from the sdk when we execute `sdk.rpc.queryBalance`. It returns `Array<[address:string, amount:string]>`

The processed data are the other values: `denom`, `asset`, `balance` (the BigNumber you are looking for, but considering the `expoent`), and `dollar`. Dollar is optional as it could not be calculated if it's missing the value from the CoinGecko api (`nam` case), or in case it's missing the CoinGecko api key

Effectively we don't use the values of `address` and `amount` on the components, so I can remove it from the  `TokenBalance` if you feel it'll be cleaner

apps/namadillo/src/hooks/useBalances.ts
pedrorezende: I understand. But in this case you're summing NAM with dollars, right? Maybe we should separate this calc from useBalances and convert the totalAmount to dollars first in order to make the proportion displayed in the chart

pedrorezende: I understand that until the mainnet we won't be able to query the value in dollars for nam, but we can mock this value temporarily to proceed with the implementation

apps/namadillo/src/atoms/masp/atoms.ts
pedrorezende: Gotcha. I think this is fine for now :)

apps/namadillo/.env.sample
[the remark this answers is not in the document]
pedrorezende: I don't remember if we got to a final decision on this. But can we get rid of the .env file in order to let everything inside the config.toml? (i.e. when we redistribute a new Namadillo release, people can get the bundle and just change the config.toml instead of rebuild everything from scratch to add the CoinGecko or other API key)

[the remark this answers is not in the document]
pedrorezende: I'm not familiar with CoinGecko API, I'm supposing this is not a secret key :)

euharrison:
Due to the recent changes of how we are going to handle the dollar amounts, I'll close this PR and open a new one integrated with the Account Overview

PR #1210, opened 2024-10-25
feat: re-enable IBC withdrawals for NAM

emccorson:
This PR re-enables IBC withdrawals for NAM (other tokens will follow later).

We have a problem where the TransferModule and cosmjs expect amounts to be in the base denom (e.g. namnam), but the Namada SDK expects amounts to be in the display denom (e.g. NAM). This has resulted in some temporary hacky code in this PR to convert between namnam and NAM which I'll remove this once I understand how to handle this properly, but for now the hacky code is enough to switch back on IBC withdrawals for NAM. Related: #1211.

---

### Changed
- Move available amount fee deduction calculation to TransferModule
- Return `undefined` instead of `0` from available balance calculation when initial balance is undefined.

### Added
- Re-enable IBC withdrawals for NAM

### Fixed
- Rename "unam" to "namnam" in Namada asset

### DIFF
--- a/apps/namadillo/src/App/Ibc/IbcTransfer.tsx
+++ b/apps/namadillo/src/App/Ibc/IbcTransfer.tsx
@@ -44,7 +44,7 @@ export const IbcTransfer: React.FC = () => {
   } = useWalletManager(keplr);
 
   const {
-    balance: totalAvailableAmount,
+    balance: availableAmount,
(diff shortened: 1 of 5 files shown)


pedrorezende (approved):
Approving and merging, so we can have this in and keep looking for the best solution to deal with the exponent for min denom.

PR #1212, opened 2024-10-25
feat/1206 - SDK: Adding events to shielded sync

### LINKED ISSUE #1206: SDK: Shielded-Sync should broadcast events
When shielded sync starts, we should transmit progress events (`ProgressBar`) that contains:

- `current` 
- `total`

More info TBD

jurevans:
Resolves #1206 
todo
- [x] export progress types - scanned/fetched/applied
- [x] remove "// total is 0 so we do not get NaN"
- [x] fix circular dependency in balance/services.ts
- [x] export progress bar status
- [x] add event payload types
- [x] ~~EventDispatcher scope type as enum?~~
- [x] check if  options.bubbles(true); etc is needed

### DIFF
--- a/apps/namadillo/src/atoms/balance/services.ts
+++ b/apps/namadillo/src/atoms/balance/services.ts
@@ -5,6 +5,9 @@ import { Balance, SdkEvents } from "@namada/sdk/web";
 import { getSdkInstance } from "utils/sdk";
 import {
   Events,
+  ProgressBarFinished,
+  ProgressBarIncremented,
+  ProgressBarStarted,
   Worker as ShieldedSyncWorkerApi,
 } from "workers/ShieldedSyncWorker";
 import ShieldedSyncWorker from "workers/ShieldedSyncWorker?worker";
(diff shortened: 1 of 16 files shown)


euharrison (commented):
I'm getting this error when the sync starts

Screenshot 2024-11-19 at 19 51 26

And this one when it finishes

Screenshot 2024-11-19 at 19 53 13

Do you know what it could be?

I already run the `yarn wasm:build:dev`

mateuszjasiuk:
I'm not sure, maybe clear the old context from the IDB?

euharrison (approved):
Nice! 🚀 

I'm still getting this error, but it's working like a charm!

Screenshot 2024-11-21 at 12 00 34

apps/namadillo/src/atoms/balance/services.ts
```diff
@@ -39,7 +41,15 @@ export function shieldedSync(
...
-    emitter.emit(event.data.type, event.data);
+    if (event.data.type === SdkEvents.ProgressBarStarted) {
+      emitter.emit(event.data.type, event.data);
+    }
+    if (event.data.type === SdkEvents.ProgressBarIncremented) {
+      emitter.emit(event.data.type, event.data);
+    }
+    if (event.data.type === SdkEvents.ProgressBarFinished) {
+      emitter.emit(event.data.type, event.data);
+    }
```
mateuszjasiuk, line 52: Those ifs are a bit ugly but I don't think we can make it better and preserver correct types 😅

mateuszjasiuk:
@euharrison can you take a look at last commit?  I added "off" and also added refetch so sync runs each time we go to masp tab 🙏

packages/shared/lib/Cargo.toml
```diff
@@ -17,6 +17,9 @@ multicore = ["rayon", "wasm-bindgen-rayon", "namada_sdk/multicore"]
 nodejs = []
 web = []
 
+[build-dependencies]
+namada_tx = { git = "https://github.com/anoma/namada", tag = "v0.45.1" }
```
jurevans, line 21: We should bump this to `v0.46.0` here - this may cause issues since this was recently merged: https://github.com/anoma/namada-interface/blob/f326f42fffba8a53e0dec8b440bb98dd907ec059/packages/shared/lib/Cargo.toml#L27

mateuszjasiuk: I will rebase and bump. thanks! :)

PR #1215, opened 2024-10-29
Namadillo: first version of "shield all" flow

pedrorezende:
This is the first stab on "Shield All" UI flow.


apps/namadillo/src/App/Ibc/ShieldAllAssetList.tsx
```diff
@@ -0,0 +1,65 @@
+import { Asset } from "@chain-registry/types";
+import { Checkbox } from "@namada/components";
+import { TokenCurrency } from "App/Common/TokenCurrency";
+import { AssetWithBalance } from "atoms/integrations";
+import BigNumber from "bignumber.js";
+import clsx from "clsx";
+import { getAssetImageUrl } from "integrations/utils";
+import { toDisplayAmount } from "utils";
+
+type AssetWithBalanceMap = Record<string, AssetWithBalance>;
```
emccorson, line 10: Maybe we could reuse this type since we also have `Record<string, AssetWithBalance>` in other places.

apps/namadillo/src/App/Ibc/ShieldAllAssetList.tsx
```diff
@@ -0,0 +1,65 @@
...
+
+type ShieldAllAssetListProps<T extends AssetWithBalanceMap> = {
```
emccorson, line 12: I think that in practice we won't get stronger typechecking by using `T extends AssetWithBalanceMap`, because we don't know the assets at compile time, so `T` will always be `AssetWithBalanceMap` instead of something more specific.

If you did want to ensure there is a `checked` entry for each asset, maybe something like dropping the `checked` prop and making `assets` be `(AssetWithBalance & { checked: boolean })[]` could work.

apps/namadillo/src/App/Ibc/ShieldAllContainer.tsx
```diff
@@ -0,0 +1,25 @@
+import clsx from "clsx";
+import { twMerge } from "tailwind-merge";
+type ShieldAllContainerProps = React.PropsWithChildren &
```
emccorson, line 3: Just a question, but what does `React.PropsWithChildren` do here?

emccorson (approved):
Looks great!

apps/namadillo/src/hooks/useAssetAmount.tsx
```diff
@@ -14,7 +14,8 @@ type useAmountTransferProps = {
 type UseAmountTransferOutput = {
   isLoading: boolean;
   balance: BigNumber | undefined;
-  availableAssets: Asset[];
+  availableAssets?: Asset[];
```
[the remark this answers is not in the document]
pedrorezende: I left `availableAssets` as optional because it will be easier to identify if their content has been loaded or not. But you're right, the way it was implemented wouldn't ever let this be undefined. Fixed!

apps/namadillo/src/App/Ibc/ShieldAllContainer.tsx
pedrorezende: I made a small change on this, but React.PropsWithChilren is a generic type that adds an optional `children` prop to the list of properties. It's another way to write `{children?: ReactNode} & otherTypeHere`

apps/namadillo/src/App/Ibc/ShieldAllAssetList.tsx
pedrorezende: Yeah, I think it will also improve readability. Made some changes on this! Thank you :)

PR #1216, opened 2024-10-29
feat: fully decode IBC assets on cosmos chains

emccorson:
If there are IBC assets on a Keplr account, attempt to look them up on the chain they came from and display them. For example, if there are NAM tokens on a Cosmos Hub account, display them as NAM.

Notes:
- This currently works for IBC assets from all cosmos chains and from the Namada internal devnet, but NOT any other Namada chains (e.g. housefire). Follow up issue: #1218
- I harcoded some registry entries temporarily, which should be removed once we've integrated the Namada registry. Follow up issue: #1217
- I don't know if this works for tokens that have been transferred more than once (e.g. Namada -> Osmosis -> Cosmos Hub). Follow up issue: #1219
- There is a bug where if you have the same asset from different chains (e.g. you sent NAM from both housefire and the internal devnet), only one of the balances will show. It is not so straightforward to fix, so follow up issue: #1220

### Testing

1. Configure Namadillo to use the internal devnet
2. Go to IBC Transfer -> From Namada and send some NAM to your Cosmos Hub testnet account (assuming Namadillo IBC features are enabled and Keplr is set up)
3. Go to IBC Transfer -> To Namada and select the Cosmos Hub testnet
4. In the asset list, you should see the tokens you sent listed as "Namada" and the denomination treated as NAM (e.g. if you have 1 NAM, it shows the balance as 1, not 1000000)

You can also try sending other IBC tokens (e.g. send some OSMO from Osmosis testnet to Cosmos Hub testnet) and it should also display properly.

### DIFF
--- a/apps/namadillo/src/atoms/integrations/functions.ts
+++ b/apps/namadillo/src/atoms/integrations/functions.ts
@@ -1,7 +1,8 @@
-import { Asset, AssetList, Chain } from "@chain-registry/types";
+import { Asset, Chain } from "@chain-registry/types";
 import { Coin } from "@cosmjs/launchpad";
 import { QueryClient, setupIbcExtension } from "@cosmjs/stargate";
 import { Tendermint34Client } from "@cosmjs/tendermint-rpc";
+import { mapUndefined } from "@namada/utils";
 import BigNumber from "bignumber.js";
 import * as celestia from "chain-registry/mainnet/celestia";
 import * as cosmos from "chain-registry/mainnet/cosmoshub";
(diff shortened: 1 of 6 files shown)


apps/namadillo/src/atoms/integrations/services.ts
```diff
@@ -52,9 +52,14 @@ const getShieldedArgs = async (
   };
 };
 
-export type AssetWithBalance = {
+export type AssetWithBalanceAndIbc = {
```
jurevans, line 55: I would maybe rename this to `AssetWithBalanceAndIbcDetails` or `AssetWithBalanceAndIbcInfo` or something that indicates what the `Ibc` part is :)

pedrorezende: Now it's possible to compose the definition of `AssetWithBalance` from `types.d.ts`

pedrorezende (approved):
LGTM! I would just review carefully the import I mentioned in one of the comments

apps/namadillo/src/atoms/integrations/functions.ts
```diff
@@ -55,55 +109,176 @@ const ibcAddressToDenom = async (
...
+    symbol: denom,
+  },
+});
+
+export const mapCoinsToAssets = async (
+  coins: Coin[],
+  chainName: string,
+  rpc: string
+): Promise<Record<string, AssetWithBalanceAndIbc>> => {
+  const chain = cosmosRegistry.chains.find(
```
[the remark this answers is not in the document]
emccorson: That's a good point. It looks like the JS bundle goes from ~5kB (which Vite is already warning us is big) to ~8kB. I was thinking a bit about how we want to access the registries anyway, but I don't know the best solution so I wrote it up in #1224 and I think I will improve this in a later PR.

apps/namadillo/src/atoms/integrations/services.ts
emccorson: Sounds good! I will rename it.

emccorson:
This is merged (I just messed up getting GitHub to realise that, so it think it's closed).

PR #1225, opened 2024-10-31
IBC: Fix "Not enough balance" bugs + use display denom for amounts

### LINKED ISSUE #1211: IBC: Decide how to represent amounts (base denom/display denom/something else)
When doing IBC deposits into Namada, we represent amounts in their base denom. For example, 1 ATOM is represented as `1000000` (uatom). This corresponds with the denomination we have to pass into cosmjs when performing the transfer.

However, when doing IBC withdrawals out of Namada, we represent amounts in their display denom. For example, 1 NAM is represented as `1` (equivalent to 1,000,000 namnam). This corresponds with the denomination we have to pass into Namada SDK when performing the transfer.

The problem is that we have the `TransferModule` component, which handles both IBC deposits and IBC withdrawals, and sometimes the values we want to pass to it or receive from it are in the base denom and sometimes in the display denom.

We should decide how to deal with this in a concrete way (instead of converting ad hoc like I'm doing in #1210) so that we don't create a bug where we send a user's token in the wrong denomination (which could be disastrous).

Other things to note:
- Namada SDK seems a bit weird about handling denoms. It seems to accept `DenominatedAmount` when creating an IBC transfer, which should let us set the denomination we want, but then seems to throw it away and assume the denomination is NAM.
- We receive NAM balances in NAM from the indexer (I think) and use the NAM denomination pretty much exclusively everywhere else in Namadillo. We use the `BigNumber` type because it allows us to hold very big numbers in NAM. I was the one who initially worked on figuring out how we should represent amounts, and I decided it was easier to store amounts in NAM using `BigNumber` than to store amounts in namnam using `BigInt` and converting for display.

emccorson:
This PR fixes two bugs that were causing the TransferModule to incorrectly detect that there is not enough balance when the user is performing an IBC transaction. It also changes all amounts in Namadillo to be represented in the display denom (which is what we normally do for non-IBC things), converting to/from the base denom only when we receive or send amounts in the base denom (e.g. when submitting an IBC transfer to cosmjs).

The commits are fairly separate so it might be easier to review commit by commit.

Closes #1211.

---

### Changed
- Represent all amount values in display denom
  - This changes all amounts to be represented in their display denom throughout Namadilllo, converting at the edges when we receive or send amounts represented in the base denom
  - Fixes the first of two bugs causing IBC transfer to incorrectly detect that there is not enough balance
- Make TransferModule validation more DRY

### Fixed
- Stop deducting fees twice in available amount
  - Fixes the second of two bugs causing IBC transfer to incorrectly detect that there is not enough balance
- Display 0 if available amount minus fees is less than 0

### DIFF
--- a/apps/namadillo/src/App/Ibc/IbcTransfer.tsx
+++ b/apps/namadillo/src/App/Ibc/IbcTransfer.tsx
@@ -132,7 +132,7 @@ export const IbcTransfer: React.FC = () => {
             sourceAddress,
             destinationAddress,
             amount,
-            token: selectedAsset.base,
+            asset: selectedAsset,
             transactionFee,
             sourceChannelId: ibcOptions.sourceChannel,
             ...(shielded ?
--- a/apps/namadillo/src/App/Ibc/IbcWithdraw.tsx
+++ b/apps/namadillo/src/App/Ibc/IbcWithdraw.tsx
@@ -32,21 +32,15 @@ export const IbcWithdraw: React.FC = () => {
 
   const [selectedAsset, setSelectedAsset] = useState<Asset>();
 
-  const namBalance = useAtomValue(accountBalanceAtom).data;
-
   // TODO: remove hardcoding and display assets other than NAM
+  const availableAmount = useAtomValue(accountBalanceAtom).data;
   const availableAssets = [namadaAsset];
-  // TODO: TransferModule expects amounts in namnam, but Namada SDK expects
-  // amounts in NAM. We should figure out how to deal with this properly.
-  const availableAmount = namBalance?.shiftedBy(6);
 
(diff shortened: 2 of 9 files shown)


euharrison (approved):
I was able to do the ibc transfer to transparent and shielded accounts without the balance issue 👍

jurevans (approved):
Nice! This is working for deposits! I tried sending NAM to cosmos testnet, got an error `Cannot read properties of undefined (reading 'sign')`, but that's likely not related

emccorson:
> Nice! This is working for deposits! I tried sending NAM to cosmos testnet, got an error `Cannot read properties of undefined (reading 'sign')`, but that's likely not related

I haven't seen that error before :thinking: The IBC withdraw code is still basically the prototype code so maybe it's to do with that.

Issue #1172, opened 2024-10-09
Update the Account Overview for phase 3


euharrison:
We need to implement the layout below for the account overview http://localhost:5173/

It's ok to split into multiple PRs and new issues in case this is a big lift

![Screenshot 2024-10-09 at 18 57 21](https://github.com/user-attachments/assets/035df275-474f-4003-98ff-bc046a7d9bc3)

PR #1227, opened 2024-11-01
feat: render dollar values

euharrison:
Render dollar values on Masp Overview and the Account Overview

We are using the https://sqs.osmosis.zone/swagger/index.html api for the price conversion

Screenshot 2024-11-05 at 01 49 44

Screenshot 2024-11-05 at 01 49 38

Closes https://github.com/anoma/namada-interface/issues/1171
Ref https://github.com/anoma/namada-interface/issues/1172


apps/namadillo/src/atoms/balance/atoms.ts
```diff
@@ -0,0 +1,230 @@
...
+  const availableAssets = get(availableAssetsAtom);
+  const assetsByDenom: Record<string, Asset> = {
+    // TODO namAsset should be returned from availableAssetsAtom
+    [NAM_DENOM]: namadaAsset,
+  };
+  availableAssets.forEach((assets) => {
+    assets.assets.forEach((asset) => {
+      asset.denom_units.forEach((unit) => {
+        // only set the asset if it doesn't exists
+        // otherwise the testnet will override the mainnet
```
emccorson, line 117: I just realised both this PR and #1228 are trying to decode assets from their IBC trace, but in different ways. Maybe we can use just one way, but we'll need to decide which way.

Like you said in this comment, if the asset exists on multiple chains, we will only use the first entry, but the problem with this is that different chains have different configurations for the same asset (and not just testnets).

I'll write it up a bit more in a message so we can discuss it.

emccorson (commented):
I've only reviewed some of this PR, but stopping for now so we can figure out the asset decoding thing I mentioned in the comments, and then I'll finish it!

apps/namadillo/src/atoms/accounts/services.ts
```diff
@@ -38,3 +38,12 @@ export const fetchAccountBalance = async (
       BigNumber(balance.amount).shiftedBy(-decimals)
     : BigNumber(0);
 };
+
+export const fetchAccountBalance = async (
+  api: DefaultApi,
+  account: Account | undefined
+): Promise<Balance[]> => {
+  if (!account) return [];
+  const balancesResponse = await api.apiV1AccountAddressGet(account.address);
```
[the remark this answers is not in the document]
euharrison: @emccorson I didn't update here because the current `namadaAsset.base` is `namnam`, and it'll create a cascade of issues. In case you prefer to change the `namadaAsset.base` to be `nam`, I can shift this value returned from the indexer api :)

apps/namadillo/src/atoms/balance/functions.ts
```diff
@@ -0,0 +1,58 @@
+import { IbcToken, NativeToken } from "@anomaorg/namada-indexer-client";
+import { Asset, AssetList } from "@chain-registry/types";
+import BigNumber from "bignumber.js";
+import { namadaAsset } from "registry/namadaAsset";
+import { TokenBalance } from "./atoms";
+
+// TODO upgrade this function to be as smart as possible
+// please refer to `tryCoinToIbcAsset` and how we could combine both
+export const findAssetByToken = (
```
emccorson, line 9: It seems that this function takes an IBC trace such as `transfer/channel-0/uatom` and checks if there is any known asset with the same trace on any chain. But I think this only works by coincidence because there happens to be a `channel-0` on Osmosis with counterparty Cosmoshub, and for other tokens we won't find anything. Or in other words, Namada does have a `channel-0` (corresponding to Cosmoshub testnet) but this is a different thing from `channel-0` on Osmosis (corresponding to Cosmoshub mainnet).

But I think we can leave this as is for now, and I can integrate the two different methods once #1228 is merged.

emccorson (approved):
I wasn't able to fully test this (possibly due to IBC channels or something), but looks good! I can follow up with the integration of the two different IBC asset decoding methods once the other PR is merged.

apps/namadillo/src/atoms/balance/functions.ts
euharrison: Thanks! Exactly this! 🚀

PR #1232, opened 2024-11-04
feat/1231: Extension - Generate shielded keys from private key import

### LINKED ISSUE #1231: Extension: Importing private key should also generate shielded accounts
We now have the ability to derive shielded keys from private keys in the SDK package.

When a user imports just a private key:

- In addition to the imported transparent address/public key, also derive a spending key and default viewing key/payment address
- Add this to the confirmation UI, as well as `View Keys` (the shielded key should reference the ID of the transparent SK account as a parent)

jurevans:
Resolves #1231 

- [x] When user imports private key, also derive shielded keys (default path)
- [x] Ensure any changes to custom path form are ignored for these
- [x] Fixed issue with Zip32 derivation: As the shielded keys are derived from a private key (which was derived with either a default or custom path), shielded keys on import should be derived with the default Zip32 path `/0'` - in the future we can allow users to derive additional keys either by incrementing (`/1'`) or providing a full path (e.g., `/0'/1`, etc.)

The confirmation page and `View Keys` views are automatically updated, as the private key account (which seeds the shielded keys) is also treated as a parent.

### Testing

- Start with a new extension install (Optional)
- Import or generate Mnemonic account (ensure existing functionality remains)
- Import another account from Private Key (can use: `00bebb118517e09c950841d849eb3e07e27e6cb9c59a8e71179220fa570d5d770f`) - should receive a shielded address
- Import mnemonic with custom path (existing functionality)
- Delete private key account, and then `Add Keys` - In the Import Mnemonic page under `Advanced`, set a custom path, then go to `Private Key` and import a private key as usual - this is to ensure that the custom path you entered is ignored when you import from private key (you shouldn't see a custom path here, because it is a default path)

Following all of this, it's good to check in the console with `await namada.accounts()` and inspect the accounts, ensuring:
- Shielded accounts' parent IDs line up with their parents
- The derivation paths are correct

### Screenshots

Private-key import confirmation:
Screen Shot 2024-11-04 at 12 09 30 PM

View Keys
Screen Shot 2024-11-04 at 12 09 46 PM

### DIFF
--- a/apps/extension/src/App/Accounts/ViewingKey.tsx
+++ b/apps/extension/src/App/Accounts/ViewingKey.tsx
@@ -15,7 +15,7 @@ export const ViewingKey = (): JSX.Element => {
         <PageHeader title="Viewing Key" />
         <p className="text-white">
           Your viewing key grants the holder access to all your balances and
-          transadction data. Please keep it secure to protect your data.
+          transaction data. Please keep it secure to protect your data.
         </p>
         <Input
           label="Viewing Key"
--- a/apps/extension/src/Setup/Common/Completion.tsx
+++ b/apps/extension/src/Setup/Common/Completion.tsx
@@ -106,16 +106,16 @@ export const Completion: React.FC<Props> = (props) => {
 
         // Do not derive shielded if this is an imported private key, and
         // ignore accounts with a non-zero 'change' path component:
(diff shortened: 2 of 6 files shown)


euharrison (approved):
✅

mateuszjasiuk (approved):
LGTM!

PR #1235, opened 2024-11-07
fix: unify asset decoding for Namada and Cosmos

emccorson:
This PR makes sure we use the same method for translating addresses to assets for both Namada and Cosmos chains. It follows on from #1227 and #1228.

This will only work with the internal devnet, but I'll add a PR tomorrow to get it working with housefire too.

I think we still should refactor more, because it is quite confusing to have `namadaTransparentAssetsAtom` (turning Namada addresses into assets with balances), and `transparentTokensAtom` (returning the same thing but also with balances in dollars). I've left the refactoring for now because my goal was just to make sure we use the same code for Namada and Cosmos. There are also other things that could be refactored (e.g. turning BigNumbers back into strings, using `Coin` type even though it is basically the same as the `Balance` type).

---

### Changed
- Store addresses for all assets, replacing IBC-specific address entry

### Fixed
- Unify asset decoding for Namada and Cosmos
  - Fixes cases where Namada addresses are decoded to the wrong chain's
    asset
- Add NAM address to registry and remove special handling for Namada
- Remove unused chain lookup in mapCoinsToAssets

### DIFF
--- a/apps/namadillo/src/atoms/balance/atoms.ts
+++ b/apps/namadillo/src/atoms/balance/atoms.ts
@@ -4,21 +4,26 @@ import {
   defaultAccountAtom,
   transparentBalanceAtom,
 } from "atoms/accounts/atoms";
-import { nativeTokenAddressAtom, tokenAddressesAtom } from "atoms/chain";
+import {
+  chainParametersAtom,
+  nativeTokenAddressAtom,
+  tokenAddressesAtom,
+} from "atoms/chain";
 import { shouldUpdateBalanceAtom } from "atoms/etc";
 import { availableAssetsAtom } from "atoms/integrations";
 import { queryDependentFn } from "atoms/utils";
 import BigNumber from "bignumber.js";
 import { atomWithQuery } from "jotai-tanstack-query";
 import { namadaAsset } from "registry/namadaAsset";
-import { unknownAsset } from "registry/unknownAsset";
-import { toDisplayAmount } from "utils";
-import { findAssetByToken } from "./functions";
+import { AddressWithAssetAndBalance } from "types";
+import {
+  findAssetByToken,
+  mapNamadaAddressesToAssets,
+  mapNamadaAssetsToTokenBalances,
+} from "./functions";
 import { fetchCoinPrices, fetchShieldedBalance } from "./services";
 
-export type TokenBalance = {
-  asset: Asset;
-  balance: BigNumber;
+export type TokenBalance = AddressWithAssetAndBalance & {
   dollar?: BigNumber;
 };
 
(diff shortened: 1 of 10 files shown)


apps/namadillo/src/registry/namadaAsset.ts
```diff
@@ -3,13 +3,14 @@ import namadaShieldedSvg from "./assets/namada-shielded.svg";
 
 export const namadaAsset: Asset = {
   name: "Namada",
-  base: "namnam",
+  base: "tnam1qy440ynh9fwrx8aewjvvmu38zxqgukgc259fzp6h",
```
euharrison, line 6: I think we cannot use this `tnam...` address because in the future it'll be replaced by the `namada-chain-registry`

I understand you explained that you will have another PR to support the housefire chain, which I believe you will update this again, but can you confirm this is on your plan before merging?

apps/namadillo/src/atoms/balance/atoms.ts
```diff
@@ -148,75 +153,84 @@ export const tokenPricesAtom = atomWithQuery((get) => {
...
-  const balance = toDisplayAmount(asset, amount);
+export const shieldedTokensAtom = atomWithQuery<TokenBalance[]>((get) => {
+  const shieldedAssets = get(namadaShieldedAssetsAtom);
+  const tokenPrices = get(tokenPricesAtom);
+
+  return {
+    queryKey: ["shielded-tokens", shieldedAssets.data, tokenPrices.data],
+    ...queryDependentFn(
+      async () =>
+        await mapNamadaAssetsToTokenBalances(
```
euharrison, line 212: This is not an async operation, it's just a sync transformation. Then maybe we should have if as a helper function than an atom, like we do with `getTotalDollar`. What do you think?

apps/namadillo/src/atoms/balance/functions.ts
```diff
@@ -56,3 +59,57 @@ export const getTotalDollar = (list?: TokenBalance[]): BigNumber | undefined =>
...
+
+const tnamAddressToDenomTrace = (
+  address: string,
+  tokenAddresses: (NativeToken | IbcToken)[]
+): DenomTrace | undefined => {
+  const token = tokenAddresses.find((entry) => entry.address === address);
+  const trace = token && "trace" in token ? token.trace : undefined;
+
+  if (typeof trace === "undefined") {
+    return undefined;
```
euharrison, line 71: For this case, we can return the `nam` denom (or `namnam`), because it'll be the `NativeToken`. 

@mateuszjasiuk just in case, can you confirm this again?
On the indexer api, the endpoint `/token`, if a token doesn't have a `trace` value it means that it's the native token, meaning the `NAM` token, correct?

apps/namadillo/src/atoms/balance/functions.ts
```diff
@@ -56,3 +59,57 @@ export const getTotalDollar = (list?: TokenBalance[]): BigNumber | undefined =>
...
+  }
+
+  const separatorIndex = trace.lastIndexOf("/");
+
+  if (separatorIndex === -1) {
+    throw new Error(`Couldn't get denom trace from ${trace}`);
```
euharrison, line 77: isn't it better to return `undefined` here? I'm afraid of breaking the whole interface because we can't find a asset. We can use the `unknowAsset` for this case as well, just to render something

apps/namadillo/src/atoms/integrations/functions.ts
```diff
@@ -168,21 +158,17 @@ const findCounterpartChainName = (
...
-  rpc: string,
+  ibcAddressToDenomTrace: (address: string) => Promise<DenomTrace | undefined>,
   chainName: string
-): Promise<AssetAndIbc | undefined> => {
+): Promise<Asset | undefined> => {
   const { denom } = coin;
 
-  if (!denom.startsWith("ibc/")) {
+  const denomTrace = await ibcAddressToDenomTrace(denom);
+  if (typeof denomTrace === "undefined") {
```
euharrison, line 167: cosmetic, we can use `if (denomTrace === undefined)` without the `typeof` :)

apps/namadillo/src/atoms/integrations/functions.ts
```diff
@@ -213,96 +199,62 @@ const tryCoinToIbcAsset = async (
...
-        denom,
-        exponent: 0,
-      },
-    ],
-    base: denom,
-    name: denom,
-    display: denom,
-    symbol: denom,
-  },
+const unknownAsset = (denom: string): Asset => ({
```
euharrison, line 209: Can you check the `apps/namadillo/src/registry/unknownAsset.ts`, maybe we can combine both.

Particularly, I like this idea of having a function which we can at least render the denom instead of `?`

apps/namadillo/src/registry/namadaAsset.ts
```diff
@@ -3,13 +3,14 @@ import namadaShieldedSvg from "./assets/namada-shielded.svg";
...
   display: "nam",
   symbol: "NAM",
   denom_units: [
     {
-      denom: "namnam",
+      denom: "tnam1qy440ynh9fwrx8aewjvvmu38zxqgukgc259fzp6h",
       exponent: 0,
+      aliases: ["unam"],
```
euharrison, line 13: Is it `unam` or `namnam`? do we need to ask someone else for this answer?

apps/namadillo/src/atoms/integrations/functions.ts
emccorson: I started doing this based on a previous PR comment, but I think the difference in behaviour is pretty minor:
https://github.com/anoma/namada-interface/pull/619#discussion_r1481013086

apps/namadillo/src/registry/namadaAsset.ts
emccorson: I always thought this was namnam, but I think the name must have changed sometime because Brent corrected it to unam in https://github.com/anoma/namada-chain-registry/pull/1#discussion_r1820076376. I will check this.

apps/namadillo/src/registry/namadaAsset.ts
emccorson: Sorry, that's my bad that I didn't write the reason for this anywhere.

Since on Namada chains we need to do an extra step to decode the `tnam...` address to an asset, I added this here. Alternatively we can do this using the indexer, but this only works for the Namada chain configured in Namadillo. So if Namadillo is configured for the internal devnet, but our account contains some NAM originating in housefire, then we need a way to look up the `tnam...` addresses for housefire tokens, which means we either need an indexer for each Namada chain, or we can just store the address in the chain config, which I think is easier.

My plan is to add the tnam address to the Namada registry for each Namada chain so this will work for all Namada chains.

apps/namadillo/src/atoms/integrations/functions.ts
emccorson: That sounds good! I think `unknownAsset.ts` is now not used anywhere. Is there a way you'd prefer to combine them i.e. which properties to keep from each type?

apps/namadillo/src/atoms/balance/atoms.ts
emccorson: Nice catch! I was thinking if we do make it a helper function, we'll need to copy the code to get `shieldedAssets` and `tokenPrices` everywhere it is used, so since it's two args it might be a bit less nice than `getTotalDollar`.

apps/namadillo/src/atoms/balance/functions.ts
emccorson: In this case, it should return specifically only for IBC tokens because it's used as part of the IBC asset lookup. I'll add a comment to make this clearer.

apps/namadillo/src/atoms/balance/functions.ts
emccorson: Fixed!

apps/namadillo/src/atoms/balance/functions.ts
mateuszjasiuk: It should mean native token. I will try to figure out why on housefire we have multiple tokens without an ibc trace. If you want to be double sure you can always get native token address from parameters and compare :)

apps/namadillo/src/registry/namadaAsset.ts
euharrison: We have the prop `address` for this case, maybe it's helpful. Do you mind giving it a try?

Screenshot 2024-11-08 at 11 59 15

apps/namadillo/src/atoms/integrations/functions.ts
euharrison: cool, no preferences, this is just to render anything for border cases that should never be used

euharrison (approved):
I'm still thinking about the `"base": "tnam..."` stuff, but this should not be a blocker for the PR :)

apps/namadillo/src/registry/namadaAsset.ts
emccorson: Oh nice! I didn't know about that. That's much better!

emccorson: I'm thinking I will merge this as is, and then next up is the chain registry integration, so I will fix the `tnam` thing there.

PR #1240, opened 2024-11-11
feat: use namada registry + IBC decoding fixes

### LINKED ISSUE #1217: Namadillo: Integrate Namada chain registry
Following on from #1216, we should add namada-chain-registry as a dependency and removed the hardcoded registry entries in Namadillo. Blocked by anoma/namada-chain-registry#2.

### LINKED ISSUE #1220: Namadillo: IBC -> Allow decoding the same asset from multiple chains
Following on from #1216, there is a bug where if you have the same asset from multiple different chains (e.g. NAM from internal devnet and from housefire), only one will be shown. The problem is that we index the asset balances object by base denom, but this will have multiple entries, so really we should key by base denom + IBC denom trace path. Fix this.

emccorson:
This PR integrates Namadillo with Namada registry, as well as fixing some other things related to decoding addresses to assets. This adds support for holding the same asset from multiple chains. To test, you can transfer some NAM from the internal devnet to cosmoshub testnet to housefire, and you should see both your housefire NAM and your internal devnet NAM listed on your account.

The package.json points to a specific branch of namada-chain-registry, but I can update this as soon as anoma/namada-chain-registry#3 is merged. The code integrating the registry is a bit clunky, but this can be improved as the registry develops.

---

### Added
- Use Namada chain registry. Closes #1217

### Fixed
- Index assets by their original address, not by base denom. Fixes #1220
  - Also refactor TransferModule accordingly
- Look up registry asset by base denom, alias or address
- Use address to index assets in 'Select Asset' list

### DIFF
--- a/apps/namadillo/package.json
+++ b/apps/namadillo/package.json
@@ -27,6 +27,7 @@
     "jotai": "^2.6.3",
     "jotai-tanstack-query": "^0.8.5",
     "lodash.debounce": "^4.0.8",
+    "namada-chain-registry": "https://github.com/anoma/namada-chain-registry",
     "react": "^18.3.1",
     "react-dom": "^18.3.1",
     "react-icons": "^5.1.0",
--- a/apps/namadillo/src/integrations/utils.ts
+++ b/apps/namadillo/src/integrations/utils.ts
@@ -44,7 +44,8 @@ export const getTransactionFee = (
     if (typeof asset !== "undefined") {
       return {
         amount: BigNumber(0.003), // TODO: remove hardcoding
-        token: asset,
+        asset,
+        originalAddress: asset.base,
       };
     }
   }
--- a/apps/namadillo/src/types.d.ts
+++ b/apps/namadillo/src/types.d.ts
@@ -189,13 +189,16 @@ export type ChainRegistryEntry = {
   ibc: IBCInfo[];
 };
 
-export type AddressWithAssetAndBalance = {
-  address: string;
+export type AddressWithAsset = {
+  originalAddress: Address;
   asset: Asset;
-  balance: BigNumber;
 };
 
-export type AddressWithAssetAndBalanceMap = Record<
-  string,
-  AddressWithAssetAndBalance
+export type AddressWithAssetAndAmount = AddressWithAsset & {
+  amount: BigNumber;
+};
+
+export type AddressWithAssetAndAmountMap = Record<
+  Address,
+  AddressWithAssetAndAmount
 >;
--- a/apps/namadillo/src/App/Ibc/IbcTransfer.tsx
+++ b/apps/namadillo/src/App/Ibc/IbcTransfer.tsx
@@ -1,4 +1,4 @@
-import { Asset, Chain } from "@chain-registry/types";
+import { Chain } from "@chain-registry/types";
 import { Window as KeplrWindow } from "@keplr-wallet/types";
 import { mapUndefined } from "@namada/utils";
 import { TransactionTimeline } from "App/Common/TransactionTimeline";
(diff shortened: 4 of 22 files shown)


apps/namadillo/package.json
```diff
@@ -27,6 +27,7 @@
     "jotai": "^2.6.3",
     "jotai-tanstack-query": "^0.8.5",
     "lodash.debounce": "^4.0.8",
+    "namada-chain-registry": "https://github.com/anoma/namada-chain-registry#add-housefire-ibc",
```
euharrison, line 30: can you confirm it'll work fine after the PR is merged and the branch is deleted? Maybe we should use the commit hash here

apps/namadillo/src/types.d.ts
```diff
@@ -189,13 +189,16 @@ export type ChainRegistryEntry = {
   ibc: IBCInfo[];
 };
 
-export type AddressWithAssetAndBalance = {
-  address: string;
+export type AddressWithAsset = {
```
euharrison, line 192: To simplify some names, I think we could call this as `Token`, wdyt?

My reference is the `tokens.toml` files that is used by the cli, that is basically the `tnam...` address plus some properties

euharrison (approved):

pedrorezende (approved):
Amazing PR! Just a few comments on variables / property names! Great work 🚀

apps/namadillo/package.json
emccorson: Updated now to point to the main branch now that my PR there is merged.

Just out of interest, I checked the `yarn.lock` and it seems like it does point to a commit. But if the branch does get deleted I have no idea if GitHub keeps track of the commit.

apps/namadillo/src/types.d.ts
emccorson: I definitely agree the names I used are not very good. I'm not sure about `Token` since I think we have other things named similar to Token (but maybe those are not needed now, I have to check). `AssetRegistry` makes me think more of the full registry, like cosmos/chain-registry.

For now, I created #1244 to track this and I think it should be fairly easy to change the names when we settle on something.

apps/namadillo/src/integrations/utils.ts
```diff
@@ -44,7 +44,8 @@ export const getTransactionFee = (
     if (typeof asset !== "undefined") {
       return {
         amount: BigNumber(0.003), // TODO: remove hardcoding
-        token: asset,
+        asset,
+        address: asset.base,
```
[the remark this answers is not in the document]
emccorson: I have added this to #1244 for tracking. I think one problem is that `denom`, `base` and `address` all exist on the `Asset` type, but what we are calling the address might actually be none of those things. The intention of the property is to be "the thing we pass to Namada/Cosmos SDK to tell it what to transfer", but I have no idea what to call that :smile:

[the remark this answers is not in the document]
emccorson: For now I'll make this `originalAddress` so it's hopefully less confusing, but this can just be a placeholder until we get something better.

apps/namadillo/src/types.d.ts
emccorson: Some ChatGPT recommendations:

1. AssetRecord (address and asset only) and AssetAmountRecord (address, asset, and amount)
2. AssetReference (address and asset only) and AssetAmountEntry (address, asset, and amount)
3. ResolvedAsset (address and asset only) and ResolvedAssetAmount (address, asset, and amount)
4. TokenEntry (address and asset only) and TokenAmountEntry (address, asset, and amount)

I'm not sure about these since they don't indicate that there is an original address, but maybe they're OK.

PR #1243, opened 2024-11-12
feat: enable IBC withdrawals for all tokens

### LINKED ISSUE #1222: Namadillo: IBC -> IBC withdraw should reveal public key if needed
IBC withdraw needs to submit a RevealPK if the public key hasn't been revealed, and will fail otherwise. Do it the same way as the other transactions do it.

### LINKED ISSUE #1242: Namadillo: IBC -> Display all available tokens for withdraw and remove prototype code
IBC Withdrawal is still basically prototype code. We should fix this up.

It also only shows NAM even when there are other available tokens on the account. Allow withdrawals of all tokens.

emccorson:
This PR allows IBC withdrawals for all assets, instead of just NAM.

There is still zero indication to the user about whether the transaction failed or succeeded, but for testing this information should be displayed in the console.

---

### Changed
- Use base denom amounts when building IBC transfers with JS SDK
- Use unknown gas limit value for txs without an indexer type

### Added
- Enable IBC withdrawals for all tokens. Closes #1242

### Fixed
- Use same gas limit and price for IBC withdrawal as rest of Namadillo
- Reveal public key if needed for IBC withdrawal. Fixes #1222

### DIFF
--- a/apps/namadillo/src/atoms/fees/services.ts
+++ b/apps/namadillo/src/atoms/fees/services.ts
@@ -1,24 +1,27 @@
 import { DefaultApi } from "@namada/indexer-client";
+import { mapUndefined } from "@namada/utils";
 import BigNumber from "bignumber.js";
 import invariant from "invariant";
 import { GasTable } from "types";
-import { txKindFromIndexer } from "./atoms";
+import { txKinds } from "types/txKind";
+import { txKindToIndexer } from "./functions";
 
 export const fetchGasLimit = async (api: DefaultApi): Promise<GasTable> => {
   const gasTableResponse = await api.apiV1GasGet();
-  const gasTable = gasTableResponse.data.reduce(
-    (acc, { gasLimit, txKind: indexerTxKind }) => {
-      const txKind = txKindFromIndexer(indexerTxKind);
-      if (txKind) {
-        const perKind = acc[txKind] || {};
-        acc[txKind] = { ...perKind, native: new BigNumber(gasLimit) };
-      }
-      return acc;
-    },
-    {} as GasTable
-  );
 
-  return gasTable;
+  return txKinds.reduce((acc, txKind) => {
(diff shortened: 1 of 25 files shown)


apps/namadillo/src/types/txKind.ts
```diff
@@ -0,0 +1,13 @@
+export const txKinds = [
```
pedrorezende, line 1: I think we need to discuss how we're going to organize the types in the codebase. One advantage of having them in a root folder is that is easy to find, import, and maintain. Do you think we should break the types in different files?

apps/namadillo/src/atoms/fees/services.ts
```diff
@@ -1,24 +1,53 @@
...
+  GasLimitTableInnerTxKindEnum as GasLimitTableIndexer,
+} from "@anomaorg/namada-indexer-client";
+import { assertNever, mapUndefined } from "@namada/utils";
 import BigNumber from "bignumber.js";
 import invariant from "invariant";
 import { GasTable } from "types";
-import { txKindFromIndexer } from "./atoms";
+import { TxKind, txKinds } from "types/txKind";
+
+const txKindToIndexer = (txKind: TxKind): GasLimitTableIndexer => {
```
pedrorezende, line 11: Nitpick: maybe this should be in `functions.ts` instead of services?

pedrorezende (approved):
LGTM!

apps/namadillo/src/atoms/fees/services.ts
emccorson: Fixed!

apps/namadillo/src/types/txKind.ts
emccorson: I actually did this just because I needed a `const` and I couldn't do that in `types.d.ts`. Now that your PR changes it to `types.ts`, we could still use that but I'll leave that until we discuss how to handle it.

PR #1249, opened 2024-11-12
feat: add components to shield and unshield

euharrison:
Add components to shield and unshield

To test, got to:
- http://localhost:5173/masp/shield
- http://localhost:5173/masp/unshield

You can observe that we have the feature os passing the query param `asset` as well, like:
- http://localhost:5173/masp/shield?asset=tnam1p5nnjnasjtfwen2kzg78fumwfs0eycqpecuc2jwz
- http://localhost:5173/masp/shield?asset=tnam1qy440ynh9fwrx8aewjvvmu38zxqgukgc259fzp6h

This is important for the `AccountOverview` and `MaspOverview` buttons that redirects user with a pre-selected asset

**Note1**: The transaction and the extension connection will be handled in another PR

**Note2**: There are small differences between the implementation and the Figma. We can handle it later, this is just to move faster with the implementation

Screenshot 2024-11-12 at 18 37 22

Screenshot 2024-11-12 at 18 37 42


pedrorezende (approved):
Approving this PR. We need to come back to these pages after merging other PRs that will change the base structure. But we can get this in so we have access to different features in the interface

PR #1255, opened 2024-11-13
feat/1251: Fetch applications feature flags from registry

### LINKED ISSUE #1251: Namadillo: Fetch feature flags asynchronously
More info _TBD_ - Requires an update to the registry to add schema and settings for "features"

Example features value - all enabled:

```
"features": [
    "claimRewards",
    "masp",
    "ibcTransfers",
    "ibcShielding",
    "shieldingRewards",
    "namTransfers"
  ],
```

jurevans:
Resolves #1251 

This PR updates the application features to now be loaded async directly from the registry:

https://github.com/anoma/namada-chain-registry

Example flags to be read:

https://github.com/anoma/namada-chain-registry/blob/b5ac94e0c3e8e091b0668f005d90920fc64b12b1/namadahousefire/chain.json#L14-L21

## DEMO

https://namada-interface-dev.netlify.app/staking
- This is configured for `namada-dryrun.abaaeaf7b78cb3ac`, but it's good to also point to Housefire to test feature states for that as well

## TESTING

- Run `namadillo` with `yarn dev:proxy` - needed to access Github registry (we need to also test this from Netlfiy, I wonder if CORS with Github is an issue there too...)
- Probably a good idea to delete any local storage, then specify  as the indexer - Only `Claim Staking` should be enabled
- Delete local storage, point to Housefire Indexer: `https://indexer.knowable.run/`, reload if necessary, all features should be enabled!
- Point to a local indexer or devnet that is not in our registry, all features should be enabled

### DIFF
--- a/apps/namadillo/scripts/proxies.json
+++ b/apps/namadillo/scripts/proxies.json
@@ -1,7 +1,7 @@
 [
   {
-    "alias": "MASP Params",
-    "url": "https://github.com/anoma/masp-mpc/releases/download/namada-trusted-setup/",
+    "alias": "Namada Chain Registry",
+    "url": "https://raw.githubusercontent.com/anoma/namada-chain-registry/refs/heads/main/",
     "proxyPort": 8010
   }
 ]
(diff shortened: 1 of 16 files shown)


euharrison (approved):
awesome!

PR #1257, opened 2024-11-14
feat: auto-fill IBC channel inputs

### LINKED ISSUE #1241: Namadillo: IBC -> Auto-detect which channel IDs to use
We currently display channel ID input boxes for IBC transfers, but we can just look up the IBC information in the registry to automatically set the channel IDs.

emccorson:
This PR makes the IBC channels on the IBC Transfer and IBC Withdraw pages be auto-filled when there is a known IBC config, but still supports manually changing the inputs.

Closes #1241.

### DIFF
--- a/apps/namadillo/src/App/Ibc/IbcTransfer.tsx
+++ b/apps/namadillo/src/App/Ibc/IbcTransfer.tsx
@@ -11,6 +11,7 @@ import {
   assetBalanceAtomFamily,
   availableChainsAtom,
   chainRegistryAtom,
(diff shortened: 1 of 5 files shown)


euharrison (approved):
Nice!

Currently, it's working only with the cosmos-testnet, but I believe this is missing on the registry, correct?

Screenshot 2024-11-14 at 16 57 25

Screenshot 2024-11-14 at 16 57 33

jurevans (approved):
LGTM!

PR #1278, opened 2024-11-19
feat: add the transfer form

euharrison:
Add the Transfer form, aka  `<NamadaTransfer>`

There are several visual tweaks that still need to be done, but they can be handled on future PRs

Upgraded the `TransferModule` state moving the  `shielded`,  `amount` and `customAddress` to the parent component implementation. This is necessary because the `<NamadaTransfer>` requires external control over theses props

https://github.com/user-attachments/assets/0cf571b6-27e8-46cf-9dcf-ce187a84bb6b


pedrorezende (approved):
I made a very quick review on this one so we can push this to the community. We can open new issues later this week if anything seems not quite right, but in general LGTM!

PR #1281, opened 2024-11-20
fix: fix up IBC amount input

### LINKED ISSUE #1271: Namadillo: Improve Amount field on IBC
To use the Amount field on IBC Transfer, currently we either have to append to the `0`, or erase it, it does not clear on its own.

The Amount input allows the user to do things like this:

Screen Shot 2024-11-18 at 1 21 38 PM

emccorson:
Fixes #1271.

- Use empty input with placeholder instead of default 0 value
- Check for max decimal places allowed by selected asset

I added the placeholder because otherwise you can't see the amount input when it's empty, but this is just a guess and not in the designs, so @ChrisHoltDesign may want this changed.


emccorson:
> LGTM! Not sure if part of this PR, but you still can type values like "000123" :)

I think that will apply anywhere we have an AmountInput, but maybe that could be fixed in a future PR!

PR #1282, opened 2024-11-20
fix: robuster way to detect/install keplr (#1267)

### LINKED ISSUE #1267: Namadillo: keplr sometimes is not detected
After trying to transfer assets via IBC transfer panel, sometimes keplr is not detected and the transfer timeline gets stuck in the "Signature" stage. This issue was commonly found in the Discord channel and we should investigate how to replicate this behaviour

zenodeapp:
PR to fix #1267. This makes use of a method described in https://docs.keplr.app/api/#how-to-detect-keplr to detect the Keplr wallet extension.

- This not only checks the `window`-object, but also checks the `document` state before it eventually decides whether Keplr is installed or not.
- Added an install method which opens the keplr download page in a _blank tab.
- Added the `install()` and `get()` functions to the `WalletConnector`-interface as this is likely going to become common among other integrations. But unsure about this, so tell me if this needs to go.
- `getSigner` needed to become an async function due to this detection method.

In my testing I didn't get problems with detecting Keplr anymore, but please double check if it's resolved. I'll also open another PR with a simpler approach, but I don't think that one is as robust as this method (#1283).

ZEN

### DIFF
--- a/apps/namadillo/src/integrations/Keplr.ts
+++ b/apps/namadillo/src/integrations/Keplr.ts
@@ -15,9 +15,30 @@ export class KeplrWalletManager implements WalletConnector {
     console.warn("Keplr is not available. Redirecting to the Keplr download page...");
     window.open("https://www.keplr.app/get", "_blank");
   }
+  
+  private async _get(): Promise<Keplr | undefined> {
+    if ((window as KeplrWindow).keplr) {
+      return (window as KeplrWindow).keplr;
+    }
 
+    if (document.readyState === "complete") {
+      return (window as KeplrWindow).keplr;
+    }
+
+    return new Promise<Keplr | undefined>((resolve) => {
+      const documentStateChange = (event: Event): void => {
+        if (event.target && (event.target as Document).readyState === "complete") {
+          resolve((window as KeplrWindow).keplr);
+          document.removeEventListener("readystatechange", documentStateChange);
+        }
+      };
+      
+      document.addEventListener("readystatechange", documentStateChange);
+    });
+  }
+  
   async get(): Promise<Keplr> {
-    const keplr = await this._getKeplr();
+    const keplr = await this._get();
     if(!keplr) this.install();
     return keplr!;
   }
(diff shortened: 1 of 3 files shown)


apps/namadillo/src/integrations/Keplr.ts
```diff
@@ -8,21 +9,55 @@ import { WalletConnector } from "./types";
 import { basicConvertToKeplrChain } from "./utils";
 
 type Signer = OfflineAminoSigner & OfflineDirectSigner;
-const keplr = (window as KeplrWindow).keplr!;
 
 export class KeplrWalletManager implements WalletConnector {
+  install(): void {
+    console.warn("Keplr is not available. Redirecting to the Keplr download page...");
+    window.open("https://www.keplr.app/get", "_blank");
+  }
```
jurevans, line 17: I think this is good! Though we may want to invoke `install()` from outside of the integration (for example, if `get()` doesn't return Keplr, we could display a `Download` button, which in turn invokes `install()`, instead of calling it automatically). I wanted to see what @pedrorezende thought of this too, as we've been chatting about how to handle that scenario for Keplr. Perhaps we can keep this method (`install()`), but invoke it outside of `get()`?

zenodeapp: That's an idea indeed! So commenting out the part where it automatically calls install, but keep the install method so it's callable? 

It would also be an idea to have like an `isInstalled()` method to quickly see whether it is around or not and base the button on this boolean value.

I'm not behind the pc atm, but I could remove the part where it auto calls install inside of the get!

jurevans: Yeah, I think for now, just take out `if(!keplr) this.install();` so that it's not automatically called. We need to figure out exactly how to handle the Download button still on our end, and we can certainly address that in another PR!

pedrorezende: I agree with @jurevans , later today I can hook this `install` method somewhere in order to download Keplr after clicking on a button to do so. Thank you very much for this PR @zenodeapp !

pedrorezende (approved):
LGTM! Thank you very much!

pedrorezende:
Fixes #1267

zenodeapp:
One final remark, it might become necessary to create an `isInstalled()` method for the button logic, due to the non-null assertion in the `get()` function, like:

```js
async isInstalled(): Promise<boolean> {
    const keplr = await _get();
    return keplr !== undefined;
}
```

PR #1283, opened 2024-11-20
fix: simple way to detect/install keplr (#1267)

zenodeapp:
Hey, **I recommend using #1282, which is a robuster way to detect Keplr**, but this is a simpler method with less code to get rid of the detect issue (#1267). It also opens the keplr download page once it goes undetected.

It's mostly up to how you guys are going to continue implementing wallets and the logic you decide to use.

ZEN


zenodeapp:
#1282 got merged.

PR #1300, opened 2024-11-21
fix: fix token price

euharrison:
Fix token prices when using the sqs api

Screenshot 2024-11-21 at 17 40 08


pedrorezende (approved):
LGTM!

PR #1301, opened 2024-11-22
feat: handle nam as denominated amount

mateuszjasiuk:
Indexer part: https://github.com/anoma/namada-indexer/pull/157
Issue: https://github.com/anoma/namada-indexer/issues/156

Changes:
- We do not have to denominate NAM manually in namadillo
- ~~Added envs so you can optionally develop against localnet. You have to specify: `VITE_LOCALNET=true, VITE_LOCALNET_CHAIN_ID, VITE_LOCALNET_NAM_TOKEN`~~
- added `localnet-config.toml`, by default added to gitignore. You can add it to `namadillo/public/localnet-config.toml` if you want to develop against local chain. Example file:
```toml
# Localnet configuration file
enabled = true
chain_id = "local.040061ff54009d1fb42c0ac5"
token_address = "tnam1qxp7u2vmlgcrpn9j0ml7hrtr79g2w2fdesteh7tu"
```


apps/namadillo/src/registry/namadaAsset.ts
```diff
@@ -1,21 +1,33 @@
...
-      exponent: 6,
-    },
-  ],
-  logo_URIs: { svg: namadaShieldedSvg },
-} satisfies Asset;
+const ADDRESS = "tnam1qxp7u2vmlgcrpn9j0ml7hrtr79g2w2fdesteh7tu";
+
+// We can't return "satisfies Asset" from fn
+// eslint-disable-next-line @typescript-eslint/explicit-function-return-type
+export const namadaAsset = () => {
```
euharrison, line 10: I'm removing this implementation and moving it to `namada-chain-registry` https://github.com/anoma/namada-interface/pull/1308

Do you think it could have some impact here?

apps/namadillo/jest.config.ts
```diff
@@ -1,9 +1,28 @@
...
 
 /** @type {import('ts-jest').JestConfigWithTsJest} */
 module.exports = {
-  ...createDefaultPreset(),
+  transform: {
+    "^.+\\.tsx?$": [
+      "ts-jest",
+      {
+        diagnostics: {
+          ignoreCodes: [1343],
```
euharrison, line 11: could you add some reference why we need this?

mateuszjasiuk: I can try xD

apps/namadillo/src/atoms/integrations/functions.ts
```diff
@@ -370,3 +370,37 @@ export const createIbcTx = async (
...
+  ...internalDevnetChain,
+  chain_name: "localnet",
+  chain_id: chainId,
+});
+
+export const namadaLocalAsset = (tokenAddress: string): AssetList => ({
+  ...internalDevnetAssets,
+  chain_name: "localnet",
+  assets: internalDevnetAssets.assets.map((asset) =>
+    asset.name === "NAM" ?
```
emccorson, line 384: I think this is not going to work once we're using the registry with anoma/namada-chain-registry#13 merged. Maybe could use `"Namada"` or check against `symbol` or `base` instead?

apps/namadillo/src/utils/index.ts
```diff
@@ -122,12 +122,14 @@ const findDisplayUnit = (asset: Asset): AssetDenomUnit | undefined => {
...
 
 const isNamAsset = (asset: Asset): boolean =>
-  asset.symbol === namadaAsset.symbol;
+  asset.symbol === namadaAsset().symbol;
 
 export const toDisplayAmount = (
   asset: Asset,
   baseAmount: BigNumber
 ): BigNumber => {
+  if (isNamAsset(asset)) return baseAmount;
```
emccorson, line 131: I think this won't work for NAM transferred from other chains. The indexer will return the balances in display denom for NAM from the current chain, but in base denom for NAM from other chains. But this check here will assume any token with `symbol` as `NAM` is in display denom.

I was wondering if this check should be outside the `toDisplayAmount` function anyway, since to me it seems like that logic is outside of the scope of the function. What do you think?

emccorson (approved):
Looks good to me! I just found a couple of things that might not be working.

apps/namadillo/src/utils/index.ts
emccorson: Actually, I just also realised I think it's not going to work for displaying NAM on Cosmos chains either. The NAM balance we get from a Cosmos chain will be in base denom, but this check will stop it being converted to display denom.

mateuszjasiuk: yeah maybe move this outside makes sense :D Just so I understand better, what do you mean by "NAM  transferred from other chains." For NAM to arrive on other chain it first has to be sent from namada chain. So if we send NAM to cosmos and we send it back, will be it treated as different token than regular NAM?

emccorson: The two cases I think won't work are:

1. NAM that has been sent to a Cosmos chain (since the balance query for cosmjs returns amounts in base denom)
2. NAM from something other than the current chain e.g. your current chain is housefire but you sent some NAM from internal devnet to housefire

In both cases, these should display properly as the Namada asset in Namadillo, but the denomation will be wrong in this PR because Namadillo thinks it's already in display denom but it's actually in base denom.

The case you're talking about where you send NAM from Namada to Cosmos and then back should work fine I think, because the indexer will still return the balance using the original tnam address and in display denom (as far as I know).

PR #1308, opened 2024-11-22
feat: use namada asset from chain-registry

### LINKED ISSUE #1285: Disable NAM in the asset dropdowns + The shield button on NAM in the overview in phases 2/3/4

euharrison:
Deletes the file `namadaAsset.ts` and use the asset from namada-chain-registry

Also, spread the function `isNamAsset` across multiple component to ensure they all use the same condition and don't create inconsistencies

closes #1285


apps/namadillo/src/App/Transfer/__tests__/AssetImage.test.tsx
```diff
@@ -25,20 +25,20 @@ describe("Component: AssetImage", () => {
 
   it("renders the shielded badge when isShielded is true", () => {
     render(<AssetImage asset={assetMock} isShielded={true} />);
-    const badgeElement = screen.getByAltText(/Namada logo/i);
+    const badgeElement = screen.getByAltText(/Nam logo/i);
```
pedrorezende, line 28: I don't think this should be correct. NAM is the symbol and Namada the name of the asset

pedrorezende (changes requested):
Maybe, just maybe something isn't quite right with the "Nam" text instead of "Namada". Requesting some changes just for us to make this conversation first. The code in general LGTM!

apps/namadillo/src/App/Transfer/__tests__/AssetImage.test.tsx
euharrison: I agree, I opened this PR to fix this on the chain registry https://github.com/anoma/namada-chain-registry/pull/13/files

euharrison: reverted :)

pedrorezende (approved):

PR #1318, opened 2024-11-26
Extension: feat/1317 - Indicate type of Transfer in Tx approval

### LINKED ISSUE #1317: Extension: Type of Transfer should be indicated in approval
When approving any kind of Transfer in the extension, the transfer is always indicated as the generic `Transfer` type, as all transfer build functions return the same type. We can improve this by comparing the `source` & `target` to display a more descriptive label:

*source*->*target* = type of transfer label:

- transparent->transparent = `Transparent Transfer`
- transparent->shielded = `Shielding Transfer`
- shielded->shielded = `Shielded Transfer`
- shielded->transparent = `Unshielding Transfer`

Example `Transfer` details:

```json
{
  "token": "tnam1q9gr66cvu4hrzm0sd5kmlnjje82gs3xlfg3v6nu7",
  "feeAmount": "0.000001",
  "gasLimit": "50000",
  "chainId": "namada-dryrun.abaaeaf7b78cb3ac",
  "commitments": [
    {
      "txType": "Transfer",
      "hash": "67C34EC66D489A8F78DAD875433BE394B951B95F40C6C87E113E3C554F1071D1",
      "txCodeId": "8745309a86ab31d1eb7e361a4ad9a7fcbd4884bb5de94846ee116ef394796b29",
      "memo": "",
      "sources": [
        {
          "owner": "tnam1qpx3r356jwjlhv3cacweynswtdkny7ntlchkc4cs",
          "token": "tnam1q9gr66cvu4hrzm0sd5kmlnjje82gs3xlfg3v6nu7",
          "amount": "123"
        }
      ],
      "targets": [
        {
          "owner": "tnam1qz4sdx5jlh909j44uz46pf29ty0ztftfzc98s8dx",
          "token": "tnam1q9gr66cvu4hrzm0sd5kmlnjje82gs3xlfg3v6nu7",
          "amount": "123"
        }
      ]
    }
  ]
}
```

jurevans:
Resolves #1317 

This PR adds a more specific label to the Tx approval view in the case of `Transfer`.

- Deserialized Tranfers should have `source` & `target` compared to determine the type of transfer
- Shielded pool address is currently indicated with `MASP`

### Screenshots

Transparent:
Screen Shot 2024-11-26 at 8 58 44 AM

Shielding:
Screen Shot 2024-11-26 at 9 08 18 AM


apps/extension/src/utils/index.ts
```diff
@@ -115,3 +122,36 @@ export const isCustomPath = (path: Path): boolean => {
...
+ * @returns string label
+ */
+export const parseTransferType = (
+  tx: TransferProps
+): { source: string; target: string; type: TransferType } => {
+  const { sources, targets } = tx;
+  const source = sources[0].owner;
+  const target = targets[0].owner;
+
+  const shieldedPoolAddress = "tnam1pcqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqzmefah";
```
euharrison, line 138: should we pull this from namada-chain-registry?

jurevans: @euharrison I'll take a look! The extension doesn't touch the registry at all yet, but perhaps it's time it does, so long as it doesn't add too much bulk to the build size. The MASP address will indeed need to match on any given chain, so I think that's the only way to do it. Thanks!

jurevans: Updated this to not even need to use the shielded pool address!

Issue #1324, opened 2024-11-27
SDK: Finalize v0.14.0 Release


jurevans:
We will soon need to publish the updated SDK to `@namada/sdk`. The following is needed:

_Recurring tasks_
- Regenerate documentation for `@namada/sdk` & `@namada/types` (there have been some changes!)
- Export any types from `@namada/shared` from SDK that are needed by Namadillo

__NOTE__ We likely will not have the remaining MASP features in time for this release, but we can follow up with a minor release once proof-gen keys functionality is added to Namada (see https://github.com/anoma/namada/pull/3797) - While proof-gen keys are soon available for Namada, we will need a Ledger release that incorporates Zip32

Hopefully we can soon get https://github.com/anoma/namada/pull/4127 incorporated into the SDK & Keychain!

jurevans:
Closing this, all features will be present in 0.17.0 SDK, and this is set to publish automatically once released!

PR #1325, opened 2024-11-27
SDK/Namadillo: feat/1324 - Move to SDK Imports

jurevans:
Relates to #1324 

- Move any `@namada/shared` imports in Namadillo to `@namada/sdk`, as that is the package that will be published


euharrison (approved):
👍

PR #1335, opened 2024-11-28
fix: fix type and add error to claim rewards toast

emccorson:
This PR adds the error details to the claim rewards toast and fixes the types so it does not appear as a success toast.

I think we could hide the error details behind the "View details" link, but this seems to not be displayed for error toasts as of #1084. We could probably move the details added in #1084 to the `decription` prop instead of the `details` prop, and then remove the hiding of "View details" for error toasts, but I haven't done it in this PR.

Before: 
image

After:
2024-11-28-114601_773x601_scrot

---

### Added
- Add error details to claim rewards error toast

### Fixed
- Make claim rewards error toast error type instead of success


pedrorezende (approved):

jurevans (approved):
LGTM!

PR #1336, opened 2024-11-28
bug/1329: Extension - Simplify chain service

jurevans:
As the extension only needs to know the `chainId` in order to validate signing requests, remove complex chain config and the associated validation (this is causing issues where old versions of the extension cannot update, and breaks the App popup!)

- [ ] Simply chain service, only track chain ID
- [ ] Remove unused `Namada` calls for `getChain`, etc.
- [ ] Remove any other unused items
- [ ] Fix any broken tests

### Testing

- Simply load this version if you're already running a local dev extension - it should gracefully update to the new storage and not throw any errors!

### DIFF
--- a/apps/extension/src/background/sdk/service.ts
+++ b/apps/extension/src/background/sdk/service.ts
@@ -1,7 +1,6 @@
 import { chains } from "@namada/chains";
 import { Sdk, getSdk } from "@namada/sdk/web";
 import sdkInit from "@namada/sdk/web-init";
-import { LocalStorage } from "storage";
 
 const {
   NAMADA_INTERFACE_NAMADA_TOKEN:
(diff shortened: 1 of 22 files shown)


apps/extension/src/background/sdk/service.ts
```diff
@@ -18,14 +17,10 @@ export class SdkService {
...
 
-  static async init(localStorage: LocalStorage): Promise<SdkService> {
-    // Get stored chain
-    const chain = await localStorage.getChain();
-    // If chain is not stored, use default chain information
-    const rpc = chain?.rpc || chains.namada.rpc;
-
+  static async init(): Promise<SdkService> {
+    // Use fake RPC, the extension does not actually issue RPC requests:
+    const rpc = chains.namada.rpc;
```
pedrorezende, line 22: Should we add an empty string to prevent using it by accident in the future?

pedrorezende (approved):
Very good refactor! 🥇

apps/extension/src/background/sdk/service.ts
jurevans: Thanks! I cleaned this up some more, this service shouldn't need this value on init, then I just passed another `"NOT USED..."` string to initialize the SDK

PR #1343, opened 2024-11-29
feat/1229: Extension style improvements - v0.3.3

### LINKED ISSUE #1229: Extension: Finalize v0.3.3 Release
This issue is to track the final round of touch-up before releasing extension.

## TODO

- Custom path inputs should not have _number_ arrows (we can force numbers, min & max without these being present)
- Fix font sizes and spacing to match designs
- Ledger HW Wallet `Advanced` - what step of set-up should this be? This needs to happen before we do the address/public key query to incorporate any customization, so perhaps it should be displayed at the top?
- Bump SDK package version to `0.11.0` before submitting final extension build
  - Once other PRs are merged that affect `types` package, regenerate docs and bump version in SDK `package.json`

### Viewing Key
377935981-a6d70563-d77d-4e99-8c82-0cc99ed4b0a4

### Figma

https://www.figma.com/design/NFyHbLZXBSl3aUsMxtffvV/Namada-End-User-Interface-V1.0-%2F-Phase-1%3A-Block-Party?node-id=6361-75720&node-type=frame&t=vEy6rGuJhxK4kFZS-0

jurevans:
Resolves #1229 

- [x] Ledger connect - move Advanced to before Step 1
- [x] Improve styles in Chrome/FF
  - Viewing key padding
  - Derivation Path number input - remove buttons
  - Firefox - Viewing Key form width (certain elements cause the width to expand in FF - I set this on a higher-level element and removed individual overrides - This could perhaps be improved in the future, but gives us good results across browsers, and can now be changed in a single place!)

### DIFF
--- a/apps/extension/src/App/Accounts/DeleteAccount.tsx
+++ b/apps/extension/src/App/Accounts/DeleteAccount.tsx
@@ -94,7 +94,7 @@ export const DeleteAccount = (): JSX.Element => {
       >
         <Stack className="flex-1 justify-center" gap={4} full>
           <Stack as="header" gap={4}>
-            <Alert type="warning" title="Alert!" className="max-w-84">
+            <Alert type="warning" title="Alert!">
               <p className="mb-4">
                 Make sure that you&apos;ve backed up your seed phrase and
                 private key.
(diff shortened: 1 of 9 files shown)


pedrorezende (approved):
LGTM!

PR #1348, opened 2024-11-30
Namadillo: Displaying all validators

### LINKED ISSUE #1339: Namadillo does not shows in the Validator list some active validator
Scenario:

1. Dry Run network: namada-dryrun.abaaeaf7b78cb3ac
2. Validator started after genesis time.
3. Validator is in consensus set (namadac validator-state) and has signed a little under 8000 blocks
4. Validator is visible in the bonded-stake since epoch 181 ( namadac bonded-stake | grep tnam1qx7yy6xwp89j947fptquxlwlvzflmzk2eyhna8v5)
5. Validator has active stake
6. It is possible to stake to the validator from command line.

Expected Behaviour
It must be possible from Namadillo to search and delegate to the validator

Current Behaviour
The validator is not visible and no NAM can be staken from Namadillo

pedrorezende:
We were filtering the `all` validators query by the active ones only. This PR fixes this behavior and possibly closes #1339 .

### DIFF
--- a/apps/namadillo/src/atoms/validators/services.ts
+++ b/apps/namadillo/src/atoms/validators/services.ts
@@ -1,6 +1,5 @@
 import {
   DefaultApi,
-  ValidatorStatus as IndexerValidatorStatus,
   VotingPower as IndexerVotingPower,
   MergedBond,
   Unbond,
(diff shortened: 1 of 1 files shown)


jurevans (approved):
LGTM!

PR #1352, opened 2024-12-01
feat/1349 - Require unlocked keychain for all access

### LINKED ISSUE #1349: Locked Keychain Allows Namadillo Login/Logout
It is possible log in and log out of Namadillo despite the keychain extension being locked, even after clearing site data. 

image

image

image

image

jurevans:
Resolves #1349 

- [x] If the user is connected to the extension in Namadillo, and they lock the extension, disconnect accounts & update Namadillo
- [x] When a user connects to extension, if the wallet is locked, first prompt for login
- [x] Successful authentication directs user to requested view
- [x] Wrap all top-level approvals with auth wrapper
- [x] Accounts cannot be queried from extension if it is locked, even if the domain is approved

### Testing

Testing should be straight-forward
- With a locked extension, load Namadillo - the user should be required to authenticate before proceeding to connection approval
- Locking the extension should disconnect Namadillo
  - `await namada.accounts()` returns empty array with locked extension
  - `await namada.defaultAccount()` returns `undefined` with locked extension
- Loading Namadillo on an approved domain and unlocked wallet works as before
- Loading Namadillo on an unapproved domain and unlocked wallet works as before
- If domain & chainId are already approved, with a locked keychain, if you authenticate, it should then automatically connect
- Lock the keychain, and in console, issue a similar command to this:
  ```javascript
  await namada.sign({ signer: "tnam1qz4sdx5jlh909j44uz46pf29ty0ztftfzc98s8dx", txs: [] });
  await namada.signArbitrary({ signer: "tnam1qz4sdx5jlh909j44uz46pf29ty0ztftfzc98s8dx", data: "" });
  ```
  - This currently returns an error: `Keychain is locked!`. This might be the safest thing to do to avoid unwarranted signing - forcing integrations to explicitly connect first. Optionally, we could just simply remove these `throw` statements which will instead launch the Login->Approval, though it may be unclear which domain is issuing the request.
    - If we decide to enable Login for signing requests, we should indicate the `domain` that is requesting this


apps/extension/package.json
```diff
@@ -47,6 +47,7 @@
     "bignumber.js": "^9.1.1",
     "buffer": "^6.0.3",
     "dompurify": "^3.0.2",
+    "fp-ts": "^2.16.9",
```
jurevans, line 50: Wanted to add this before I forget - it previously got it's dependency as it was specified in Namadillo. Makes it explicit here

apps/extension/src/background/vault/service.ts
```diff
@@ -84,12 +84,6 @@ export class VaultService {
     return (await this.sessionStore.get("password")) || "";
   }
 
-  // TODO: I think we shouldn't expose the password like this, but it's required
-  // for SDK methods. Use cautiously.
-  public async UNSAFE_getPassword(): Promise<string> {
-    return await this.getPassword();
-  }
```
jurevans, line 91: clean up

apps/extension/package.json
```diff
@@ -47,6 +47,7 @@
     "bignumber.js": "^9.1.1",
     "buffer": "^6.0.3",
     "dompurify": "^3.0.2",
+    "fp-ts": "^2.16.1",
```
jurevans, line 50: Extension previously did not specify this dependency, and was only building because it is specified in Namadillo, so added the same version here

apps/extension/src/Approvals/Approvals.tsx
```diff
@@ -65,22 +70,36 @@ export const Approvals: React.FC = () => {
...
+            </WithAuth>
+          }
         />
         <Route
           path={TopLevelRoute.ApproveDisconnection}
-          element={<ApproveDisconnection />}
+          element={
+            <WithAuth>
+              <ApproveDisconnection />
+            </WithAuth>
```
jurevans, line 84: I don't think it's possible to reach a state where the wallet is locked on this route (and the `WithAuth` instances that follow) in Namadillo, but it could be the case on another interface.

apps/namadillo/src/hooks/useExtensionEvents.tsx
```diff
@@ -26,4 +26,8 @@ export const useExtensionEvents = (): void => {
       (await injectedNamada?.isConnected()) ? "connected" : "idle"
     );
   });
+
+  useEventListenerOnce(Events.ExtensionLocked, async () => {
+    setNamadaExtensionConnected("idle");
+  });
 };
```
pedrorezende, line 33: One thing that is still not working for me is when you have connected to the extension before and unlock the extension. Namadillo still appears like it is disconnected, until you hit the button to connect. I think it would be desirable to have another event here to check when the extension gets unlocked.

pedrorezende (changes requested):
Great job, Justin! 🥇
I'm just requesting a change because I found a behavior that we might want to address. Let me know if we should address that in a different PR.

apps/namadillo/src/hooks/useExtensionEvents.tsx
jurevans: @pedrorezende Thanks for checking this! I ran into the same thing. The extension `App` doesn't fire any event when it's directly unlocked (that behavior only happens on connect). I will take a look and see if it's relatively do-able. Basically we want the Namadillo to re-check if it's connected to the Keychain on `App`->`Login`

pedrorezende (approved):
Working niiiice 🚀

mateuszjasiuk (approved):
Tested and LGTM :)

PR #1362, opened 2024-12-02
feat: listen for extension

euharrison:
Listen for extension to be enabled.

Useful if the user allow the extension only on click or any other barrier that delay the extension to be loaded

https://github.com/user-attachments/assets/3c85d3aa-99ec-4b64-86b6-76b868f79165


emccorson (approved):
Looks good to me!

PR #1365, opened 2024-12-02
Namadillo: Reviewing transaction timelines and fixing a few things

pedrorezende:
- Simplifying `mapCoinsToAssets` code
- Checking for `baseDenom` on chain registry, so if we don't have the channel added to the registry, it might display everything correct
- Improvements on Namada transfer timeline

### DIFF
--- a/apps/namadillo/src/atoms/integrations/functions.ts
+++ b/apps/namadillo/src/atoms/integrations/functions.ts
@@ -227,30 +227,36 @@ export const mapCoinsToAssets = async (
   const chainName = cosmosRegistry.chains.find(
     (chain) => chain.chain_id === chainId
   )?.chain_name;
+
   const assets = mapUndefined(assetLookup, chainName);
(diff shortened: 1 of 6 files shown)


apps/namadillo/src/atoms/integrations/functions.ts
```diff
@@ -170,6 +170,15 @@ const tryDenomToIbcAsset = async (
 
   const { path, baseDenom } = denomTrace;
 
+  const assetOnRegistry = tryDenomToRegistryAsset(
```
emccorson, line 173: Maybe you already accounted for this, but there are cases where the same asset will have different configurations on different chains, so doing this allows more assets to be decoded but at the risk of showing the wrong information for an asset (which would mainly matter if the exponents are different, since we could be showing the wrong amount).

emccorson (approved):
Looks good!

PR #1370, opened 2024-12-03
feat: force bond in the claim and stake flow

mateuszjasiuk:
We have to do this otherwise in the scenario that we have more rewards than balance, the namada_sdk validation will fail with `NotEnoughBalance` error. Forcing bond will still log error in terminal, but we can't do much about it atm.

### DIFF
--- a/apps/namadillo/src/atoms/staking/services.ts
+++ b/apps/namadillo/src/atoms/staking/services.ts
@@ -129,6 +129,9 @@ export const createClaimAndStakeTx = async (
     props: ClaimRewardsProps | BondProps
   ): Promise<TxMsgValue> => {
     if ("amount" in props) {
+      // We have to force it in case: current balance < rewards to claim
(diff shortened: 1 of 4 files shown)


apps/namadillo/src/atoms/staking/services.ts
```diff
@@ -129,6 +129,9 @@ export const createClaimAndStakeTx = async (
     props: ClaimRewardsProps | BondProps
   ): Promise<TxMsgValue> => {
     if ("amount" in props) {
+      // We have to force it in case: current balance < rewards to claim
+      // This will still log the error msg in the terminal, unfortunately we can't do much about it
+      wrapperTxProps.force = true;
```
jurevans, line 134: Interesting. Will this potentially break deserialization in the extension? If it simply ignores it, I think we're good!

jurevans (approved):
LGTM! Left one small comment, but I think this will be fine!

PR #1378, opened 2024-12-04
feat: shielded rewards intergration

### LINKED ISSUE #1371: Add endpoint for shielded rewards
We need an endpoint to query the estimated shielded rewards for a given address. 

image

mateuszjasiuk:
Resolves #1371

This  integrates the shielded rewards.
~~The downside is that is seems that the endpoint in namada rpc returns some weird values,  so testing might be a bit tricky.
How  it works for me atm is that if I shield a lot of nam and wait one masp epoch I will see some rewards for one masp epoch, but then it will go back to 0.
You can shield more and repeat  the process~~

Not sure if campfire has nam rewards enabled :/ 
Also added a function which can simulate rewards for this:
image


mateuszjasiuk:
Moved to draft as Marco is changing some stuff in namada.

jurevans (approved):
Code looks good! Tested locally, shielded rewards look great!

PR #1389, opened 2024-12-06
feat/1047: Ledger transport updates, Updated Step 2 text

### LINKED ISSUE #1047: bug: Ledger HW Wallet - when sending multiple Tx, Pending notification doesn't clear on success
In testing Ledger HW wallet with multiple transactions, I noticed a couple issues:

- Console would log several `Unreachable` errors (wasm), but this didn't affect the Tx - they all still successfully went through
- Pending notification does not clear when Success notifications display

As Ledger Tx must be signed & submitted as multiple Txs, we see multiple success notifications. Perhaps we should have a Pending notification for each Tx so that it can be identified correctly, each notification linked to its Tx hash.

jurevans:
Resolves #1047 

- [x] Bump USB transport packages
- [x] Bump Namadillo
- [x] Display which Tx out of all Txs are currently being prompted for signing
  - If only one Tx is being signed, just display `Approve on your device`
  - If more than one needs signing, also display `Tx x of y`

### Design

Suggested updated Step 2 text:

Screenshot 2024-12-06 at 12 52 16

### Updated Ledger Approval Step 2

Screen Shot 2024-12-06 at 9 44 21 AM

### DIFF
--- a/apps/extension/package.json
+++ b/apps/extension/package.json
@@ -41,9 +41,9 @@
   "dependencies": {
     "@cosmjs/encoding": "^0.29.0",
     "@dao-xyz/borsh": "^5.1.5",
-    "@ledgerhq/hw-transport": "^6.30.0",
-    "@ledgerhq/hw-transport-webhid": "^6.28.0",
-    "@ledgerhq/hw-transport-webusb": "^6.28.0",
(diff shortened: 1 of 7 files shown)


euharrison (approved):

mateuszjasiuk (approved):
Tested on ledger nano s. All works!

PR #1391, opened 2024-12-06
bug/1047: Namadillo Ledger Notifications

### LINKED ISSUE #1047: bug: Ledger HW Wallet - when sending multiple Tx, Pending notification doesn't clear on success

jurevans:
Resolves #1047 

- [x] Submitting any number of signed Ledger transactions results in one Pending notification, and one Success/PartialSuccess/Error notification that clears the pending one
- [x] Bump Namadillo version - this should be in next tagged release! `1.0.2` -> `1.0.3`

_NOTE_ - the major change in this PR was collecting commitments from any number of Txs, such that when we launch the notification, batch Tx & multiple-Txs can be handled the same way.

### Testing

I'm testing on Housefire, which has 3 validators.

- Faucet: https://faucet.knowable.run
- Indexer: https://indexer.knowable.run
- RPC: https://indexer.knowable.run

#### Txs

- Test Txs with mnemonic/private key Keychain accounts (single & multiple Bond, for example)
- Test single and multiple Ledger HW Wallet Tx (single & multiple Bond, with & without `RevealPK` _NOTE_ If you've already revealed PK on Housefire, simply go to `Add Keys`, and on the Ledger connection page, click `Advanced` and specify a different path, thus generating a different Public Key & Address, then transfer from faucet to that address!)

#### Errors

- You can trigger error states by overriding the `feeAmount` & `gasLimit`
  - Setting `gasLimit` too low should trigger the wrapper rejection (2nd to last screenshot below)
  - Setting `feeAmount` too low should trigger the Tx applied errors (last screenshot)

#### Example Tx sent for signing with Ledger:
Screen Shot 2024-12-07 at 8 35 09 AM

#### Pending Ledger Tx notification (bonds summed):
Screen Shot 2024-12-07 at 8 35 34 AM

#### Multiple Ledger-signed Txs combined into one Success message with appropriate details:
Screen Shot 2024-12-07 at 8 35 44 AM

#### Wrapper error (rejects quickly if gas limit is incorrect)
Screen Shot 2024-12-07 at 8 58 26 AM

#### Tx applied error if all fail
Screen Shot 2024-12-07 at 9 00 51 AM

### DIFF
--- a/apps/namadillo/package.json
+++ b/apps/namadillo/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@namada/namadillo",
-  "version": "1.0.2",
+  "version": "1.0.3",
   "description": "Namadillo",
   "repository": "https://github.com/anoma/namada-interface/",
   "author": "Heliax Dev <info@heliax.dev>",
--- a/apps/namadillo/src/App/Governance/SubmitVote.tsx
+++ b/apps/namadillo/src/App/Governance/SubmitVote.tsx
@@ -29,14 +29,12 @@ import { IoClose } from "react-icons/io5";
 import { useNavigate } from "react-router-dom";
 
 const dispatchVoteTx = (tx: TransactionPair<VoteProposalProps>): void => {
-  tx.signedTxs.forEach((signedTx) => {
-    broadcastTx(
-      tx.encodedTxData,
-      signedTx,
-      tx.encodedTxData.meta?.props,
-      "VoteProposal"
-    );
-  });
+  broadcastTx(
+    tx.encodedTxData,
+    tx.signedTxs,
+    tx.encodedTxData.meta?.props,
+    "VoteProposal"
+  );
 };
 
 export const SubmitVote: React.FC = () => {
--- a/apps/namadillo/src/App/Ibc/IbcWithdraw.tsx
+++ b/apps/namadillo/src/App/Ibc/IbcWithdraw.tsx
@@ -135,15 +135,11 @@ export const IbcWithdraw: React.FC = () => {
       };
 
       setTransaction(tx);
-      await Promise.allSettled(
-        signedTxs.map((signedTx) => {
-          return broadcastTx(
-            encodedTxData,
-            signedTx,
-            encodedTxData.meta?.props,
-            "IbcTransfer"
-          );
-        })
+      await broadcastTx(
+        encodedTxData,
+        signedTxs,
+        encodedTxData.meta?.props,
+        "IbcTransfer"
       );
     } catch (err) {
       setGeneralErrorMessage(err + "");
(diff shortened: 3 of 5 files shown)


euharrison (approved):
Nice! ✅ 

I tested only the notification with the Chrome Extension and stressed a lot it.

The unique border case I saw is:

1. I have a `Reveal + Bond` popup, and I don't sign it, but hold it opened. 
2. Then, I open a second `Reveal + Bond` popup.
3. So, I sign both popups and wait for the notification
4. And, I had two pending notifications (good) but both show as failed (bad) and, in the end, the transaction was executed successfully, the reveal and both bond tx

I believe they showed the failed status because the 'second Reveal' failed and updated both notifications

But, I see this as an ultra border case, so I don't believe we need to be worried about this

Screenshot 2024-12-11 at 09 09 17

Screenshot 2024-12-11 at 09 09 40

jurevans:
> Nice! ✅
> 
> I tested only the notification with the Chrome Extension and stressed a lot it.
> 
> The unique border case I saw is:
> 
>     1. I have a `Reveal + Bond` popup, and I don't sign it, but hold it opened.
> 
>     2. Then, I open a second `Reveal + Bond` popup.
> 
>     3. So, I sign both popups and wait for the notification
> 
>     4. And, I had two pending notifications (good) but both show as failed (bad) and, in the end, the transaction was executed successfully, the reveal and both bond tx
> 
> 
> I believe they showed the failed status because the 'second Reveal' failed and updated both notifications
> 
> But, I see this as an ultra border case, so I don't believe we need to be worried about this
> 
@euharrison Thanks for looking into this and testing! Yeah, I think this is a difficult & unusual scenario to support. The way IDs are calculated for Toasts means that if you attempt to submit the same Tx twice (while the first approval is open) you can run into React key collisions, but also I'm not sure we should allow launching multiple approvals at once. Do you think Namadillo should prevent this behavior to begin with? I think it could be useful if Namadillo makes it clear that an approval is pending. Being able to do this doesn't seem like it should be supported to me, and I suspect it's currently like that on `main` too. I won't try to support that in this PR, since it's just a bugfix, but we can take a look at this!

euharrison:
I'm good with the current behavior. And ok as well opening more than one popup. We can handle improvements later when we got more feedbacks from users

PR #1397, opened 2024-12-09
fix: decode continuous pgf proposal to ibc target

mateuszjasiuk:
Without it pgf continuous proposals with ibc target won't be decoded by io-ts.

https://github.com/user-attachments/assets/477f712d-d373-4f69-96ca-b032f91e1329


jurevans (approved):
LGTM!

PR #1407, opened 2024-12-13
fix: query only relevant bonds to see if it's possible to vote

mateuszjasiuk:
With this change we only query bonds that are active at certain epoch.


jurevans (approved):
LGTM!

PR #1410, opened 2024-12-13
feat: add gas token

### LINKED ISSUE #1284: Namadillo: Fee Modal / Select Fee Token drop-down
Users are able to specify alternate tokens when paying fees, but Namadillo doesn't yet offer this

- We need a component that allows a user to specify fees, as well as fee-token
- By default, we can have the source token specified to pay fees

### Designs

Figma:
image

image

euharrison:
Allow user to change the gas token on transactions

- Save on local storage the `namadillo:gasToken` with the last selected token used for gas
- Upgrade the type `GasConfig` adding the prop `gasToken` that could be the `tnam` address or the base denom for ibc.
- Also, add the `asset` for render fee values friendly
- On transfers, all references for `transactionFee` was replaced by `gasConfig`
- **The `gasLimit` and `gasPrice` are non-denominated amount (minBaseDenom)** and we only convert these values when rendering, or sending to keychain
- Create the atom `prices` to handle the dollar prices, sharing between components and atoms

closes #1284 

Screenshot 2024-12-13 at 18 11 21

Screenshot 2024-12-13 at 18 12 18

https://github.com/user-attachments/assets/b13c67c8-1e11-44ad-ab68-550cc0b10a28


apps/namadillo/src/utils/assets.ts
```diff
@@ -0,0 +1,34 @@
...
+
+export const unknownAsset = (denom: string): Asset => ({
+  denom_units: [{ denom, exponent: 0 }],
+  base: denom,
+  name: denom,
+  display: denom,
+  symbol: shortenAddress(denom, 4, 4),
+});
+
+export const findAssetByToken = (
```
pedrorezende, line 13: Can we keep the TODO comment of the original function?

apps/namadillo/src/App/Common/GasFeeModal.tsx
```diff
@@ -0,0 +1,102 @@
...
+        <div className="text-center">
+          <h2 className="font-medium">Select Gas Token</h2>
+          <div className="text-sm mt-1">
+            Gas fees deducted from your Namada accounts
+          </div>
+        </div>
+        <div className="flex flex-col mt-4 max-h-[60vh] overflow-auto">
+          {!data.length ?
+            <SkeletonLoading height="100px" width="100%" />
+          : data.map(({ token, minDenomAmount }) => {
```
pedrorezende, line 60: Maybe we should sort this list in order to display NAM token on top, what do you think?

apps/namadillo/src/utils/assets.ts
euharrison: Ok! I'll update to refer to the `tnamAddressToDenomTrace` that is the new function. Maybe also exploring resolving this at all :)

apps/namadillo/src/App/Common/GasFeeModal.tsx
euharrison: For sure! I'll do it

apps/namadillo/src/integrations/utils.ts
```diff
@@ -40,24 +39,33 @@ export const getAssetImageUrl = (asset?: Asset): string => {
...
   const feeToken = registry.chain.fees?.fee_tokens?.[0];
   if (typeof feeToken !== "undefined") {
+    const gasPrice =
+      feeToken.average_gas_price ??
+      feeToken.low_gas_price ??
+      feeToken.fixed_min_gas_price ??
+      feeToken.high_gas_price ??
+      0;
+    // TODO to be more precise on gasLimit, we should estimate the transfer gas fee
+    // `await client.simulate(sourceAddress, [transferMsg], undefined)`
```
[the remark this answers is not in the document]
euharrison: we don't need to sign :)

apps/namadillo/src/utils/assets.ts
euharrison: it'll be a bigger refactoring, so I left the comment 😁

PR #1415, opened 2024-12-17
feat: vks birthdays

mateuszjasiuk:
- Adds vks birthdays
- Adds migration

### How it works?
- when you create a new account we store a timestamp and the source of the account, either imported or generated
- based on source we create `DatedViewingKeys` in namadillo. For imported accounts `block_height` in `DatedViewingKeys` is always `0`(as it was created before, and we do not have the UI to insert timestamp or `block_height`). For generated accounts we get the timestamp and query indexer for corresponding block height(we look for block which timestamp is exactly the same or slightly lower)

### How to test?
- Add new account using import flow, look at shielded sync - it should start from 0%
- Add new account using generate flow, look at shielded sync - it should start from the last processed block

### Migrations
To migrate the data I've added schemas for each versions of `Vault` and the logic is as follows:
- we check the version of the data by decoding it using schemas, starting from the newest version(more about it in code)
- once we know persisted version we apply migrations starting from persisted and ending at newest
- example: currently persisted data is at version 2. newest version is 4. We try to decode data starting from `schemaV4 -> schemaV3 -> schemaV2`. Once we know that data is at version 2 we apply migrations: `v2->v3, v3-v4` and persist the `Vault`
- next time when you run migration it does nothing as detected version is 4

To test migrations you can checkout main, add account, then checkout this branch and check if entries in vault have source and timestamp specified


apps/extension/src/utils/index.ts
```diff
@@ -170,14 +170,24 @@ export const parseTransferType = (
...
   const account: Account = {
     alias,
     address,
     type,
     pseudoExtendedKey,
+    source,
+    timestamp,
   };
   if (isShielded) {
     account.viewingKey = owner;
```
jurevans, line 193: Once thought: Move `pseudoExtendedKey` (and maybe `timestamp` if it's required in Namadillo) down here, so we don't return `undefined` values in transparent accounts. If we don't need `source` in Namadillo, I would vote to drop it from here. Maybe timestamp can be a shielded-only thing too?

apps/extension/src/utils/index.ts
```diff
@@ -185,12 +185,13 @@ export const toPublicAccount = (derivedAccount: DerivedAccount): Account => {
     alias,
     address,
     type,
-    pseudoExtendedKey,
     source,
```
jurevans, line 188: One last comment - if we don't need `source` on transparent, I would maybe omit it here. It gets appended below for shielded which makes sense!

jurevans (approved):
Code LGTM! Just one comment about whether `source` should be included in transparent accounts. Tested migrating from old store, works great!

apps/extension/src/utils/index.ts
mateuszjasiuk: Yeah I forgot to remove the line :D

mateuszjasiuk: done!

euharrison (approved):
Very nice! ✅

PR #1421, opened 2024-12-18
feat: add button outline

### LINKED ISSUE #1342: Change button roll over states from black to white
Currently rollover in buttons makes them dissapear black on black

Either rollover states are white or we make a white keyline on the black
Screenshot 2024-11-28 at 20 45 35
Screenshot 2024-11-28 at 20 45 41

euharrison:
Add button outline on hover for cases where the button is on a black background

closes #1342


jurevans (approved):
LGTM!

PR #1432, opened 2024-12-19
feat: add time title helper

### LINKED ISSUE #1426: Namadillo: Add current time/date on dashboard
Screenshot_2024-12-18_20-59-31

# Problem
Live proposals have a date ending time posted but its unclear what the current time of the interface is

# Solution

Add a *current* date/time label some where on the dashboard

euharrison:
Add a `title` to proposal times so user can check the timezone

Also, fix the table name that was saying `UTC` while it was a localized value. Now it render the timezone as well

closes #1426

Some sample simulating different cities:

berlin

london

san francisco

### DIFF
--- a/apps/namadillo/src/utils/dates.ts
+++ b/apps/namadillo/src/utils/dates.ts
@@ -1,0 +1,53 @@
+import * as fns from "date-fns";
+import { DateTime } from "luxon";
+
+export const epochToString = (epoch: bigint): string =>
+  `Epoch ${epoch.toString()}`;
+
+export const secondsToDateTime = (seconds: bigint): DateTime =>
+  DateTime.fromSeconds(Number(seconds));
+
+export const secondsToTimeString = (seconds: bigint): string =>
+  secondsToDateTime(seconds).toLocaleString(DateTime.TIME_SIMPLE);
+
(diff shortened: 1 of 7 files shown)


apps/namadillo/src/utils/dates.ts
```diff
@@ -0,0 +1,54 @@
...
+
+export const secondsToFullDateTimeString = (seconds: bigint): string =>
+  secondsToDateTime(seconds).toLocaleString(DateTime.DATETIME_FULL);
+
+export const secondsToTimeRemainingString = (
+  startTimeInSeconds: bigint,
+  endTimeInSeconds: bigint
+): string | undefined => {
+  // TODO is this useful?
+  return;
```
jurevans, line 27: I would say this isn't useful, maybe we should delete any references to this function - it's been immediately returning for many months now :D

jurevans (approved):
This LGTM!

apps/namadillo/src/utils/dates.ts
euharrison: Yeah, I was thinking the same!

But I explored the feature and Figma and realized this is related to the remaining time for the ongoing proposal. So it was an one line change and I did it already  😁 

Screenshot 2024-12-20 at 12 09 12

PR #1433, opened 2024-12-20
feat: enable shielded transfers

mateuszjasiuk:
This enables shielded transfers(also shielding/unshielding) in the Transfers tab. I wrapped everything in TODO blocks as logic will be unified later.


euharrison (approved):
Nice! ✅

Issue #1439, opened 2024-12-23
Shielded Sync Improvements


mateuszjasiuk:
- vks birthdays
- shielded context invalidation(button + chain change)
- flag to prevent multiple syncs

brentstone:
Completed in #1441

PR #1441, opened 2024-12-23
feat: shielded sync improvements

### LINKED ISSUE #1420: Namadillo: Speed up Shielded-sync by adapting block bundle size to height
It takes about 12 seconds to `shielded-sync` from height 0 to 140,000 in the CLI, and about 1.5 minute in Namadillo, both using the MASP indexer with HTTP/2 enabled.

To reduce web worker serialization overhead, Namadillo requests blocks in bundles of 30, which adds up to 5800 bundles at this height. 

To speed up `shielded-sync`, the block bundle size could dynamically adapt to the height difference, to switch to bundles of for example 10,000, 1,000 and 30 blocks.

image

mateuszjasiuk:
Resolves #1420 and #1439 
- increase number of concurrent fetches
- add ability to invalidate shielded context

https://github.com/user-attachments/assets/6adbb5b6-da25-4553-93b1-d657ee7aef28

- store shielded context by chain id

image

### DIFF
--- a/apps/namadillo/src/App/Masp/ShieldedBalanceChart.tsx
+++ b/apps/namadillo/src/App/Masp/ShieldedBalanceChart.tsx
@@ -17,7 +17,6 @@ export const ShieldedBalanceChart = (): JSX.Element => {
   const shieldedTokensQuery = useAtomValue(shieldedTokensAtom);
   const [{ data: shieldedSyncProgress, refetch: shieledSync }] =
     useAtom(shieldedSyncAtom);
-
   const [showSyncProgress, setShowSyncProgress] = useState(false);
   const [progress, setProgress] = useState({
     [ProgressBarNames.Scanned]: 0,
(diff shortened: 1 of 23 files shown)


apps/namadillo/src/atoms/balance/services.ts
```diff
@@ -75,13 +77,14 @@ export function shieldedSync(
 
 export const fetchShieldedBalance = async (
   viewingKey: DatedViewingKey,
-  addresses: string[]
+  addresses: string[],
+  chainId: string
 ): Promise<Balance> => {
   // TODO mock shielded balance
```
euharrison, line 83: Can you delete from here as well?

apps/namadillo/src/App/Masp/ShieldedBalanceChart.tsx
```diff
@@ -73,13 +83,6 @@ export const ShieldedBalanceChart = (): JSX.Element => {
 
   useEffect(() => {
     shieledSync();
-
-    const timeoutId = setTimeout(() => {
```
euharrison, line 77: The original idea here is to avoid flashing a "100%" in case it syncs quickly (less than 3s)

euharrison (approved):

apps/namadillo/src/App/Masp/ShieldedBalanceChart.tsx
mateuszjasiuk: Yeah! I changed it because I had some problems getting right behavior when navigating from settings. RIght now instead of timeout it does this:
```ts
        // If total is more than one batch of 100, show progress
        if (total > 100) {
          setShowSyncProgress(true);
        }
 ```
so we show progress immediately but only if there is more than 100 blocks to fetch and parse :)

PR #1454, opened 2024-12-30
Namadillo: timeline animations

### LINKED ISSUE #1418: Namadillo: Progress bars and confetti for transaction status and shielded-sync
Whenever starting a transaction, the interface is more or less static, except for maybe a faintly blinking yellow circle. In fact, it takes a few seconds before the Keychain asks the user to sign the transaction. 

image

Then, when shielding, it takes a while for the ShieldedSyncWorker to catch up.  

image

And when the transaction is finished, the only indication is that the "Transfer Complete" becomes brighter. Very anticlimactic. I was still waiting for the transaction to finish up.

To the end-user, this dead time and too subtle visualization looks as if nothing is happening. I suggest to spice things up with progress bars and confetti. 🎉

Namadillo 1.0.6

pedrorezende:
This PR adds anime.js as a new animation engine and creates some cool animations for the timeline component.

#1418

[scrnli_qcGfQG808ebZeP.webm](https://github.com/user-attachments/assets/756c81d5-a946-414d-b851-0776a63a3419)


package.json
```diff
@@ -44,6 +44,8 @@
   },
   "packageManager": "yarn@4.0.2",
   "dependencies": {
+    "@types/animejs": "^3.1.12",
```
euharrison, line 47: @types should be `devDependencies`

But how about adding this only on `apps/namadillo`?
So, it'll be easier to eject each project later :)

apps/namadillo/src/App/Common/Timeline.tsx
```diff
@@ -59,8 +63,113 @@ export const Timeline = ({
...
+        scale: [0, 1],
+        opacity: [0, 1],
+      });
+    },
+    containerRef,
+    [complete]
+  );
+
+  useLayoutEffect(() => {
+    if (!containerRef.current) return;
```
euharrison, line 155: 🤔

euharrison (commented):
Really cool! 🌟

euharrison (approved):

package.json
pedrorezende: Ops, thanks for the catch! This should lie on apps/namadillo! Will make this change

apps/namadillo/src/App/Common/Timeline.tsx
pedrorezende: 🤔

PR #1459, opened 2025-01-02
feat: redirect to tx page

euharrison:
After signing a Namada transfer, redirects user to the transaction page.

This is mimicking the new implementation done on Ibc Transfers.

Please, notice this is not upgrading the timeline steps, neither the notifications. So existing issues are still happening. They will be addresses in the next PRs


pedrorezende (approved):
Neat :)

PR #1467, opened 2025-01-04
Namadillo: adding ibc notifications

pedrorezende:
Adding IBC notifications events when transaction gets confirmed or fails for any reason


euharrison (approved):
nice!

PR #1471, opened 2025-01-06
feat/746 - Updates for Ledger HW (shielded keys)

### LINKED ISSUE #746: Ledger: Add shielded address when importing account from Ledger HW Wallet
As of Ledger `NamadaApp` v0.0.27 (https://github.com/Zondax/ledger-namada/releases/tag/v0.0.20) we can generate a shielded address, but this is incorrect. We are awaiting `zip32` derivation in the Ledger. See https://github.com/Zondax/ledger-namada/pull/61 - we need the `xfvk` as returned in this PR.

- Update `Connect to Ledger` process to generate shielded address and viewing key for use with Ledger
- Update NPM package once published.

For testing, install the latest `Namada` app via the the Zondax Hub: https://hub.zondax.ch/

jurevans:
Resolves #746 

Also this implements a stepped process, so users can know what's happening on the ledger (now that there's 3 approvals instead of 1):
1. Transparent address & public key
2. Zip32 Viewing Key derivation
3. Zip32 Proof-Gen Key derivation

See https://github.com/anoma/namada-interface/pull/1458 from @mateuszjasiuk  - We'll also need these changes.

- [x] Update to latest `@zondax/ledger-namada` package (currently at `2.0.0`)
- [x] Remove HID transport (see: https://developers.ledger.com/docs/device-interaction/ledgerjs/integration/web-application/web-hid-usb - we should use either HID or USB, but not include support for both!). WebUSB seems to have the most support (supports older versions than WebHID), so we'll use that.
  - [WebUSB](https://caniuse.com/webusb)
  - [WebHID](https://caniuse.com/webhid)
- [x] Updates to Ledger class in SDK
  - [x] Methods for `getViewingKey` and `getProofGenerationKey` (split up `getShieldedKeys` so that we may turn these into steps, as they are separate approvals)
  - [x] Update type returned for ViewingKey (now returns `xfvk`)
  - [x] Before storing xfvk, convert to proper bech32m encoding (`zvknam1`)
- [x] Store `xfvk`, `ak` and `nsk`
- [x] It currently takes about 10 seconds to derive viewing key and proof-gen key (each!): We should notify the user what is happening, as there is not indication on Ledger when it is deriving these keys, they will only see each prompt after waiting about 10 seconds

### TESTING

- Run `yarn` on this branch to get updated packages
- Install `v2.0.2` of Ledger app (via https://hub.zondax.ch/namada for now) __NOTE__ This will require that you are on the latest firmware, so update if needed!
- Import Ledger account

### DESIGNS

**Approval steps**
image

### SCREENSHOTS

Version detection (NOTE: `2.0.0` is the correct min-version, I just did this as a test since I am already upgraded to the latest Ledger app):
Screen Shot 2025-01-09 at 6 44 34 AM


apps/extension/src/Setup/Common/Completion.tsx
```diff
@@ -20,7 +20,7 @@ type Props = {
   status?: CompletionStatus;
   statusInfo: string;
   parentAccountStore?: AccountStore;
-  shieldedAccount?: DerivedAccount;
+  paymentAddress?: string;
```
jurevans, line 23: was just simplifying here - we do not need to track the whole shielded account, only the payment address to display in confirmation

mateuszjasiuk (approved):
Tested and LGTM!

PR #1475, opened 2025-01-06
Namadillo: Top Navigation improvements

### LINKED ISSUE #1409: Proposal to simplify transfer of all assets for end users
Currently transfers are broken down into 

- IBC Transfers
- Internal inter account transfers
- Transfers to other accounts or namada users
-
This creates three transfer streams and three different transfer modules relying on end users knowing what IBC is vs a normal transfer, knowing that IBC is the way to shield from external wallets. + that shielding can also take place internally. 

Confusing if you are not crypto native

-------
Proposal!
We break transfers down into **user Actions** not assumptions of what they do and do not know

SHIELD / UNSHIELD / SEND

SHIELD
- Assets can be shielded via IBC or by sending from a users transparent account to shielded.
- The UI will determin whether it's an IBC transfer or an internal transfer based on the users selected actions

The module would just state whether the transfer is IBC vs Internal

e.g I want to shield some assets. I select to shield some osmo from keplr in the transfer UI = powered by IBC transfer
e.g I want to shield some assets. I select the ATOM from my transparent address in the UI = powered by Internal Namada transfer.

The tech that is used for each transfer is under the hood, the user doesn't care how the asset was shielded just that it happend

The same principle applies for unshielding just in the reverse on the destination input

Also for sending assets away from namada to other parties or wallets. A user could specify IBC vs non to send assets

UI wise then we have three distinct call to actions for users 

Screenshot 2024-12-13 at 12 30 13

And UI that just calls out IBC under the hood when IBC is necessary to facilitate a users action

Screenshot 2024-12-13 at 12 30 58

Small tweaks will be outlined to the transfer modules in due course.

pedrorezende:
- Adds a modal to help users navigate between options of shielding and unshielding
- Adds a copy to clipboard to transaction hash

Closes #1409


apps/namadillo/src/App/Layout/TopNavigation.tsx
```diff
@@ -87,6 +104,14 @@ export const TopNavigation = (): JSX.Element => {
...
         <KeplrAccount />
       </div>
+
+      {shieldingModalOpen && (
+        <ShieldAssetsModal onClose={() => setShieldingModalOpen(false)} />
+      )}
+
+      {unshieldingModalOpen && (
+        <UnshieldAssetsModal onClose={() => setUnshieldingModalOpen(false)} />
+      )}
```
euharrison, line 114: We could use the same navigation structure of `backgroundLocation` because we have other Shield buttons, then it'll be easier to reuse this implementation :)

euharrison (approved):
Nice!

apps/namadillo/src/App/Layout/TopNavigation.tsx
pedrorezende: I tried to keep it simple for now, do you have any other place that we have such a generic "shield" button that would justify moving this piece to a route?

euharrison: Tha yellow banner on the sidebar

euharrison: Oh, you are right, thanks for reminding it!

We also have this banner on Transparent Overview. But we need to check with @ChrisHoltDesign if we need to update this action as well

Screenshot 2025-01-07 at 15 53 55

ChrisHoltDesign: So that shield in the transparent assets section should ideally load the Internal shield option, not the IBC version

As it's an action to shield assets that are in your Namada transparent account

euharrison: Thanks Chris! 

So we are good having this modal only on top navigation :)

ChrisHoltDesign: Correct :)

PR #1479, opened 2025-01-07
Namadillo: querying next rpc if the previous one fails

### LINKED ISSUE #1464: Namadillo: query next RPC if the previous one fails
We use RPC list of endpoints provided by cosmos chain registry. Sometimes these endpoints are offline, or don't allow CORS. We need to query the next RPC endpoint if a previous one fails, and also save the reference of the working one on localStorage.

pedrorezende:
If we query an RPC endpoint and it fails for whatever reason (timeout, CORS, etc), we should query the next one in the list. 

Closes #1464

### DIFF
--- a/apps/namadillo/src/atoms/integrations/services.ts
+++ b/apps/namadillo/src/atoms/integrations/services.ts
@@ -24,7 +24,7 @@ import {
 } from "types";
 import { toBaseAmount } from "utils";
 import { getSdkInstance } from "utils/sdk";
-import { workingRpcsAtom } from "./atoms";
+import { rpcByChainAtom } from "./atoms";
 import { getRpcByIndex } from "./functions";
 
 type CommonParams = {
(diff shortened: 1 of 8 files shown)


euharrison (approved):
Really nice!

For Cosmos, I ended up using the lavenderfive, but the balance is zero on UI
It can be an issue with the TransferModule or the api, I'll investigate it better

Screenshot 2025-01-07 at 15 48 44

Screenshot 2025-01-07 at 15 48 38

Screenshot 2025-01-07 at 15 49 40

apps/namadillo/src/atoms/integrations/services.ts
```diff
@@ -142,28 +142,28 @@ export const submitIbcTransfer = async (
...
     return output;
   } catch (err) {
-    if (chain.chain_id in workingRpcs) {
-      delete workingRpcs[chain.chain_id];
-      set(workingRpcsAtom, { ...workingRpcs });
-    }
+    // On error, saves the next available rpc
+    set(rpcByChainAtom, {
+      ...lastRpc,
+      [chain.chain_id]: getRpcByIndex(chain, rpc.index + 1),
```
euharrison, line 165: Maybe in the future we could get from a random order. Loading all rpc urls into an atom and shuffling them. It could helps to don't have a bias for one or another rpc. But I see it is more work to do, so we can do it later

pedrorezende: We had that in the past, but we noticed that the first RPCs on the list were more reliable, so we switched to this solution

PR #1483, opened 2025-01-07
Namadillo: more stages to IBC transactions

pedrorezende:
Describes better on which stage of the transaction is being processed


euharrison (approved):
✅

PR #1488, opened 2025-01-08
feat: add transfer balance warning

euharrison:
Add transfer balance warning in case there are some balance but it's below the fee

This is important so user don't think Namadillo has a bug not fetching the balance

Screenshot 2025-01-08 at 16 26 50


pedrorezende (commented):
Good one! But I was talking to @ChrisHoltDesign about displaying this validation message in the button text. For me it's approved in terms of code, but I'll let him take a look on that first before raising the green flag :)

euharrison:
@pedrorezende he gave the green flag :) 

The button will preserve the same current flow

In case the user don't fill the amount it shows "Define an amount to transfer"
Screenshot 2025-01-10 at 11 23 18


And in case it fills any amount, it shows "Not enough balance"
Screenshot 2025-01-10 at 11 23 11

pedrorezende (approved):

PR #1506, opened 2025-01-13
feat/1442 - Extension: Updates to MASP Tx Details

### LINKED ISSUE #1442: Display MASP TX details in the confirmation popup
Details to display:

- General Tx props (wrapper Tx, like all other Tx, fees, fee token, etc.)
- Commitments
  - Sources
    - `owner`
    - `token`
    - `amount`
  - Targets
    - `owner`
    - `token`
    - `amount`
- Shielded section hash (we should probably encode this as a string on Details, probably hex)

**TODO** What is left? We are currently displaying the above ^ for Shielding, confirming the other MASP tx...

List of Txs overview should indicate what type of shielded transfer (if it doesn't already, should be tested and confirmed to do the following):

- Shielding (Transparent -> Shielded)
- Shielded (Shielded -> Shielded)
- Unshielding (Shielded -> Transparent)

__NOTE__ It is already set up to do this, so this issue should primarily be about the Tx Data view, and what can be decoded from the submitted Tx to display.

jurevans:
Resolves #1442 

- [x] Display shielded section hash correctly
- [x] Detect MASP transfer type correctly
  - [x] Shielding
  - [x] Unshielding
  - [x] Shielded

### DIFF
--- a/apps/extension/src/Approvals/Commitment.tsx
+++ b/apps/extension/src/Approvals/Commitment.tsx
@@ -22,7 +22,7 @@ import { TransactionCard } from "./TransactionCard";
 
 type CommitmentProps = {
   commitment: CommitmentDetailProps;
-  wrapperFeePayer?: string;
+  wrapperFeePayer: string;
 };
 
 const IconMap: Record<TxType, React.ReactNode> = {
@@ -61,7 +61,7 @@ const formatTransferAddress = (address: string): string =>
 
 const renderContent = (
   tx: CommitmentDetailProps,
-  wrapperFeePayer?: string
+  wrapperFeePayer: string
 ): ReactNode => {
   switch (tx.txType) {
     case TxType.Bond:
--- a/apps/extension/src/Approvals/Commitment.tsx
+++ b/apps/extension/src/Approvals/Commitment.tsx
@@ -17,11 +17,7 @@ import { FaVoteYea } from "react-icons/fa";
 import { FaRegEye, FaWallet } from "react-icons/fa6";
 import { GoStack } from "react-icons/go";
 import { PiDotsNineBold } from "react-icons/pi";
-import {
-  hasShieldedSection,
-  parseTransferType,
-  ShieldedPoolLabel,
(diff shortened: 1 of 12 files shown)


mateuszjasiuk (approved):
LGTM! :)

apps/extension/src/Approvals/Commitment.tsx
```diff
@@ -17,15 +17,12 @@ import { FaVoteYea } from "react-icons/fa";
...
-  hasShieldedSection,
-  parseTransferType,
-  ShieldedPoolLabel,
-} from "utils";
+import { isShieldedPool, parseTransferType, ShieldedPoolLabel } from "utils";
 import { TransactionCard } from "./TransactionCard";
 
 type CommitmentProps = {
   commitment: CommitmentDetailProps;
+  wrapperFeePayer?: string;
```
jurevans, line 25: Question: Wouldn't `wrapperFeePayer` always be defined here? Even if we don't provide it as an initial Tx argument, I think this will always be set to an address by the time it is deserialized. If that's the case, this wouldn't need to be optional, but I'm not completely sure

mateuszjasiuk: Yeah I think it's always there. I can typecast :D

apps/extension/src/Approvals/Commitment.tsx
```diff
@@ -59,7 +56,13 @@ const TitleMap: Record<TxType, string> = {
 const formatAddress = (address: string): string =>
   shortenAddress(address, 6, 6);
 
-const renderContent = (tx: CommitmentDetailProps): ReactNode => {
+const formatTransferAddress = (address: string): string =>
+  isShieldedPool(address) ? ShieldedPoolLabel : formatAddress(address);
+
+const renderContent = (
+  tx: CommitmentDetailProps,
+  wrapperFeePayer?: string
```
jurevans, line 64: I think this `wrapperFeePayer` doesn't need to be optional

PR #1534, opened 2025-01-13
feat: add version warning banner (anoma#1491)

h3ll0ofr1end:
This PR introduces a feature to verify the interface version for compatibility by dynamically fetching the required version from the remote package.json file hosted on GitHub.

- Introduced checkInterfaceCompatibilityError to handle interface version validation.
- Enhanced the useCompatibilityErrors hook to include this new check.
- The required interface version is fetched dynamically, ensuring accuracy and up-to-date compatibility.
- Displays a warning banner if the interface version is outdated or incompatible.

This update aligns the interface version check with the existing validations for the indexer and keychain, ensuring consistency across compatibility checks.

by 5ElementsNodes

### DIFF
--- a/apps/namadillo/src/utils/compatibility.tsx
+++ b/apps/namadillo/src/utils/compatibility.tsx
@@ -2,6 +2,7 @@ import { routes } from "App/routes";
 import compatibility from "compatibility.json";
 import { wallets } from "integrations";
 import { Link } from "react-router-dom";
+import semver from "semver";
 import semverSatisfies from "semver/functions/satisfies";
 import semverLtr from "semver/ranges/ltr";
 import { version as interfaceVersion } from "../../package.json";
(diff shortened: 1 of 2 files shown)


apps/namadillo/src/utils/compatibility.tsx
```diff
@@ -102,3 +113,32 @@ export const checkKeychainCompatibilityError = (
...
+    const [currentMajor, currentMinor] = interfaceVersion
+      .split(".")
+      .map(Number);
+    const [requiredMajor, requiredMinor] = requiredInterfaceVersion
+      .split(".")
+      .map(Number);
+
+    if (
+      requiredMajor > currentMajor ||
+      (requiredMajor === currentMajor && requiredMinor > currentMinor)
```
euharrison, line 131: How about using the `semver` package to compare this?
It'll be easier to read and probably handling border cases as well

euharrison (approved):
Nice! Thanks!

apps/namadillo/src/utils/compatibility.tsx
h3ll0ofr1end: Agreed with you.
Using the **compatibility.json** file (currently used for the other checks) to exclude patches from the Interface check invalidates the check itself, because it's a local file.
Excluded `semver` for this reason, but maybe we can use an internal normalize function to achieve the same result. On it.

pedrorezende:
@h3ll0ofr1end Thank you so much for the contribution!

PR #1545, opened 2025-01-14
feat: fetching known ibc channels from github chain-registry repository

pedrorezende:
Tries to fetch known IBC channels from namada-chain-registry repository, instead of relying on the ibc channels added to Namadillo build. Maybe in the future we can revert these changes, but now we'll need a more dynamic way to fetch the latest changes and info from relayers.

This PR also introduce new config.toml entries:

github_chain_registry_base_url = "https://raw.githubusercontent.com/anoma/namada-chain-registry/refs/heads/main"
github_namada_interface_url = "https://raw.githubusercontent.com/anoma/namada-interface/refs/heads/main"

### DIFF
--- a/apps/namadillo/public/config.toml
+++ b/apps/namadillo/public/config.toml
@@ -3,3 +3,6 @@
 #rpc_url = ""
 #masp_indexer_url = ""
 #localnet_enabled = false
+
+github_chain_registry_base_url = "https://raw.githubusercontent.com/anoma/namada-chain-registry/refs/heads/main"
+github_namada_interface_url = "https://raw.githubusercontent.com/anoma/namada-interface/refs/heads/main"
--- a/apps/namadillo/src/atoms/integrations/functions.ts
+++ b/apps/namadillo/src/atoms/integrations/functions.ts
@@ -17,6 +17,7 @@ import * as osmosisTestnet from "chain-registry/testnet/osmosistestnet";
 import * as stargazeTestnet from "chain-registry/testnet/stargazetestnet";
 import { DenomTrace } from "cosmjs-types/ibc/applications/transfer/v1/transfer";
 import { TransactionPair, buildTxPair } from "lib/query";
+import namadaMainnetChain from "namada-chain-registry/namada/chain.json";
 import {
   Address,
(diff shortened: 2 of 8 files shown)


apps/namadillo/public/config.toml
```diff
@@ -3,3 +3,6 @@
 #rpc_url = ""
 #masp_indexer_url = ""
 #localnet_enabled = false
+
+github_chain_registry_base_url = "https://raw.githubusercontent.com/anoma/namada-chain-registry/refs/heads/main"
+github_namada_interface_url = "https://raw.githubusercontent.com/anoma/namada-interface/refs/heads/main"
```
euharrison, line 8: I didn't understand the use case of this 🤔

apps/namadillo/src/atoms/integrations/functions.ts
```diff
@@ -43,21 +44,20 @@ import internalDevnetChain from "namada-chain-registry/_testnets/namadainternald
...
+export const namadaTestnetChainList = [
   internalDevnetChain,
   campfireChain,
   housefireChain,
   housefireOldChain,
-  namadaChain
-);
+] as Chain[];
+
+cosmosRegistry.chains.push(...namadaTestnetChainList, namadaChain);
```
euharrison, line 60: I vaguely remember having a `find()` in the code that captures the first item on array, and it needs to be the mainnet, maybe an asset related?

Do you think we could invert parameters to be `.push(namadaChain, ...namadaTestnetChainList)`?

euharrison (approved):
Nice! Very good!

jurevans (approved):
LGTM!

apps/namadillo/public/config.toml
pedrorezende: Just in case someone would want to point to their own chain registry for some reason. Probably won't be used

apps/namadillo/src/atoms/integrations/functions.ts
pedrorezende: Hmmm, I can do it, but this `find()` is not right :/ if we find this case, we should refactor

PR #1547, opened 2025-01-15
feat: fetch shielded balances on transfer

### LINKED ISSUE #1484: Namadillo: Disable MASP transfers until synced
Namadillo currently allows initiating a MASP transfer while the shielded-sync is still processing. That will of course cause an error.

image

Namadillo should signal to the user that shielded-sync is busy and prevent the user submitting a transfer. Allowing the user to fill form input fields during syncing could be acceptable, although the balance will be outdated. Although user may switch to the MASP page to check the sync status and then lose the form input. Disabling form input may prevent user from wasting time.

Maybe shielded-sync status could be included in the current RPC/indexer sync status indicator.

### LINKED ISSUE #1493: Namadillo: MASP balance loading hangs when you access the page directly
Maybe it's a race condition with the other balances loading

### LINKED ISSUE #1495: Update balances after any transfer
Update balances after transfer

Please check all these pages:
- Overview (Home)
- Staking
- Masp

With these transfers:
- Transparent
- Shielding
- Shielded
- Unshield
- Ibc To Namada (deposit)
- Ibc From Namada (withdraw)

### LINKED ISSUE #1497: Add a loading indicator for shielded balances on overview
Currently, we are showing 0 (zero) as a balance while the shielded sync is running

We should do one or another:

1 - Cache the value and show it while it's loading
or
2 - Show a loading indicator while the value is stale

### LINKED ISSUE #1498: Fix dollar values
Currently, the dollar values on Overview is showing 0(zero), and on Masp it's showing N/A

But the dollar values are showing correctly on the Transparent Overview Table and Shielded Overview Table

euharrison:
This PR resolves most of the current issues related to the shielded balance

- Re-run the shielded sync and the shielded balance after any successful transfer. 
- The balance is now saved on local storage and we can provide the last cached value
- Added a shielded sync indicator on the sidebar that shows the percentage of the shielded sync on any Masp page and the Overview
- Also, update the indicator text when the shielded sync is converting assets, which currently takes a longer time and we don't have a progress indicator on the sdk side
- Fix the dollar value considering the NAM as zero until it's available on osmosis api. 
- Add a 2h timeout for transfers that do not completed and aren't available on chain for whatever reason

closes #1484
closes #1493
closes #1495
closes #1497 
closes #1498 

https://github.com/user-attachments/assets/464c039b-4da7-4b7c-a608-acc455fdd543

### DIFF
--- a/apps/namadillo/src/App/AccountOverview/AccountOverview.tsx
+++ b/apps/namadillo/src/App/AccountOverview/AccountOverview.tsx
@@ -1,6 +1,7 @@
 import { Panel } from "@namada/components";
 import { ConnectPanel } from "App/Common/ConnectPanel";
 import { PageWithSidebar } from "App/Common/PageWithSidebar";
+import { ShieldedSyncProgress } from "App/Masp/ShieldedSyncProgress";
 import { EpochInformation } from "App/Sidebars/EpochInformation";
 import MainnetRoadmap from "App/Sidebars/MainnetRoadmap";
 import { ShieldAllBanner } from "App/Sidebars/ShieldAllBanner";
@@ -55,6 +56,7 @@ export const AccountOverview = (): JSX.Element => {
       </div>
       <aside className="flex flex-col gap-2">
         <EpochInformation />
+        <ShieldedSyncProgress />
         <ShieldAllBanner />
         <MainnetRoadmap />
       </aside>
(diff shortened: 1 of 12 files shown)


pedrorezende (approved):
Very good improvements! But I still feel we would need something better to handle the masp updates. Like an optmistic updates. I'll think of a solution too

PR #1548, opened 2025-01-15
fix: fix fee component on transfers

### LINKED ISSUE #1492: Small UI tweaks on Namadillo
The underscore should be visible across all the wording including the value.

lets add a rollover to yellow

Screenshot 2025-01-10 at 16 41 13

### LINKED ISSUE #1501: Gas fee for Transparent transfer is using the fee of Shielded transfer
Screenshot 2025-01-10 at 15 22 45

euharrison:
- Fix Namada Transfer fee adding the `txKind` to transfer query key
- Allow changing fee token for Ibc Withdraw
- Update the underline and hover of the <TransactionFee> button

Screenshot 2025-01-15 at 14 23 26

Screenshot 2025-01-15 at 14 22 11

closes #1501
closes #1492

### DIFF
--- a/apps/namadillo/src/App/Transfer/TransferDestination.tsx
+++ b/apps/namadillo/src/App/Transfer/TransferDestination.tsx
@@ -47,6 +47,8 @@ export const TransferDestination = ({
   openChainSelector,
   openProviderSelector,
 }: TransferDestinationProps): JSX.Element => {
+  const isExternalDestination = !!openChainSelector;
+
   return (
     <div
       className={clsx("relative bg-neutral-800 rounded-lg px-4 pt-8 pb-4", {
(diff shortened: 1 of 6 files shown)


apps/namadillo/src/App/Transfer/TransferDestination.tsx
```diff
@@ -47,6 +47,8 @@ export const TransferDestination = ({
   openChainSelector,
   openProviderSelector,
 }: TransferDestinationProps): JSX.Element => {
+  const isExternalDestination = !!openChainSelector;
```
pedrorezende, line 50: I'm not sure this is the correct way of verifying that we can define the fee token. I think we should verify if the transfer is coming from Namada instead

pedrorezende (changes requested):

apps/namadillo/src/App/Transfer/TransferDestination.tsx
euharrison: yeah, agreed. I added the prop `changeFeeEnabled` to have a better control over it 👍

pedrorezende (approved):
LGTM!

PR #1562, opened 2025-01-16
feat: support shielded tx with ledger 

mateuszjasiuk:
Support for ledger masp txs with disposable gas payer.

Testing:
- signing txs with or without ledger and see if it works :D


jurevans (approved):
Code looks good! Was able to successfully sign & submit each MASP transfer type!

PR #1567, opened 2025-01-17
feat: fixing ibc shielded gas estimate

### LINKED ISSUE #1486: Namadillo: Gas estimation for IBC MASP is too low
The gas estimation when trying to make an IBC MASP transaction is too low. 

image

We can manually increase that on Keplr as a temporary workaround, but we need to provide a better estimation.

pedrorezende:
First version of estimate gas calculation for IBC shielded transfers. Still requires some minor improvements / tweaks that I'll leave for later PRs, like loading state or refresh button.

I also added 5% more over the estimated gas just to make sure that the transaction will succeed. Maybe we can tweak this percentage using the gas usage selection modal /cc @euharrison

Closes #1486

### DIFF
--- a/apps/namadillo/src/utils/index.ts
+++ b/apps/namadillo/src/utils/index.ts
@@ -91,9 +91,7 @@ export const toDisplayAmount = (
   if (!displayUnit) {
     return baseAmount;
   }
-  return baseAmount
-    .shiftedBy(-displayUnit.exponent)
-    .decimalPlaces(6, BigNumber.ROUND_DOWN);
+  return baseAmount.shiftedBy(-displayUnit.exponent);
 };
 
 export const toBaseAmount = (
--- a/apps/namadillo/src/utils/index.ts
+++ b/apps/namadillo/src/utils/index.ts
@@ -91,7 +91,9 @@ export const toDisplayAmount = (
   if (!displayUnit) {
     return baseAmount;
   }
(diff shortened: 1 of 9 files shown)


apps/namadillo/src/utils/index.ts
```diff
@@ -91,7 +91,9 @@ export const toDisplayAmount = (
   if (!displayUnit) {
     return baseAmount;
   }
-  return baseAmount.shiftedBy(-displayUnit.exponent);
+  return baseAmount
+    .shiftedBy(-displayUnit.exponent)
+    .decimalPlaces(6, BigNumber.ROUND_DOWN);
```
euharrison, line 96: we are using this function to set some value for any transfer, I'm a bit afraid of this imprecision spread to border cases. Shouldn't we apply this only on the visual components on the render phase?

euharrison (approved):
Nice! 

How about defining a fallback value just in case we can't estimate (like below with ATOM)?

Screenshot 2025-01-17 at 15 40 40

apps/namadillo/src/utils/index.ts
pedrorezende: Changed. The error on Cosmos is because we can't estimate the gas required using an invalid channel. I've tried to fix that, but we don't have a right way to calculate a fallback for that. Before, the amount of gas used was hardcoded :/

PR #1572, opened 2025-01-20
feat: disable submit button for masp tx when using disconnected ledger device

mateuszjasiuk:
` quit app -> open app -> unplug device -> plug device in and open app again`

https://github.com/user-attachments/assets/8a5cc3b4-22f3-4f76-9f06-f6bacdd23bd2

### DIFF
--- a/apps/namadillo/src/atoms/ledger/atoms.ts
+++ b/apps/namadillo/src/atoms/ledger/atoms.ts
@@ -1,0 +1,67 @@
+import { ledgerUSBList } from "@namada/sdk/web";
+import { LedgerError } from "@zondax/ledger-namada";
+import { isLedgerAccountAtom } from "atoms/accounts";
+import { atom } from "jotai";
+
+import { atomWithQuery } from "jotai-tanstack-query";
+import { getSdkInstance } from "utils/sdk";
+
+export type LedgerStatus = {
+  connected: boolean;
+  errorMessage: string;
+};
+
+export const ledgerStatusAtom = atomWithQuery<LedgerStatus | undefined>(() => {
+  return {
+    refetchInterval: 1000,
+    queryKey: ["ledger-status"],
+    queryFn: async () => {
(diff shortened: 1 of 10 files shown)


apps/namadillo/src/atoms/ledger/atoms.ts
```diff
@@ -0,0 +1,67 @@
...
+export type LedgerStatus = {
+  connected: boolean;
+  errorMessage: string;
+};
+
+export const ledgerStatusAtom = atomWithQuery<LedgerStatus | undefined>(() => {
+  return {
+    refetchInterval: 1000,
+    queryKey: ["ledger-status"],
+    queryFn: async () => {
```
jurevans, line 18: I think something may need to change here after testing. What I found was that this loop continues to execute _after_ I've connected and submitted a Tx, and I hit a few cases with different errors thrown. It will definitely be an issue if we try to open transport on Web when the Keychain is attempting to open transport to sign. I'll give it some thought! Maybe if we, say, detect it once, if it's undetected, require the user to click to _Connect_, once connected, fetch build params then submit. I think maybe it would be better if validation happens once, or, if the check is discontinued after first successful connection? Not sure, will take some thought

jurevans:
@mateuszjasiuk looks good! I did encounter some weird behavior when trying to Unshield, I think related to the loop and transport management. I think it's leading to edge cases like these:

```
TransportInterfaceNotAvailable: Failed to execute 'claimInterface' on 'USBDevice': The device must be opened first.
```

```
InvalidStateError: Failed to execute 'close' on 'USBDevice': An operation that changes interface state is in progress.
```

We might want to require the user to initiate connection to Ledger to have better management of transport open & close, or ensure that transport is closed once we have successfully connected. We can chat about it though!

jurevans (approved):
Code looks good, retested the Unshielding tx, and is working great with Ledger!

PR #1573, opened 2025-01-20
chore: bump namada_sdk to current main and indexer client to 1.1.6

mateuszjasiuk:
- bumps namada_sdk to the latest commit from main
- bumps namada indexer client to 1.1.6
- fixes types errors
- adds "voted unknown" label - this happens when someone voted then redelegated then voted again, this should not happen too often
- unified api configuration - sth changed and we have to override default headers produced by indexer client 
- fixed small problem with shielded sync not running when masp indexer url is not present
- fixed small issue in viewing keys atom which was checking shielded account by alias

obraz


pedrorezende (approved):

jurevans (approved):
LGTM!

PR #1574, opened 2025-01-20
feat: Improvements on IBC withdrawal

pedrorezende:
This PR contains a few improvements on IBC withdrawals and also a few refactors and fixes for IBC queries


apps/namadillo/src/hooks/useTransaction.tsx
```diff
@@ -100,58 +98,62 @@ export const useTransaction = <T,>({
...
+        }
+      });
 
-      if (onError) {
-        onError(err);
-      } else {
-        throw err;
-      }
-    }
+    onBroadcasted?.();
```
euharrison, line 144: We don't have an `await` on the `broadcastTxWithEvents`, so I'm thinking if we should call it like `onStart` or something similar? 

Or, maybe, the `onSuccess` should be renamed to `onBroadcasted`

euharrison (approved):

PR #1575, opened 2025-01-21
Ledger MASP Integration

jurevans:
Details _TBD_


apps/namadillo/src/atoms/ledger/atoms.ts
```diff
@@ -0,0 +1,75 @@
...
+    queryKey: ["ledger-status"],
+    queryFn: async () => {
+      const devices = await ledgerUSBList();
+
+      if (devices.length > 0) {
+        try {
+          // Disable console.warn to prevent the warning from showing up in the UI in the loop
+          // eslint-disable-next-line @typescript-eslint/no-explicit-any
+          (console as any).warnOld = console.warn;
+          console.warn = () => {};
```
euharrison, line 28: I see it has the condition `isLedgerAccount` before doing this. But I'm felling it could create border cases in the future

What do you think if we only skip the undesired message with a filter? Or even just accept the warn message?

If filter, something like this

```ts
console.warn = (...args) => {
   if (String(args[0]).startsWith("something")){
     return;
   }
   console.warnOld(...args)
};
```

euharrison (approved):
Nice!

I don't have a ledger device, but I tested creating a new account, sending some nams on campfire to it, shielding, deleting the account, importing the spending key only, checking the balance zero for transparent and balance correct on masp, switching account, viewing spending key, and accessing the ledger settings modal

apps/namadillo/src/atoms/ledger/atoms.ts
jurevans: @mateuszjasiuk pinging you since you'll have a better idea what is going on here

jurevans:
> Nice!
> 
> I don't have a ledger device, but I tested creating a new account, sending some nams on campfire to it, shielding, 

@euharrison thanks for testing! It's good to get some regression testing in regardless

pedrorezende (approved):
LGTM! Tested and didn't notice any regression in any other part of the app

PR #1582, opened 2025-01-23
feat: adding multicore as default option for Namadillo and publishing sdk packages automatically to npm

pedrorezende:
- Adding multicore to Namadillo build
- Publishing SDK to NPM repository


pedrorezende:
Not merging this until we fix ibc transactions with multicore enable

PR #1587, opened 2025-01-27
refactor: updating refereces of namada chain registry

pedrorezende:
Change references of namada-chain-registry to @namada/chain-registry


jurevans (approved):
Nice!

PR #1592, opened 2025-01-30
feat: check for ibc timeout if a response was not provided by the chain

pedrorezende:

### DIFF
--- a/apps/namadillo/src/atoms/integrations/services.ts
+++ b/apps/namadillo/src/atoms/integrations/services.ts
@@ -9,15 +9,15 @@ import {
   StargateClient,
   StdFee,
 } from "@cosmjs/stargate";
-import * as Comlink from "comlink";
-import { TxRaw } from "cosmjs-types/cosmos/tx/v1beta1/tx";
-
 import { sanitizeUrl } from "@namada/utils";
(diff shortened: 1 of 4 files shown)


apps/namadillo/src/atoms/integrations/services.ts
```diff
@@ -176,6 +177,77 @@ export const queryAndStoreRpc = async <T>(
...
+  let isTimeout = isTransferLocalTimeout(tx);
+  let timeoutHash = "";
+
+  if (!isTimeout) {
+    const timeoutQuery = await queryForIbcTimeout(client, tx);
+    isTimeout = timeoutQuery.length > 0;
+    timeoutHash = timeoutQuery[0]?.hash || "";
+  }
+
+  return [isTimeout, timeoutHash];
```
euharrison, line 248: I'm thinking if we should use object on these cases. The array tuple is great for useState because it's a well know function and we rename the returned values. But, on other functions, it's harder to read and understanding the code, also, usually we don't rename the value, actually, we use the same name 
```tsx
const [isTimeout, timeoutHash] = await checkForIbcTransferTimeout(client, tx);
```

euharrison (approved):
Nice!

PR #1594, opened 2025-01-31
feat: improving masp transactions feedback for users and other gas estimation fixes

pedrorezende:
This PR refactors all the transactions logic separating the build step from the signing step. With this we will be able to improve user experience while waiting for long steps, like the generation of masp params. I've only implemented multiple status on Shielding because we're going to change a few things on the UI in the next PRs.

It also creates a quick workaround to increase gas fees while we improve the indexer api for estimate. The reason behind it is that many transactions were frequently failing because of gas (ex: unstake, re-delegate, etc)

### DIFF
--- a/apps/namadillo/src/App/Masp/MaspShield.tsx
+++ b/apps/namadillo/src/App/Masp/MaspShield.tsx
@@ -120,7 +120,7 @@ export const MaspShield: React.FC = () => {
   };
 
   return (
-    <Panel className="relative min-h-[600px]">
+    <Panel className="relative min-h-[600px] flex-1">
       <header className="flex flex-col items-center text-center mb-3 gap-6">
         <h1 className="mt-6 text-lg text-yellow">Shield</h1>
         <NamadaTransferTopHeader
--- a/apps/namadillo/src/App/Masp/MaspShield.tsx
+++ b/apps/namadillo/src/App/Masp/MaspShield.tsx
@@ -26,9 +26,10 @@ import { Address, TransferTransactionData } from "types";
 export const MaspShield: React.FC = () => {
   const navigate = useNavigate();
   const [searchParams, setSearchParams] = useSearchParams();
+  const { storeTransaction } = useTransactionActions();
   const [displayAmount, setDisplayAmount] = useState<BigNumber | undefined>();
   const [generalErrorMessage, setGeneralErrorMessage] = useState("");
-
+  const [currentStep, setCurrentStep] = useState("");
   const rpcUrl = useAtomValue(rpcUrlAtom);
   const chainParameters = useAtomValue(chainParametersAtom);
   const defaultAccounts = useAtomValue(allDefaultAccountsAtom);
@@ -36,8 +37,6 @@ export const MaspShield: React.FC = () => {
   const { data: availableAssets, isLoading: isLoadingAssets } = useAtomValue(
     namadaTransparentAssetsAtom
   );
-
-  const { storeTransaction } = useTransactionActions();
 
   const chainId = chainParameters.data?.chainId;
   const sourceAddress = defaultAccounts.data?.find(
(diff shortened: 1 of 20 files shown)


apps/namadillo/src/App/Masp/MaspShield.tsx
```diff
@@ -26,9 +26,10 @@ import { Address, TransferTransactionData } from "types";
 export const MaspShield: React.FC = () => {
   const navigate = useNavigate();
   const [searchParams, setSearchParams] = useSearchParams();
+  const { storeTransaction } = useTransactionActions();
   const [displayAmount, setDisplayAmount] = useState<BigNumber | undefined>();
   const [generalErrorMessage, setGeneralErrorMessage] = useState("");
-
+  const [currentStep, setCurrentStep] = useState("");
```
euharrison, line 32: Can you apply this on MaspUnshield and NamadaTransfer as well? Or is this part of a future PR?

apps/namadillo/src/App/Staking/IncrementBonding.tsx
```diff
@@ -80,10 +80,6 @@ const IncrementBonding = (): JSX.Element => {
         </>
       ),
     }),
-    parseErrorTxNotification: () => ({
```
euharrison, line 83: If we remove this, the notification failed doesn't appear. I tested this trying to stake with a custom small amoutn of gas (0.01 NAM) and you can observe the notification there forever

Screenshot 2025-01-31 at 14 06 47

apps/namadillo/src/atoms/fees/atoms.ts
```diff
@@ -51,6 +52,18 @@ export const gasEstimateFamily = atomFamily(
             counter("RevealPk"),
             counter("Redelegate"),
           ]);
+
+          // TODO: we need to improve this estimate API. Currently, gasEstimate.min returns
+          // the minimum gas limit ever used for a TX, and avg is failing in most of the transactions.
+          // So we estabilish the average value between gasEstimate.avg and gasEstimate.max as the min value.
+          // The other values derive from this one.
+          const newMin = Math.ceil((gasEstimate.avg + gasEstimate.max) / 2);
```
euharrison, line 60: I understand the reason, but I'm afraid of this math creates an infinite loop of gas growing infinitely. We need to discuss this with other folks, but I'm good for now on campfire

For me, this endpoint should be the return value of a dry-run, then we apply a const like the 1.25 and 1.5 you did

apps/namadillo/src/lib/query.ts
```diff
@@ -96,30 +96,32 @@ export const buildTx = async <T>(
...
-  const { tx } = sdk;
   const wrapperTxProps = getTxProps(account, gasConfig, chain, memo);
   const txs: TxMsgValue[] = [];
   const txProps: TxProps[] = [];
 
   // Determine if RevealPK is needed:
-  if (!publicKeyRevealed) {
-    const revealPkTx = await tx.buildRevealPk(wrapperTxProps);
-    txs.push(revealPkTx);
+  if (!skipRevealPk) {
```
euharrison, line 107: just to avoid "not-not", how about renaming this to `revealPk?: boolean = true` and `if (revealPk)`?

apps/namadillo/src/hooks/useTransfer.ts
```diff
@@ -23,13 +23,14 @@ type useTransferParams = {
...
 
 type useTransferOutput = (
-  | useTransactionOutput<TransparentTransferMsgValue>
-  | useTransactionOutput<ShieldedTransferMsgValue>
-  | useTransactionOutput<ShieldingTransferMsgValue>
-  | useTransactionOutput<UnshieldingTransferMsgValue>
+  | UseTransactionOutput<TransparentTransferMsgValue>
+  | UseTransactionOutput<ShieldedTransferMsgValue>
+  | UseTransactionOutput<ShieldingTransferMsgValue>
+  | UseTransactionOutput<UnshieldingTransferMsgValue>
```
euharrison, line 33: ops, thanks!

euharrison (approved):
Really cool! Nice!

Left a few comments, but my main concern is the notifications. But it's also ok if we address this in a new PR

apps/namadillo/src/App/Masp/MaspShield.tsx
pedrorezende: Hello! This is going to be part of a PR I'm working now. We're going to change the whole interface after submitting the transaction, that's why I didn't apply this change to other pages

PR #1600, opened 2025-01-31
fix: empty reward storage

euharrison:
Fix race condition when the user has empty value for rewards

Screenshot 2025-01-31 at 12 22 02


jurevans (approved):
Nice!

Issue #1551, opened 2025-01-15
Namadillo: Update shielded sync indicator layout


euharrison:
Update the background progress like the imagem below

![Screenshot 2025-01-15 at 16 35 03](https://github.com/user-attachments/assets/82c28060-4fda-4b83-9c93-0a84044afaaf)

PR #1601, opened 2025-01-31
feat: update shield sync indicator layout

euharrison:
- Update shielded sync indicator layout
- Create the shared `<Sidebar>` component

Closes: #1551

https://github.com/user-attachments/assets/790fb8fe-1fa4-4253-9c8b-44bc82182570


pedrorezende (approved):
LGTM, however I'm seeing an error when we already have masp fully synced. We can address it in a later PR if you want. If this is the case, could you open a new issue?
[scrnli_Of4Ap3o8p7q41S.webm](https://github.com/user-attachments/assets/3c542731-fe03-4315-8998-657d419b4b36)

euharrison:
Thanks for the review! This is partially correct, because we refetch every time we access the MASP page. But I removed the animation when goes to zero, so it looks less weird 🙂

PR #1609, opened 2025-02-05
feat: add dollar support for multiple channels

### LINKED ISSUE #1608: Render the dollar price when we have two channels for the same token
Image

euharrison:
We can have more than one channel for the same base denom

For these cases, we need to fetch and map all dollar values to all possible addresses

We also removed the `N/A` in case it's missing a token dollar value for whatever reason. It simplifies how user understand these values and don't looks like buggy

closes #1608

Screenshot 2025-02-05 at 11 52 46

Screenshot 2025-02-05 at 13 04 44


pedrorezende (approved):
LGTM!

PR #1615, opened 2025-02-07
feat: add optimistic response for shielded transfers

### LINKED ISSUE #1553: Namadillo: Add optimistic response for transfer
When signing a transfer, we could update de balance with a optimistic response. 

This is more meaningful for shielded and ibc transfers where it takes many seconds to complete

euharrison:
Add optimistic response for shielded transfers

Plus, show more detailed information on transfer history: chain_id in case the name is missing on registry, and target in case of shielded transfers

closes #1553

https://github.com/user-attachments/assets/c2b752ac-b3ed-415b-b734-888eefbb495f

### DIFF
--- a/apps/namadillo/src/App/Transactions/TransactionCard.tsx
+++ b/apps/namadillo/src/App/Transactions/TransactionCard.tsx
@@ -48,7 +48,7 @@ export const TransactionCard = ({
   const chainName =
     chainId in availableChains ?
       parseChainInfo(availableChains[chainId].chain)?.pretty_name
-    : "Unknown";
+    : chainId;
 
   return (
     <article
--- a/apps/namadillo/src/atoms/balance/atoms.ts
+++ b/apps/namadillo/src/atoms/balance/atoms.ts
@@ -107,7 +107,7 @@ export const viewingKeysAtom = atomWithQuery<
 });
 
 export const storageShieldedBalanceAtom = atomWithStorage<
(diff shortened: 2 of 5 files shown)


apps/namadillo/src/App/Transactions/TransactionCard.tsx
pedrorezende, line 1: Good solution!

apps/namadillo/src/atoms/balance/atoms.ts
```diff
@@ -107,7 +107,7 @@ export const viewingKeysAtom = atomWithQuery<
 });
 
 export const storageShieldedBalanceAtom = atomWithStorage<
-  Record<Address, { address: string; minDenomAmount: BigNumber }[]>
+  Record<Address, { address: Address; minDenomAmount: string }[]>
```
pedrorezende, line 110: I think we should store amount as BigNumber and save unnecessary conversions later

pedrorezende (approved):
Very nice! :)

apps/namadillo/src/atoms/balance/atoms.ts
euharrison: @pedrorezende I agree, but this is the storage, meaning it's saving on LocalStorage as a string, so, unfortunately, we can't save it as a BigNumber

I found this issue because we are now manipulating this cache with the optimistic response and previously it don't have it, the it never trigger it

PR #1617, opened 2025-02-10
feat: several improvements on Transfer Module and transaction receipts

pedrorezende:
This PR contains an improved version of the transfer module, separating the submit process into different steps and with a new success animation. It also adds a new structure for transaction receipt, when you check the history of transactions


euharrison (approved):
Amazing! So cool! 👏

PR #1624, opened 2025-02-11
feat: Update Keychain for new modified-zip32

### LINKED ISSUE #1611: Extension: Update modified Zip32 derivation path to match CLI
The extension should use a different default zip32 derivation path for Shielded accounts, to match update in CLI.

See: https://github.com/anoma/namada/pull/4307

Essentially, the private key from which shielded keys are derived should itself be derived from `m/44'/877'/0'/0'/2147483647'`, then from that, we can derive `m/32'/877'/0'` Zip32 paths.

## NOTES

This has some consequences that we'll need to consider on the interface-side of things:

- Since shielded accounts are no longer derived from the same private key as transparent accounts, it doesn't make sense to _automatically_ generate these.
- As such, it makes a lot more sense to have these as separate flows as there's no longer any relation between these keys

1. Transparent: User can import/generate mnemonic, and derive with a default (or custom) BIP44 path
2. Shielded: User can import/generate mnemonic, and derive a with a default (or custom) Zip32 path from a private key derived from `m/44'/877'/0'/0'/2147483647'` (This is a fixed path for the private key from which all shielded accounts will derived for any given seed/phrase)

We currently only have a custom BIP44 form (as before, we would derive shielded keys using _default_ zip32 path of `m/32'/877'/0'`), but with this change, it would mean we should implement a ZIP32 form so users can derive _any_ shielded account from the seed (by specifying zip32). They cannot override the BIP44 path of `m/44'/877'/0'/0'/2147483647'` which is used as the basis for all shielded accounts (as we need to match CLI where this value is fixed)

### UX ramifications in Keychain & Namadillo

That said, this also breaks our concept of "Active Account". Since we shouldn't group them anymore, an "Active Account" _could_ be a transparent _or_ shielded account, which also makes no sense. 

In my mind, a "Wallet" is the mnemonic you imported, which is used to seed any number of accounts, transparent or shielded, and having an _Active_ wallet makes sense - user could use any addresses associated with that wallet, whereas currently "Active" account means a transparent and shielded account _pair_. But, of course, this would require a reworking of the import process, but here's the thought:
 
- If user imports a seed phrase (let's call it the wallet), this is tagged as the Parent
- If user wants to derive additional accounts from that wallet, they wouldn't import their seed phrase again, they'd choose an existing wallet and derive a new account from that
  - Which could be shielded or transparent

^^ This was the original design of the Extension, but we hijacked the `parentId` field of child accounts in order to associate Shielded Accounts with their respective Transparent _parents_, in order to group them together in the UI. To reintroduce something like this, we would have to leave current accounts in a stranded state and unassociated with any mnemonic. Technically users could still choose to import the same seed phrase and simply choose another path & alias, that's not a problem. But, I think it would be simpler and more convenient to give the option of deriving from an existing seed. The Keychain storage already supports that (as it was designed this way in the beginning).

Alternatively, we could keep this like the current flow, where users always provide a seed phrase (or private key), and we maintain a sort of flat hierarchy of accounts, but that also means that "Active Account" doesn't really do us much good. Choosing a transparent _or_ shielded account as active (as they're not not grouped together) seems clunky and meaningless. Showing all transparent & shielded accounts in the listing is easy to do, but we should distinguish between them as there will be 2x the amount of accounts in the list if they're ungrouped. They're already stored this way, so that part is trivial.

jurevans:
Resolves #1611 

- [x] Modified zip32 is handled with a constant BIP44 path of `m/44'/877'/0'/0'/2147483647'`
- [x] Update SDK tests for shielded accounts
- [x] Add field to storage (shielded accounts only) indicating modified zip32 path
  - This should be `modifiedZip32Path: "m/44'/877'/0'/0'/2147483647'"`
  - This should _only_ apply to shielded accounts derived from a _mnemonic_! Imported private keys (seed for zip32) are already derived, therefore, we can't make any assumptions as to _how_ it was derived, so leave this field out!)
- [x] Accounts (mnemonic-based, shielded-only) _without_ this field should display a warning to re-import
- [x] Add Zip32 form in `Advanced` (will attach designs soon)
- [x] Handle "Account Already Imported" better
  - Say you specify custom paths (BIP44 and ZIP32), if the BIP44 account hasn't already been used, but the ZIP32 has, you get an error on import
  - When this happens, the Transparent account _exists_ and is set as the active account, but, you can't see it in the overview :( This needs to be fixed! 
- [x] Only return shielded keys to Namadillo if they are _**not**_ outdated!
- [x] Also, if you import accounts but get an error on the Shielded keys because that address is already used, you will still see the transparent account in the extension (was hidden/filtered out before) - This relates to a bug reported by a user and is fixed here. Transparent Namada keys _without_ shielded keys should display in both Extension & Namadillo.

### Testing

New functionality:

- Set some accounts up in an extension based on `main` (running extension locally so its store is preserved when switching branches)
- Switch to this branch, reload extension in `Manage Extensions` - When launching the account overview, you should see a warning to update your accounts
- Re-import one of the accounts, and that account should no longer have the warning!
- Also, can test custom derivation paths
  - You _must_ specify a custom path for Zip32 _if_ you have already imported the default path (should see an error that the account already exists)

The most critical thing to confirm is:

- The same mnemonic & path should produce the same keys in:
  - Keychain extension (this branch)
  - CLI wallet (must be on `namada`  version `v1.1.1`+!)
    - Check viewing keys, as CLI payment addresses are diversified - Keychain currently uses `default_address` off of the viewing key (will introduce _additional_ payment addresses after this release!)
      ```bash
      # Derive some shielded keys
      namadaw derive --shielded --hd-path "m/32'/877'/0'" --alias "shielded /0"
      namadaw derive --shielded --hd-path "m/32'/877'/1'" --alias "shielded /1"

      # This will display viewing key: Ensure this matches the same mnemonic/zip32 path in Extension!
      namadaw list
      ```
  - Ledger HW Wallet (version `3.0.0`+)
    - Verify viewing key
    - **NOTE** This is trickier to test, but I can check against Ledger in https://github.com/anoma/namada-interface/pull/1575

### Notes

New shielded accounts generated from a _mnemonic_ (not private key import) now look like this in storage:
```json
{
    "id": "0459542.....",
    "address": "znam1...",
    "alias": "Tester",
    "parentId": "485057eb-...",
    "path": {
        "account": 0
    },
    "type": "shielded-keys",
    "modifiedZip32Path": "m/44'/877'/0'/0'/2147483647'",
    "owner": "zvknam1....",
    "pseudoExtendedKey": "039cd768...",
    "source": "imported",
    "timestamp": 0
}
```
This indicates:
1. That is was derived using modified-zip32
2. How exactly the _seed_ for the zip32 derivation was derived (this path is a constant)
3. The path components of the zip32 derivation itself (account = 0 - user can now change this)

So, with this we can check whether or not the user's shielded keys are supported.


apps/extension/src/background/keyring/keyring.ts
```diff
@@ -450,16 +432,16 @@ export class KeyRing {
...
+      path,
       parentId,
       type,
       alias,
       info,
       source,
       timestamp
     );
-    return derivedAccount;
+    return derivedShieldedAccount;
```
jurevans, line 444: Just a note about this change: So we had a `deriveAccount` method, which would alternatively call `deriveTransparentAccount` or `deriveShieldedAccount`, but it was never used for transparent, so I consolidated it into one function, `deriveShieldedAccount`

jurevans: If/when we allow deriving only Transparent, we can add a specific call for that at that time

euharrison (approved):
Nice! I builded the extension locally and can observe the new shielded address and the new viewing key, but transparent address was preserved

mateuszjasiuk:
Tested and LGTM! :) So if we have old account the MASP tab looks like this:

obraz

I assume this is fine, right?

mateuszjasiuk (approved):

jurevans:
> I assume this is fine, right?

@mateuszjasiuk I think this is fine, things like sync won't work because this no longer returns shielded accounts (if they're outdated). But I also do think we need to communicate that this will happen if user has outdated accounts. Otherwise we'd need to make more changes to Namadillo

PR #1625, opened 2025-02-11
fix: fixing ibc assets transfer fees

### LINKED ISSUE #1581: Fix transfer error when paying with a different token without balance
My balance is 0.01 OSMO on masp
I tried to pay a transfer that was 0.06 OSMO

The button is hanging on "Submitting..." while we got this sdk error below

Image

### LINKED ISSUE #1621: Add scrollbar on fee modal
We need a scrollbar when selecting a token on the transaction fee modal

Image

pedrorezende:
Making more explicit when we're using display amounts and base denomination amounts. We were having some errors because of that, like in #1581 , also for selecting and displaying fees.

Fixes #1581 
Fixes #1621

### DIFF
--- a/apps/namadillo/src/App/Common/GasFeeModal.tsx
+++ b/apps/namadillo/src/App/Common/GasFeeModal.tsx
@@ -172,7 +172,7 @@ export const GasFeeModal = ({
           containerProps={{
             className: twMerge(
               "text-sm w-full flex-1 border border-white rounded-sm",
-              "px-4 py-[9px] "
+              "px-4 py-[9px]"
             ),
           }}
           arrowContainerProps={{ className: "right-4" }}
--- a/apps/namadillo/src/App/Masp/MaspShield.tsx
+++ b/apps/namadillo/src/App/Masp/MaspShield.tsx
@@ -78,6 +78,7 @@ export const MaspShield: React.FC = () => {
       setCurrentStatus("");
       setCurrentStatusExplanation("");
     },
+    asset: selectedAsset?.asset,
   });
 
   const onChangeSelectedAsset = (address?: Address): void => {
--- a/apps/namadillo/src/App/Masp/MaspUnshield.tsx
+++ b/apps/namadillo/src/App/Masp/MaspUnshield.tsx
@@ -76,6 +76,7 @@ export const MaspUnshield: React.FC = () => {
       setCurrentStatus("");
       setCurrentStatusExplanation("");
     },
+    asset: selectedAsset?.asset,
   });
 
   const onChangeSelectedAsset = (address?: Address): void => {
--- a/apps/namadillo/src/App/NamadaTransfer/NamadaTransfer.tsx
+++ b/apps/namadillo/src/App/NamadaTransfer/NamadaTransfer.tsx
@@ -109,6 +109,7 @@ export const NamadaTransfer: React.FC = () => {
     onSuccess: () => {
       setCompletedAt(new Date());
     },
+    asset: selectedAsset?.asset,
   });
 
   const isSourceShielded = isShieldedAddress(source);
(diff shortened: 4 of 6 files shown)


apps/namadillo/src/App/Transfer/TransferModule.tsx
```diff
@@ -144,27 +146,27 @@ export const TransferModule = ({
...
     ) {
       return availableAmount;
     }
 
-    const amountMinusFees = availableAmount.minus(
-      displayGasFee.totalDisplayAmount
-    );
+    const amountMinusFees = availableAmount
+      .minus(displayGasFee.totalDisplayAmount)
+      .decimalPlaces(6);
```
euharrison, line 168: shouldn't we apply this only on render components to be safer?

apps/namadillo/src/atoms/fees/atoms.ts
```diff
@@ -83,8 +83,10 @@ export const gasPriceTableAtom = atomWithQuery<GasPriceTable>((get) => {
         const baseAmount = BigNumber(minDenomAmount);
         return {
           token,
-          // TODO should we run `toDisplayAmount` for all tokens or only NAM?
-          gasPrice: asset ? toDisplayAmount(asset, baseAmount) : baseAmount,
+          gasPrice:
+            asset && isNamadaAsset(asset) ?
```
euharrison, line 87: so is this the answer of the issue!? 🫠

apps/namadillo/src/atoms/integrations/services.ts
```diff
@@ -338,16 +338,20 @@ export const fetchIbcChannelFromRegistry = async (
...
-  return estimatedGas * (1 + additionalPercentage);
+  try {
+    const estimatedGas = await stargateClient.simulate(
+      sourceAddress!,
+      [transferMsg],
+      undefined
+    );
+    return estimatedGas * (1 + additionalPercentage);
+  } catch (error) {
+    throw error;
```
euharrison, line 355: I believe this is not necessary anymore

apps/namadillo/src/types.ts
```diff
@@ -40,7 +40,7 @@ export type AddressBalance = Record<Address, BigNumber>;
 
 export type GasConfig = {
   gasLimit: GasLimit;
-  gasPrice: GasPrice;
+  gasPriceInMinDenom: GasPrice;
```
euharrison, line 43: so, could we review this just in case? I can feel your frustration, and I share this as well. What I believe on this specifically case is that it could be the gasPrice on minDenom (for all other tokens), but for Namada this is on displayDenom, or at least this is what I understood on `gasPriceTableAtom`. If this is true, having the `InMinDenom` could lead to more confusion later. So do you think it's reasonable for this case preserve the term `gasPrice` only?

apps/namadillo/src/atoms/integrations/services.ts
pedrorezende: I would not risk :P I'll leave that for now until we have everything related to gas estimate 100% fixed

apps/namadillo/src/types.ts
pedrorezende: minDenomAmount on Namada has the same value than the displayAmount :/ this is the biggest cause of confusion. So for Namada it never makes any difference, we'll never pass unam anywhere - I think we only use unam on the CLI to transfer to another chain

apps/namadillo/src/atoms/fees/atoms.ts
pedrorezende: yep :/ we're receiving unam from the indexer, and we use nam everywhere, even on the cli, differently of other tokens

apps/namadillo/src/App/Transfer/TransferModule.tsx
pedrorezende: I think we're fine, since we won't have more than 6 decimal places. In the worst case we'll have a difference of less than 0.000001 for a given amount

PR #1628, opened 2025-02-11
fix: rewards fee race condition

euharrison:
When loading the interface directly on http://localhost:5173/staking/claim-rewards or if you is too fast clicking on the 'Claim' button, the native token asset is not found and the gas is not used as displayAmount, so there is an error

The solution was to add `chainAssetsMap` on the `queryKey` plus changing the type of `chainTokensQuery` to `Record<Address, Asset | undefined>` with the undefined piece, so we can map similar cases better

https://github.com/user-attachments/assets/79f82092-f212-4a82-9419-b964e7c2f100


mateuszjasiuk (approved):
LGTM!

PR #1636, opened 2025-02-12
fix: fixing gas and transactions for 1.7.0

pedrorezende:


apps/namadillo/src/App/Masp/MaspShield.tsx
```diff
@@ -165,8 +167,8 @@ export const MaspShield: React.FC = () => {
           isShielded: true,
         }}
         feeProps={feeProps}
-        isSubmitting={isPerformingTransfer}
-        errorMessage={generalErrorMessage}
+        isSubmitting={isPerformingTransfer || isSuccess}
```
euharrison, line 170: I'm just reading this semantically, not checking the implementation, but looks strange `isSubmitting === isSuccess`

Should it be renamed to `disabled={isPerformingTransfer || isSuccess}` ?

apps/namadillo/src/utils/index.ts
```diff
@@ -85,12 +85,18 @@ export const isNamadaAsset = (asset: Asset): boolean =>
 
 export const toDisplayAmount = (
   asset: Asset,
-  baseAmount: BigNumber
+  baseAmount: BigNumber,
+  skipNamada = false
```
euharrison, line 89: This smells, are we using this somewhere else? If yes, should we add this condition on this place instead of here?

Looks like we are passing a `displayAmount` to this fn, like `toDisplayAmount(displayAmount)` instead of `toDisplayAmount(baseAmount)`

apps/namadillo/src/App/Masp/MaspShield.tsx
pedrorezende: Maybe we need to rephrase this, but we need to keep the submitting animation running while we have the tx watcher verifying if the transaction was applied on chain. The execution on useTransfer only handles the broadcasting, so we keep it submitting while we make this verification. I'll try to improve this

euharrison: hmm it gives me one idea, if we could add the pooling of the wrapper/tx to useTransfer, then we'll have the applied status used for all transactions, not only transfers

euharrison (approved):

apps/namadillo/src/App/Masp/MaspShield.tsx
pedrorezende: I don't think it's a good idea cause this way we won't be able to track transactions when users leave the page :/

Issue #1631, opened 2025-02-12
Unshield button needs to not offer IBC until we enable direct IBC unshielding


ChrisHoltDesign:
As we can't IBC unshield direct from the shield pool currently. 

The unshield button in the top NAV should currently only open the internal unshield feature.

Users should not be able to select IBC here as they will be 'withdrawing' and not unshielding from transparent accounts

<img width="793" alt="Image" src="https://github.com/user-attachments/assets/00e2b29f-0b80-4529-a3f7-3947b22af8e3" />

<img width="632" alt="Image" src="https://github.com/user-attachments/assets/161d52d2-7215-45db-af87-2bf43dc05f72" />

euharrison:
This is a temp solution. We should roll it back after implementing https://github.com/anoma/namada-interface/issues/1638

Issue #1494, opened 2025-01-10
Add the Ibc/Namada transfer selector modal to more buttons


euharrison:
We need to add the Ibc/Namada transfer selector button to more places

![image](https://github.com/user-attachments/assets/77897415-c4bd-47b3-94c3-60a8e1bc929d)

![Screenshot 2025-01-10 at 14 47 51](https://github.com/user-attachments/assets/6cf2cb6a-e72d-4ffa-9540-f34ed46d4d57)

PR #1641, opened 2025-02-13
feat: update shield/unshield navigation flow

euharrison:
Update masp transfer flow 

- Add the missing modals for Shield/Unshield like we have on TopNavigation
- Temporally removes the IBC Unshield from Shielded Pool while it's not implemented
- Also, add the full banner of extension connect if it's not yet, so we don't show a eternal loading state which confuses the user 

closes https://github.com/anoma/namada-interface/issues/1494
closes https://github.com/anoma/namada-interface/issues/1631

https://github.com/user-attachments/assets/b7fff28d-e664-44af-9581-8dcf3193a194


pedrorezende (approved):
Very nice! Can we just update the title, so we have a better changelog entry?

Issue #1568, opened 2025-01-17
Namadillo: Ranking of assets in the selection dropdown


ChrisHoltDesign:
<img width="526" alt="Image" src="https://github.com/user-attachments/assets/c9c565ae-a028-4e15-b8bd-34a942fbdd99" />

Would be good to move the assets that will be incentivised to the top of the list e.g Osmo in this case is ranked last and out of sight

euharrison:
we could sort by this criteria https://github.com/satoshilabs/slips/blob/master/slip-0044.md

brentstone:
@euharrison is this solved by #1658?

euharrison:
@brentstone I don't think so. We are raking them by the chain registry order, and it could be only coincidence that it's resolved. But when we expand for more tokens, this order could not be sufficient

brentstone:
Ok cool just wanted to check, thx

neocybereth:
I've got a PR for this [here](https://github.com/anoma/namada-interface/pull/1687) but I'll need somebody with a ton of random tokens to test. All I have is ATOM.

PR #1658, opened 2025-02-17
feat: sorting assets on transfer module

pedrorezende:
This PR sorts the available assets from the user wallet, so the fee tokens appear first according to cosmos chain-registry order.


euharrison:
Do you think it worth connecting this implementation with this issue that sort the available assets?
https://github.com/anoma/namada-interface/issues/1568

Maybe this list could be a guide for our ranking https://github.com/satoshilabs/slips/blob/master/slip-0044.md

euharrison (approved):

pedrorezende:
Hmm, I think we're not ranking them by being incentivized, so I'm not sure this will fix the issue mentioned by Chris. I'll leave that issue open for now, maybe we'll have a better criteria in the future for this sort

PR #1679, opened 2025-02-19
feat: blocking masp usage if extension is not compatible with modified-zip32

pedrorezende:


jurevans (approved):
LGTM! I tested this with the store version (0.3.7) and works great! I wonder if we should put the minimum version in that message, `0.4.0`? (though I think this is perfectly fine as is!)

apps/namadillo/src/hooks/useKeychainVersion.tsx
```diff
@@ -0,0 +1,19 @@
+import { useEffect, useState } from "react";
+import { useNamadaKeychain } from "./useNamadaKeychain";
+
+export const useKeychainVersion = (): string | undefined => {
+  const keychain = useNamadaKeychain();
+  const [version, setVersion] = useState<string>();
+
+  useEffect(() => {
+    (async () => {
+      const namadaKeychain = await keychain.namadaKeychain.get();
```
euharrison, line 10: Thinking loud here. Literally the `keychain` here is only `new NamadaKeychain()`, and the whole hook idea is not necessary. It was the reason that created the other issue on useEffect dependencies. Maybe we should just the `new NamadaKeychain()` directly

euharrison (approved):

PR #1680, opened 2025-02-19
fix: revert keychain compatibility to 0.3.x

pedrorezende:


jurevans (approved):
I think this is good for current mainnet users to not see a warning, but I do see a warning now if I'm running 0.4.0. I'll leave it up to you which compromise we want to make, I am mostly worried that mainnet users on 0.3.7 will see the warning, but not have an extension to update to yet (while it's in review)

PR #1684, opened 2025-02-20
fix: remove trailing period on urls

euharrison:
Remove trailing period on urls

This fixes the url on proposal like this one https://interface.namada.tududes.com/governance/proposal/8


mateuszjasiuk (approved):
Nice! Works for me.

PR #1686, opened 2025-02-20
fix: show validators outside consensus and update tooltip icon

### LINKED ISSUE #1618: Validators outside of consensus are not displayed in Namadillo staking list unless searched for
- Validators with less that 1000 NAM should be included in the 'Inactive dropdown selection' with users being able to stake to them. 
- Currently the only way to find user outside of consensus is by directly searching there name

Image

neocybereth:
Screenshot 2025-02-20 at 4 48 47 PM

Closes #1618


apps/namadillo/src/App/Staking/AmountField.tsx
```diff
@@ -32,7 +30,7 @@ export const AmountField = ({
 
   return (
     <div className="relative">
-      {(isActive || forceActive) && (
+      {(isActive || forceActive || isInactive) && (
```
neocybereth, line 33: this now allows inactive validators to show an amount field

apps/namadillo/src/App/Staking/AmountField.tsx
```diff
@@ -62,18 +60,6 @@ export const AmountField = ({
         {isJailed && hasStakedAmounts && (
           <p>Validator is jailed. You have stopped receiving rewards</p>
         )}
-
-        {isInactive && !hasStakedAmounts && <p>Validator is inactive</p>}
-
-        {isInactive && hasStakedAmounts && (
```
neocybereth, line 68: Just want to double check that both of these statements should be removed now that we're allowing stakes to "inactive" validators with a stake of <1,000 as mentioned here:

https://github.com/anoma/namada-interface/issues/1618#issue-2842548540

apps/namadillo/src/App/Staking/IncrementBonding.tsx
```diff
@@ -104,6 +105,10 @@ const IncrementBonding = (): JSX.Element => {
     seed: seed.current,
   });
 
+  const topValidatorsByRank = useMemo(() => {
```
neocybereth, line 108: I moved this up one level b/c previously it was using `filteredValidators` so all of the sudden in the `inactive` view those validators were "topValidatorsByRank"

apps/namadillo/src/App/Staking/AmountField.tsx
euharrison: @ChrisHoltDesign I checked the docs https://docs.namada.net/operators/validators/validator-sets#jailed-and-inactive-validators and I think we should not allow to stake on inactive validators

> Inactive validators have manually removed themselves from consideration in consensus activities with a deactivate transaction. This can typically signal that a validator has ceased operations and might signal delegators to unbond from it. See more details [here](https://docs.namada.net/operators/validators/validator-actions#deactivation-and-reactivation).


Do you have a opinion here?

ChrisHoltDesign: Inactive validators should not receive stake no!

However a running validator who is below consensus <1000 NAM we should be able to stake to.

These validators should be displayed as active but have a warning symbol that shows they are out of consensus

ChrisHoltDesign: So an amendment to the first issue, initial i said these validators appear in the 'innactive list'

Maybe we need a new filter option that is 'Active below threshold' or something

euharrison: I like this `Active below threshold` idea, it's less confusing, at least for me :)

euharrison:
Hello @neocybereth, I'm back here with more information

On the indexer api endpoint https://housefire-api.sproutstake.space/api/v1/pos/validator/all, we have the `state` of the validator, like `"state": "belowThreshold"`.

Screenshot 2025-02-24 at 09 26 12

Here you can see all possible options for this value https://github.com/anoma/namada-indexer/blob/main/swagger.yml#L995-L1006

Talking with other devs, we should add the tooltip `This validator is outside of the consensus set. You will not receive staking rewards until they are active` to all validators that are not with the `"state": "consensus"`. Which means that we should not introduce the logic of `<1000 nams` on Namadillo. It'll be handled by the indexer based on the chain parameters

We should preserve other tooltips on top this, like `Validator is jailed. You have stopped receiving rewards` and any other that exists

Please, if you can compile all these informations and update the PR, it'll be great!

neocybereth:
Awesome thank you for the info @euharrison! I've updated the PR accordingly. 

So basically: 

1) Reinstate the Validator is Jailed messages etc ✅
2) Add a tooltip for any item where the `state` variable is !== `consensus` that says: `This validator is outside of the consensus set. You will not receive staking rewards until they are active`

euharrison (approved):
Nice!

euharrison:
@neocybereth FYI, created a new issue for the filter https://github.com/anoma/namada-interface/issues/1759

PR #1687, opened 2025-02-20
fix: update token sorting logic

### LINKED ISSUE #1568: Namadillo: Ranking of assets in the selection dropdown

neocybereth:
Closes #1568


neocybereth:
Just FYI I'm waiting for this to be tested.

See comment here: https://github.com/anoma/namada-interface/issues/1568#issuecomment-2670518188

euharrison:
Hi @neocybereth, sorry, I suggest the slip-44 but actually, re-thinking, I believe we should check the tokens on our chain

For example, in case someone has ETH, on the wallet, it'll be on top of ATOM because it has a higher rank on slip-44, but currently we even don't support ETH on the chain

So, we could check the token indexer endpoint (ex: https://namada-housefire-idx.emberstake.xyz/api/v1/chain/token) and pull the tokens from it to rank them higher and incentivize them. Whe have this fetch already in the code if you search for `chainTokensAtom`

What do you think?

apps/namadillo/src/utils/registeredCoinTypes.ts
```diff
@@ -0,0 +1,6074 @@
+// https://github.com/satoshilabs/slips/blob/master/slip-0044.md
+
+export const registeredCoinTypes = [
```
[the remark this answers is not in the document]
neocybereth: Ah epic, think I could use that as well instead.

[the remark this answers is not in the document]
neocybereth: Updated to use that package.

neocybereth:
sounds good @euharrison I updated the code to reflect that. Again, I don't have enough random tokens to test this well so I'll need someone else to test *OR* will need some tokens sent to my addy's.

apps/namadillo/src/App/Transfer/TransferModule.tsx
```diff
@@ -306,17 +297,27 @@ export const TransferModule = ({
...
-        const asset1Index = findTokenFeeIndex(source.chain, asset1.asset);
-        const asset2Index = findTokenFeeIndex(source.chain, asset2.asset);
-
-        // No fee assets, so we sort them by the amounts owned by the user
-        if (asset1Index === -1 && asset2Index === -1)
-          return asset1.amount.gt(asset2.amount) ? -1 : 1;
-
-        if (asset1Index === -1) return 1;
-        if (asset2Index === -1) return -1;
+        const bip44Index1 = allUsersAssets.findIndex(
```
euharrison, line 300: do you think we should rename this variable name now `bip44Index2`?

apps/namadillo/src/App/Transfer/TransferModule.tsx
```diff
@@ -306,17 +297,27 @@ export const TransferModule = ({
...
+        const bip44Index2 = allUsersAssets.findIndex(
+          (asset) =>
+            asset?.symbol.toLowerCase() === asset2.asset?.symbol.toLowerCase()
+        );
+
+        // If either asset is in the list, sort based on its position.
+        if (bip44Index1 !== -1 || bip44Index2 !== -1) {
+          // If both assets are in the list, sort by their order in registeredCoinTypes.
+          if (bip44Index1 !== -1 && bip44Index2 !== -1) {
+            return bip44Index1 - bip44Index2;
```
euharrison, line 313: if both assets are in the list, we should sort by the amount, because this token list is not necessarily the best for user

euharrison: hmm maybe we should just revert back the original implementation and only replace the `findTokenFeeIndex` by the new `findIndex`, wdyt?

euharrison:
@neocybereth sorry, I think we talked somewhere about hiding assets that is not accepted by the chain currently. So, maybe, instead of the current implementation, we should do a `filter(chainAssets)` + `sort(asset.amount)`, wdyt?

neocybereth:
Alright @euharrison I updated the code so that we:

1) Filter out assets accepted by the chain
2) Sort assets by amount

Below image is what it shows which is *much* cleaner than before

Screenshot 2025-02-25 at 12 50 18 PM

euharrison (approved):

PR #1699, opened 2025-02-21
Check Ledger app version before enabling zip32

### LINKED ISSUE #1698: Extension - Custom Zip32 paths for Ledger
The new Zip32 custom path form needs to be incorporated for Ledger HW Wallet. Also, we need to check Ledger version and only import viewing & auth keys for Ledger NamadaApp `v3.0.1+`

- Hook zip32 form up for Ledger
- Only enable zip32 features for Ledger if NamadaApp is at least version `3.0.1`

jurevans:
Resolves #1698 

- [x] If zip32 is supported on this Ledger app version, allow specifying custom path
- [x] If zip32 supported, include shielded import steps
- [x] If zip32 isn't supported, display warning and skip shielded import steps
- [x] Store correct zip32 path in Ledger account storage
  - Custom paths should be correct in `View Keys` - this is the only view where paths are seen, and only when they are non-default!

### TODO
- [ ] Improve custom path form and approval steps UI

### TESTING

- Use 3.0.0+ version of Ledger (latest is `v3.0.1`) - flows should be identical to before with shielded accounts in place
  - Custom zip32 paths are now correct (UI to be improved after this PR)
- Downgrade to version in Ledger Live (1.0.5) and go through import flows: You'll see a message that shielded accounts will be enabled in 3.0.0, and it will only import transparent
  - Non-shielded Tx work as before

### DIFF
--- a/apps/extension/src/Setup/Common/LedgerApprovalStep.tsx
+++ b/apps/extension/src/Setup/Common/LedgerApprovalStep.tsx
@@ -2,10 +2,12 @@ import { Heading, ProgressIndicator, Stack } from "@namada/components";
 
 type LedgerApprovalStepProps = {
   currentApprovalStep: number;
+  isZip32Supported: boolean;
 };
 
 export const LedgerApprovalStep = ({
   currentApprovalStep,
+  isZip32Supported,
 }: LedgerApprovalStepProps): JSX.Element => {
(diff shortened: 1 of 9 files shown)


mateuszjasiuk (approved):
LGTM!

PR #1701, opened 2025-02-21
fix: general fixes on IBC Withdrawal, notifications and TransferModule

pedrorezende:


apps/namadillo/src/hooks/useTransaction.tsx
```diff
@@ -33,12 +32,11 @@ type PartialNotification = Pick<ToastNotification, "title" | "description">;
 export type UseTransactionPropsEvents<T> = {
   onSigned?: (tx: TransactionPair<T>) => void;
   onError?: (err: unknown) => void;
-  onSuccess?: (tx: TransactionPair<T>) => void;
```
pedrorezende, line 36: Removed to improve clarity. Success will only happen when the transaction is actually applied, not after being broadcasted to the chain.

apps/namadillo/src/hooks/useTransaction.tsx
```diff
@@ -120,80 +113,78 @@ export const useTransaction = <T,>({
     mutationFn: async (
       additionalParams: Partial<BuildTxAtomParams<T>> = {}
     ) => {
-      invariant(
```
pedrorezende, line 123: Nothing really changed here, I just wrapped with a try catch. The relevant part is related to error throwing below.

apps/namadillo/src/utils/index.ts
```diff
@@ -46,9 +46,17 @@ export const useTransactionEventListener = <T extends keyof WindowEventMap>(
 
   useEffect(() => {
```
pedrorezende, line 47: Added support to a list of event names instead of a single name

apps/namadillo/src/hooks/useTransaction.tsx
euharrison: great idea! I was thinking about it on the custom events as well, renaming them to `Broadcasted` and `Applied`

PR #1702, opened 2025-02-21
feat: filtering assets with zero amount in it

pedrorezende:
With recent changes on indexer, a few zeroed assets are being returned from the balances endpoint. This PR filters those out.

### DIFF
--- a/apps/namadillo/src/App/Transfer/TransferModule.tsx
+++ b/apps/namadillo/src/App/Transfer/TransferModule.tsx
@@ -17,6 +17,7 @@ import {
   GasConfig,
   WalletProvider,
 } from "types";
+import { filterAvailableAsssetsWithBalance } from "utils/assets";
 import { checkKeychainCompatibleWithMasp } from "utils/compatibility";
 import { getDisplayGasFee } from "utils/gas";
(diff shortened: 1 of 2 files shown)


apps/namadillo/src/App/Transfer/TransferModule.tsx
```diff
@@ -144,6 +144,19 @@ export const TransferModule = ({
...
+      (previous, current): AddressWithAssetAndAmountMap => {
+        if (source.availableAssets![current].amount.gt(0)) {
+          return { ...previous, [current]: source.availableAssets![current] };
+        }
+        return previous;
+      },
+      {}
+    );
+  }, [source.availableAssets]);
+
```
jurevans, line 159: This works great for me! Are we also going to figure the assets with zero balance displayed in the Overview?

jurevans (approved):
LGTM!

apps/namadillo/src/App/Transfer/TransferModule.tsx
pedrorezende: Hmmm I'll extract this filter into a new function and open a new PR for that! Thanks for remembering that

PR #1718, opened 2025-02-23
chore: update viewing key language

### LINKED ISSUE #1714: Extension: specific language for what a viewing key is
Current test is the screenshot. Should be clear that "access" is only read-access. Maybe say "the ability to see" instead of "access too".

Image

neocybereth:
Closes #1714

### DIFF
--- a/apps/extension/src/App/Accounts/ViewingKey.tsx
+++ b/apps/extension/src/App/Accounts/ViewingKey.tsx
@@ -14,8 +14,9 @@ export const ViewingKey = (): JSX.Element => {
       <Stack full gap={GapPatterns.TitleContent}>
         <PageHeader title="Viewing Key" />
         <p className="text-white">
-          Your viewing key grants the holder access to all your balances and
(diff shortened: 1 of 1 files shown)


apps/extension/src/App/Accounts/ViewingKey.tsx
```diff
@@ -14,8 +14,9 @@ export const ViewingKey = (): JSX.Element => {
       <Stack full gap={GapPatterns.TitleContent}>
         <PageHeader title="Viewing Key" />
         <p className="text-white">
-          Your viewing key grants the holder access to all your balances and
-          transaction data. Please keep it secure to protect your data.
+          Your viewing key grants the holder the ability to see to all your
```
euharrison, line 17: can you replace `to see to all` by `to see all` (removing the second `to`) ?

ref: https://github.com/anoma/namada-interface/issues/1714#issuecomment-2677952091

euharrison (changes requested):

euharrison (approved):

PR #1716, opened 2025-02-23
WIP: fix: update mismatched data hashes

neocybereth:
Screenshot 2025-02-23 at 9 36 54 PM

```
namadac query-proposal --proposal-id 10 --node $RPC 
 
Last committed epoch: 324
Proposal Id: 10
Type: Default with Wasm
Author: tnam1qqgll8x8rz9fvtdv8q6n985za0vjvgyu0udxh7fp
Content: {"details": "In anticipation of updating the claim-rewards tx wasm in the Namada on-chain storage, we first must increase the maximum allowed gas in a block. This must be done because the tx to initialize that proposal consumes about 4.3M gas units, whereas the current block maximum is 3M. This proposal increases the block maximum from 3M to 5M.\n\nThe code used to produce the attached WASM can be viewed here: https://github.com/anoma/namada-governance-upgrades/blob/b26aff300654a791b91d8ff40483d7d2f0bf9d5f/max_block_gas/src/lib.rs", "discussions-to": "https://forum.namada.net/t/two-proposals-to-update-the-claim-rewards-wasm-tx-on-chain/1548", "title": "Increase max block gas to 5M gas units"}
Start Epoch: 318
End Epoch: 330
Activation Epoch: 331
Status: on-going
Data: Hash: DB27ADA6B1F734008E4A3AF440F0666A441FDDAF2D8FF44CBC9B6D187949ECD1
```

Closes #1713

Issue #1713, opened 2025-02-22
Namadillo: incorrect wasm data hashes


brentstone:
Displayed `Data hash` and the `sha256sum` of the wasm that can be downloaded from Namadillo (while consistent with each other) are incorrect relative to what is actually on-chain and can be queried from the CLI. 

Consider:
```
namadac query-proposal --proposal-id 10 --node $RPC 
 
Last committed epoch: 324
Proposal Id: 10
Type: Default with Wasm
Author: tnam1qqgll8x8rz9fvtdv8q6n985za0vjvgyu0udxh7fp
Content: {"details": "In anticipation of updating the claim-rewards tx wasm in the Namada on-chain storage, we first must increase the maximum allowed gas in a block. This must be done because the tx to initialize that proposal consumes about 4.3M gas units, whereas the current block maximum is 3M. This proposal increases the block maximum from 3M to 5M.\n\nThe code used to produce the attached WASM can be viewed here: https://github.com/anoma/namada-governance-upgrades/blob/b26aff300654a791b91d8ff40483d7d2f0bf9d5f/max_block_gas/src/lib.rs", "discussions-to": "https://forum.namada.net/t/two-proposals-to-update-the-claim-rewards-wasm-tx-on-chain/1548", "title": "Increase max block gas to 5M gas units"}
Start Epoch: 318
End Epoch: 330
Activation Epoch: 331
Status: on-going
Data: Hash: DB27ADA6B1F734008E4A3AF440F0666A441FDDAF2D8FF44CBC9B6D187949ECD1
```

but compare to a Namadillo like https://namadillo.emberstake.xyz/governance/proposal/10, which currently has `274AF385120A25485E68A2A22B58FD9EE8DE339B691E8536FBBD84849209981D`.

neocybereth:
<img width="617" alt="Image" src="https://github.com/user-attachments/assets/5fe45bc4-5739-4555-baef-ece83cc93456" />

Seems this line of rust could be the culprit...we're never returning `default_with_wasm` just `default`

neocybereth:
<img width="628" alt="Image" src="https://github.com/user-attachments/assets/8e31fe95-f8e0-46c3-b4e1-0a2062654edf" />

Alright it goes deeper....this is the result I'm getting back from the indexer. Looks like the indexer is returning the wrong numbers for `yayVotes` , `abstainVotes`, `nayVotes`. @Fraccaman

brentstone:
[...]

PR #1721, opened 2025-02-24
fix: proposals checksums and data download

mateuszjasiuk:
Indexer is already returning a hash and we have to use separate endpoint to get the data.


euharrison (approved):

jurevans (approved):
LGTM!

brentstone:
@mateuszjasiuk did this close #1713? Ok to close #1716 in that case?

PR #1731, opened 2025-02-24
fix: update transparent balance after transfer

euharrison:
Update transparent balance after transfer

We were updating one atom balance, but not the other. This PR fixes this unifying both balance atoms


pedrorezende (approved):
LGTM!

PR #1732, opened 2025-02-24
fix: fix claim rewards fee button

euharrison:
Fix claim rewards fee button adding one controller for each transaction

This layout could update in the future to provide a better UX, but this is a quick win for the current claim issue\

Screenshot 2025-02-24 at 15 11 36


pedrorezende (approved):
Good one!

Issue #1734, opened 2025-02-24
Namadillo: provide configurable tx fee on `Withdraw` button


brentstone:
Right now, if someone hits a `Withdraw` button on the `Staking` tab to withdraw some tokens, they aren't shown what the tx fees are estimated to be or given the option to configure their gas limits.

brentstone:
There might also be other gas issues related to the Withdraw tx

euharrison:
It'll be partially resolved by https://github.com/anoma/namada-interface/pull/1739 

The ideal fix is to update the indexer to return the ibc estimates, instead of 502

PR #1739, opened 2025-02-24
fix: add fallback for missing gas estimates

euharrison:
Add fallback for missing gas estimates

To test, you can simulate to stake for 15 validator at once with the network tab open, you will see the indexer returning 502, but the fee still increasing

closes https://github.com/anoma/namada-interface/issues/1734

https://github.com/user-attachments/assets/519c364c-af93-454d-b594-c4bc6c4747d1

### DIFF
--- a/apps/namadillo/src/atoms/fees/services.ts
+++ b/apps/namadillo/src/atoms/fees/services.ts
@@ -6,35 +6,23 @@ import {
 } from "@namada/indexer-client";
 import { TxKind } from "types/txKind";
 
-const convertLegacyGasKind = (kind: GasLimitTableInnerTxKindEnum): TxKind => {
-  switch (kind) {
-    case "revealPk":
-      return "RevealPk";
-    case "bond":
-      return "Bond";
-    case "unbond":
-      return "Unbond";
-    case "redelegation":
-      return "Redelegate";
-    case "claimRewards":
-      return "ClaimRewards";
-    case "voteProposal":
-      return "VoteProposal";
-    case "transparentTransfer":
-      return "TransparentTransfer";
-    case "shieldingTransfer":
-      return "ShieldingTransfer";
-    case "shieldedTransfer":
-      return "ShieldedTransfer";
-    case "unshieldingTransfer":
-      return "UnshieldingTransfer";
-    case "ibcMsgTransfer":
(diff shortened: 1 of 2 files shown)


apps/namadillo/src/atoms/fees/services.ts
```diff
@@ -1,14 +1,95 @@
...
+  // if fails, try to fetch from the legacy endpoint
+  try {
+    const legacyValues = (await api.apiV1GasGet()).data;
+    const sum = legacyValues.reduce((sum, item) => {
+      const txKind = convertLegacyGasKind(item.txKind);
+      return sum + item.gasLimit * (counter(txKind) ?? 0);
+    }, 0);
+    return {
+      min: sum,
+      avg: sum,
```
pedrorezende, line 78: Can we add some low and high percentages increments here, so we don't repeat values. Maybe 10%?

apps/namadillo/src/atoms/fees/services.ts
```diff
@@ -1,14 +1,95 @@
 import {
   DefaultApi,
   GasEstimate,
+  GasLimitTableInnerTxKindEnum,
   GasPriceTableInner,
 } from "@namada/indexer-client";
+import { TxKind } from "types/txKind";
+
+const convertLegacyGasKind = (kind: GasLimitTableInnerTxKindEnum): TxKind => {
+  switch (kind) {
```
pedrorezende, line 10: Nitpick: maybe we could use a map

pedrorezende (approved):

apps/namadillo/src/atoms/fees/services.ts
```diff
@@ -1,14 +1,95 @@
...
+    counter("Withdraw"),
+    counter("RevealPk"),
+    counter("Redelegate"),
+  ];
+
+  // fetch from the estimate endpoint
+  try {
+    return (await api.apiV1GasEstimateGet(...params)).data;
+  } catch (e) {
+    console.error(e);
```
pedrorezende, line 66: Can we place a more descriptive error here? I think it will be important to register when the gas estimate fails. Maybe a message, like `console.error("gas estimate failed", e)` (a better message than that of course)

pedrorezende (approved):

PR #1746, opened 2025-02-25
fix: register ledger device

mateuszjasiuk:
There is one annoying thing that after going back from settings we have to choose asset once again, but we can improve it later.

https://github.com/user-attachments/assets/e2304830-3757-4bfc-847e-12d77c416660

### DIFF
--- a/apps/namadillo/src/App/Settings/SettingsLedger.tsx
+++ b/apps/namadillo/src/App/Settings/SettingsLedger.tsx
@@ -1,0 +1,23 @@
+import { ActionButton, Stack } from "@namada/components";
+
+export const SettingsLedger = (): JSX.Element => {
+  const onInvalidateShieldedContext = async (): Promise<void> => {
+    await navigator.usb.requestDevice({
+      filters: [{ vendorId: 0x2c97 }],
+    });
+  };
+
+  return (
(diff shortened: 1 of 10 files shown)


apps/namadillo/src/App/Settings/SettingsLedger.tsx
```diff
@@ -0,0 +1,28 @@
...
+      <h2 className="text-base">Ledger</h2>
+      <p className="text-sm">
+        To use a Ledger device with Namadillo, you must register it first. This
+        only needs to be done once. If you ever need to unregister a device, go
+        to your browser settings and remove it from the list of registered
+        devices.
+      </p>
+      <p className="text-sm">
+        The <b>Register Ledger Device</b> button remains active because we have
+        no way to check if a device is already registered.
```
jurevans, line 20: I would maybe reword this:

> The <b>Register Ledger Device</b> button remains active because we have
> no way to check if a device is already registered.

It is normal to see the <b>Register Ledger Device</b> button after registering due to limitations of Ledger registration detection in browser. 

...

Or something. Hmmm, I dunno, it just feels a little weird, but is likely fine! I get why it should be here

jurevans (approved):
This looks really good! Take or leave my comments, they are very minor suggestions!

PR #1754, opened 2025-02-25
WIP: feat: add executed proposal status to FE

neocybereth:

### DIFF
--- a/apps/namadillo/src/App/Governance/AllProposalsTable.tsx
+++ b/apps/namadillo/src/App/Governance/AllProposalsTable.tsx
@@ -122,6 +122,7 @@ const statusFilters = [
   "ongoing",
   "passed",
   "rejected",
+  "executed",
 ] as const;
 type StatusFilter = (typeof statusFilters)[number];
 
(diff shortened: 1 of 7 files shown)


brentstone:
@neocybereth mind getting this to build? Failing in CI

neocybereth:
This one's failing b/c it's missing the type from the indexer client so I'll need to wait for that to get the pipe passing

brentstone:
You shouldn't need a new `Executed` type from the indexer. On Namadillo, `Executed` can be considered the `Passed` type AND current_epoch >= proposal activation epoch.

neocybereth:
> You shouldn't need a new `Executed` type from the indexer. On Namadillo, `Executed` can be considered the `Passed` type AND current_epoch >= proposal activation epoch.

You're 100% correct that it's just a matter of changing the "Passed" status flag
I can, and have done that, at the data layer. BUT the way the component code is written it checks for types and so prevents a caveat in this case. 

That's a good thing IMO b/c it keeps the codebase clean and uniform, b/c if we start adding even tiny caveats it can become muddied real quick

PR #1771, opened 2025-02-26
fix: left align retro card amounts

### LINKED ISSUE #1766: Left-align the Retro PGF amounts in a Gov Proposal
I'd prefer to see the amounts left-aligned rather than center-aligned
Image

neocybereth:
Closes #1766


pedrorezende (approved):
Hmm it would be better to render this part as a table, but this helps a lot already without many changes! Let's go! o/

PR #1779, opened 2025-02-27
feat: changing indexer compatibility message to redirect users to Namada Apps

pedrorezende:


apps/namadillo/src/utils/compatibility.tsx
```diff
@@ -48,9 +46,11 @@ export const checkIndexerCompatibilityErrors = (
   if (checkResult === CompatibilityOutput.IncompatibleVersion) {
     return (
       <>
-        You&apos;re using an outdated version of Namada Indexer. Please update
-        your indexer URL in the{" "}
-        <Link to={routes.settingsAdvanced}>Advanced Settings</Link> section.
+        This Namadillo instance is running outdated infra. Please try another
```
euharrison, line 49: @pedrorezende I think @ChrisHoltDesign suggested to replace all variants of the messages for this one

Maybe we can use these variants with a `console.warn` as they are already coded

euharrison (approved):

PR #1783, opened 2025-02-27
feat: limit ibc withdrawal of assets only for the same chain

### LINKED ISSUE #1747: Prevent users from 'withdrawing' from namada to non native asset source chains
Currently the UI is allowing users to send assets to other chains over IBC and not source.

The Namadillo UI shows these are successful transactions with no errors, However they are then not easily viewabe in keplr as they are namada wrapped tokens

See grab below of a confirmation of TIA to COSMOS HUB 

Image

----------

We need to prevent the users from making these transfers in the first place.

- The destination source should not be pre-filled when there is no asset selected as it currently shows. (See below)

Image

-----------------------------------------

- Once a user selects the asset they want to withdraw the UI should then pre-select the destination source based on the asset. The only source chains in the drop down should be the native chain to the asset + Stride

in the case of OSMO the follwing should appear in the dropdown

Image

pedrorezende:
The target chain on Withdrawal should always be the original chain of the selected asset.

Fixes #1747

### DIFF
--- a/apps/namadillo/src/atoms/integrations/functions.ts
+++ b/apps/namadillo/src/atoms/integrations/functions.ts
@@ -440,3 +440,18 @@ export const addLocalnetToRegistry = (config: LocalnetToml): void => {
 export const getDenomFromIbcTrace = (ibcAddress: string): string => {
   return ibcAddress.replaceAll(/(transfer\/|channel-\d+\/)*/gi, "");
 };
+
+export const chainHasFeeTokenDenom = (chain: Chain, denom: string): boolean => {
+  return !!chain.fees?.fee_tokens?.some(
+    (feeToken) => feeToken.denom.toLowerCase() === denom.toLowerCase()
+  );
+};
+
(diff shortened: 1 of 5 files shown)


apps/namadillo/src/atoms/integrations/functions.ts
```diff
@@ -440,3 +440,18 @@ export const addLocalnetToRegistry = (config: LocalnetToml): void => {
 export const getDenomFromIbcTrace = (ibcAddress: string): string => {
   return ibcAddress.replaceAll(/(transfer\/|channel-\d+\/)*/gi, "");
 };
+
+export const chainHasFeeTokenDenom = (chain: Chain, denom: string): boolean => {
+  return !!chain.fees?.fee_tokens?.some(
+    (feeToken) => feeToken.denom.toLowerCase() === denom.toLowerCase()
```
euharrison, line 446: 👍

euharrison (approved):
nice!

PR #1784, opened 2025-02-27
feat: tweaking black color on all apps

### LINKED ISSUE #1764: There should be no solid black on namadillo #000000. Change to #0F0F0F
This will soften the contrast on the eye on the module elements and should be applied globally to anything that is #000000

pedrorezende:
Fixes #1764

### DIFF
--- a/packages/components/src/theme.ts
+++ b/packages/components/src/theme.ts
@@ -4,7 +4,7 @@ export const colors = {
   orange: "#FF8A00",
   fail: "#DD1539",
   white: "#fff",
-  black: "#0B0B0B",
+  black: "#0F0F0F",
   gray: "#292929",
   rblack: "#000000",
   upcoming: "#7DA0A8",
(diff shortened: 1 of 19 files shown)


packages/components/src/theme.ts
```diff
@@ -4,7 +4,7 @@ export const colors = {
   orange: "#FF8A00",
   fail: "#DD1539",
   white: "#fff",
-  black: "#0B0B0B",
+  black: "#0F0F0F",
   gray: "#292929",
   rblack: "#000000",
```
euharrison, line 9: should we delete this?

euharrison (approved):

packages/components/src/theme.ts
pedrorezende: Ops, definitely

pedrorezende: Hmm I would say to not delete it just in case we need to switch to a black black somewhere (like some text or something)

PR #1797, opened 2025-03-01
feat: adding masp asset rewards panel info

### LINKED ISSUE #1334: Add in the shielded reward rate table for phase 3 on the MASP and transfer modules
Screenshot 2024-11-27 at 19 35 24

pedrorezende:
This PR adds the related methods to the sdk and shared packages to load global masp asset rewards from the RPC, and also implements the sidebar panel enumerating the rewards.

Closes #1334

### DIFF
--- a/packages/sdk/src/rpc/types.ts
+++ b/packages/sdk/src/rpc/types.ts
@@ -44,6 +44,15 @@ export type StakingPositions = {
   unbonds: Unbonds[];
 };
 
+export type MaspTokenRewards = {
+  name: string;
+  address: string;
+  max_reward_rate: number;
+  kp_gain: number;
+  kd_gain: number;
+  locked_amount_target: number;
+};
+
 /**
  * DelegationTotals
  * Record<address, totalDelegations>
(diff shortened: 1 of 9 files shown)


mateuszjasiuk (approved):
Nice! LGTM! :) We can potentially move the query to the indexer later. Also I'm not sure if it's enough to display those rates in the MASP view. In the figma it shows up in multiple places,

packages/sdk/src/rpc/types.ts
```diff
@@ -44,6 +44,15 @@ export type StakingPositions = {
...
 
+export type MaspTokenRewards = {
+  name: string;
+  address: string;
+  max_reward_rate: number;
+  kp_gain: number;
+  kd_gain: number;
+  locked_amount_target: number;
+};
+
```
jurevans, line 55: only suggestion here would be to map these values to match TS conventions over Rust snake-case, but it's not a huge deal!

jurevans (approved):
LGTM! Tested and works!

mateuszjasiuk (approved):

PR #1809, opened 2025-03-03
feat: Keychain signing - only require password on session timeout

### LINKED ISSUE #1411: Namadillo / Keychain. We should implement a timer on password input.
This has been discussed previously.

It's very tedious having to retype your password into the keychain to perform every action. Particularly if you performing multiple actions in one session.

- Previously we had discussed a locker timer where by password re-input is required after 5 minutes after first input
- Alternatively would be a checkbox in the settings whereby users can disable having to keep re-inputting the password

jurevans:
Resolves #1411

**NOTE** The timeout period is set to 5 minutes.

- [x] Extension can sign _without_ entering password every time
- [x] Track session start time (times out after 5 minutes), able to query background to determine if session has expired
- [x] Require auth if session is timed-out, otherwise, it is not needed
  - [x] If session is valid, clicking `Approve` submits the signature
  - [x] If session is invalid, user must enter password (like before), then clicking `Authenticate` submits the signature
- [x] Implement for `ConfirmSignTx`
- [x] Implement for `ConfirmSignArbitrary`

### Testing

- With this Keychain, unlock and submit a transaction
  - You should not need to enter your password
- Check Chrome service worker (if testing on Chrome) - wait for it to go `Inactive`, then submit another Tx (~1 minute, make sure you don't have service worker dev tools open! This causes the service worker to stay running)
  - If it's been under 5 minutes since you last entered your password, you should not be prompted for it when signing
  - This test is just to demonstrate that an inactive Chrome service worker preserves the existing session and shouldn't have any effect on this feature
- After 5 minutes since you last entered your password, submit a Tx (you can change this line to like one minute if it's too annoying to wait): https://github.com/anoma/namada-interface/blob/7a30b9a173ee4ab40b1433331d78fbd7fcefb349/apps/extension/src/background/vault/service.ts#L81
  - At this point, you need to enter password.
  - Any Tx within the next 5 minutes will not need a password

### DIFF
--- a/apps/extension/src/Approvals/Approvals.tsx
+++ b/apps/extension/src/Approvals/Approvals.tsx
@@ -19,12 +19,6 @@ import { ConfirmSignLedgerTx } from "./ConfirmSignLedgerTx";
 import { ConfirmSignTx } from "./ConfirmSignTx";
 import { WithAuth } from "./WithAuth";
 
-export enum Status {
-  Completed,
-  Pending,
-  Failed,
-}
-
 export type ExtensionLockContextType = {
   isUnlocked?: boolean;
   setIsUnlocked?: Dispatch<SetStateAction<boolean>>;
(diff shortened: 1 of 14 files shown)


mateuszjasiuk (approved):
Tested and LGTM! Seems weird not having to type in password every time :D

PR #1820, opened 2025-03-05
fix: convert asset type to the latest one before calculating shielded…

mateuszjasiuk:
… rewards

Also bumps namada and deps


jurevans (approved):
Looks good!

PR #1830, opened 2025-03-06
feat: use indexer services timestamps in status check

mateuszjasiuk:
With this change the "status dot" will turn red if one of indexers services - chain, pos, parameters or transactions lags behind.
This skips checks for rewards and governance as they do not update timestamps fast enough(computation can take a while). I plan to change this but first we need to update the indexer.

I think the best way to test it is to use different indexers and see if dot is turning red :D 
image


jurevans (approved):
Nice! This works great!

PR #1836, opened 2025-03-06
feat: bumping minimum compatibility for the indexer

pedrorezende:
Namadillo should be compatible with the indexer version above 2.3.0


brentstone (approved):

pedrorezende:
Good observation! Fixed that! Thanks

Issue #1834, opened 2025-03-06
Namadillo: Better inform users that fees can be paid with other assets other than NAM


ChrisHoltDesign:
Alter the Fee UI in the transfer modules

- Separate the Fee token into a separate entity that has a dropdown arrow.
- The actual fee on the right will then update based on the users choice
- This will open up the same modal we have at current

<img width="635" alt="Image" src="https://github.com/user-attachments/assets/1eb22f01-fd70-41df-bbcd-e96437bfb51c" />

ChrisHoltDesign:
Open to copy suggestion here for the tooltip @brentstone @zjcio

ChrisHoltDesign:
Alt option

<img width="635" alt="Image" src="https://github.com/user-attachments/assets/2006732d-8ec5-4f62-aa17-5e435c3c441b" />

ChrisHoltDesign:
I wonder if we should change the 'Change fee token' to 'Fee options' as it's more than just the token you can alter

ChrisHoltDesign:
<img width="865" alt="Image" src="https://github.com/user-attachments/assets/99d2a437-c440-4970-b0ff-10264174cade" />

PR #1843, opened 2025-03-07
feat: update transaction fee button

euharrison:
Update transaction fee button to be more evident that user can change the token and the gas limit

Closes https://github.com/anoma/namada-interface/issues/1834

Screenshot 2025-03-07 at 12 38 41

Screenshot 2025-03-07 at 12 38 51

Screenshot 2025-03-07 at 12 38 56


mateuszjasiuk (approved):
Tested! Nice!

PR #1852, opened 2025-03-07
feat: falling back to other fee tokens if user has no nams

### LINKED ISSUE #1787: Namadillo: Fee Token detection for MASP transactions
Currently in Namadillo, MASP txs like `unshield` and `transfer` require one to pay fees from a shielded address. However, the `Transaction fee` button defaults to NAM right now, which is problematic in Phase 3 since nobody has shielded NAM yet!

It would be great if Namadillo could detect what tokens the user does have shielded and then default to one of those (doesn't really matter, as long as they have the token and enough of it) for fee payment.

pedrorezende:
If the user doesn't have NAM tokens, we set other tokens as default token to pay for fees
Fixes #1787

### DIFF
--- a/apps/namadillo/src/hooks/useTransactionFee.ts
+++ b/apps/namadillo/src/hooks/useTransactionFee.ts
@@ -8,6 +8,7 @@ import {
   gasPriceTableAtom,
   isPublicKeyRevealedAtom,
 } from "atoms/fees";
+import { tokenPricesFamily } from "atoms/prices/atoms";
 import BigNumber from "bignumber.js";
 import { useAtomValue } from "jotai";
 import { useMemo, useState } from "react";
@@ -44,6 +45,33 @@ export const useTransactionFee = (
   const { data: gasPriceTable, isLoading: isLoadingGasPriceTable } =
     useAtomValue(gasPriceTableAtom);
 
+  const gasDollarMap =
+    useAtomValue(
+      tokenPricesFamily(gasPriceTable?.map((item) => item.token) ?? [])
+    ).data ?? {};
+
+  const findCheapestToken = (
+    gasPriceTable: GasPriceTable
+  ): string | undefined => {
+    let minPriceInDollars: BigNumber | undefined,
+      cheapestToken: string | undefined;
+
+    for (const gasItem of gasPriceTable) {
+      const price = gasDollarMap[gasItem.token];
+      if (!price) return;
+      const gasPriceInDollars = price.multipliedBy(gasItem.gasPrice);
+      if (
+        typeof minPriceInDollars === "undefined" ||
+        gasPriceInDollars.lt(minPriceInDollars)
+      ) {
+        minPriceInDollars = gasPriceInDollars;
+        cheapestToken = gasItem.token;
+      }
+    }
+
+    return cheapestToken;
+  };
+
   const availableGasTokenAddress = useMemo(() => {
     if (!gasPriceTable) return nativeToken;
 
@@ -70,7 +98,7 @@ export const useTransactionFee = (
       (balance) => balance.tokenAddress === nativeToken
     );
 
-    if (new BigNumber(nativeAddressBalance?.minDenomAmount || "0").gt(0)) {
+    if (BigNumber(nativeAddressBalance?.minDenomAmount || "0").gt(0)) {
       return nativeAddressBalance?.tokenAddress;
     }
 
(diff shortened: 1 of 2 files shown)


apps/namadillo/src/hooks/useTransactionFee.ts
```diff
@@ -37,9 +44,52 @@ export const useTransactionFee = (txKinds: TxKind[]): TransactionFeeProps => {
...
+    }
+
+    if (!isShielded && !userTransparentBalances.data) {
+      return nativeToken;
+    }
+
+    // Separate shielded amount from transparent
+    const balances =
+      (isShielded ?
+        // TODO: we need to refactor userShieldedBalances to return Balance[] type instead
```
euharrison, line 61: Definitely! I think about this almost daily 🙏

apps/namadillo/src/hooks/useTransactionFee.ts
```diff
@@ -37,9 +44,52 @@ export const useTransactionFee = (txKinds: TxKind[]): TransactionFeeProps => {
...
+      (balance) => balance.tokenAddress === nativeToken
+    );
+
+    if (new BigNumber(nativeAddressBalance?.minDenomAmount || "0").gt(0)) {
+      return nativeAddressBalance?.tokenAddress;
+    }
+
+    // Fallback to another token containing balance
+    const gas = gasPriceTable.filter((gas) => {
+      return !!balances.find(
```
euharrison, line 79: How about if we invert this?
- We can sort the balance by amount, meaning we can choose the balance with higher funds (or other criteria)
- It's almost 100% correct to say that, if you have a token in your balance, it should exists on the gasPriceTable. But the opposite is not true, we could have many tokens on gasPriceTable that we have zero balance

apps/namadillo/src/hooks/useTransactionFee.ts
```diff
@@ -37,9 +44,52 @@ export const useTransactionFee = (txKinds: TxKind[]): TransactionFeeProps => {
...
+        (balances) =>
+          balances.tokenAddress === gas.token &&
+          new BigNumber(balances.minDenomAmount).gt(0)
+      );
+    });
+
+    // If no token found, fallback to NAM
+    return gas.length > 0 ? gas[0].token : nativeToken;
+  }, [userTransparentBalances, gasPriceTable, isShielded]);
```
euharrison, line 88: Should we add `userShieldedBalances.data` as a dependency?

euharrison (approved):
Nice!

apps/namadillo/src/hooks/useTransactionFee.ts
mateuszjasiuk: Piggybacking here, not exactly related, but I was curious if it would make sense to change the fee token to the token that we are transferring(if not enough nam). But that would probably mean that we would have to display fee token buton after we picked token to transfer :P

apps/namadillo/src/hooks/useTransactionFee.ts
```diff
@@ -37,9 +44,52 @@ export const useTransactionFee = (txKinds: TxKind[]): TransactionFeeProps => {
...
+          tokenAddress: balance.address,
+        }))
+      : userTransparentBalances.data) || [];
+
+    // Check if user has enough NAM to pay fees
+    const nativeAddressBalance = balances.find(
+      (balance) => balance.tokenAddress === nativeToken
+    );
+
+    if (new BigNumber(nativeAddressBalance?.minDenomAmount || "0").gt(0)) {
```
mateuszjasiuk, line 73: btw "new" is optional, it's a bit shorter that way :D

mateuszjasiuk (approved):
Tested and it works nicely! Left couple of comments, not that important :)

apps/namadillo/src/hooks/useTransactionFee.ts
pedrorezende: > It's almost 100% correct to say that, if you have a token in your balance, it should exists on the gasPriceTable.
I don't think this is true. After we enable new coins, the number of possible tokens a user could have in their account will be much higher than the allowed assets to pay the gas. I think I'll keep like it is just to avoid any future problem

> We can sort the balance by amount
Definitely! I'll do that. We need to convert them to dollars first, tho. Maybe we should pick the cheaper token for fees instead of the one users have a higher amount, wdyt?

> it would make sense to change the fee token to the token that we are transferring
I think it would make more sense to select the best token to pay the fees, if users don't have any nam. Sometimes the token they're transferring is not enough to pay the fees, or has a much more expensive cost for the fees

mateuszjasiuk (approved):
LGTM!

Issue #1845, opened 2025-03-07
Invalidate shielded context after masp indexer is fixed


mateuszjasiuk:
### Status
🟢 Accepted

### Context
We need to automatically invalidate shielded context for future(working version) of masp indexer.
MASP Indexer version: **TBD**

### Decision
We need to add a feature to invalidate shielded context automatically.

### Consequences
Users will be bale to unshield without any additional work(manually invalidating).

PR #1853, opened 2025-03-07
feat: invalidate shielded context based on indexer version

euharrison:
Save the indexer version (per chain) when invalidating the shielded context. 

Then uses this saved version to compare with the minimum accepted indexer version:
- in case the saved version is below the minimum accepted
- and the indexer version is greater then the minimum accepted
- execute a new context invalidation saving the new indexer version

Closes: https://github.com/anoma/namada-interface/issues/1845

https://github.com/user-attachments/assets/ad9404d2-b032-4a57-8ce8-420d23cce06d

### DIFF
--- a/apps/namadillo/src/hooks/useShouldInvalidateShieldedContext.ts
+++ b/apps/namadillo/src/hooks/useShouldInvalidateShieldedContext.ts
@@ -1,0 +1,37 @@
+import { chainParametersAtom } from "atoms/chain/atoms";
+import {
+  indexerHeartbeatAtom,
+  lastInvalidateShieldedContextAtom,
+} from "atoms/settings/atoms";
+import { useAtomValue } from "jotai";
+import { useEffect } from "react";
+import semverSatisfies from "semver/functions/satisfies";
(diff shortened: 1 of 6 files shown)


apps/namadillo/src/hooks/useShouldInvalidateShieldedContext.ts
```diff
@@ -0,0 +1,37 @@
...
+  indexerHeartbeatAtom,
+  lastInvalidateShieldedContextAtom,
+} from "atoms/settings/atoms";
+import { useAtomValue } from "jotai";
+import { useEffect } from "react";
+import semverSatisfies from "semver/functions/satisfies";
+import { useInvalidateShieldedContext } from "./useInvalidateShieldedContext";
+
+// change this value to the minimun indexer version that demands to clear the context
+const minIndexerVersion = ">=9.9.0";
```
euharrison, line 12: This version is a higher number on purpose, to avoid invalidating the current context from the current user without providing a real user value.

When we have the final indexer version, with masp updated, we only need to update this number

mateuszjasiuk (approved):
Tested! Works nicely! :)

PR #1859, opened 2025-03-08
feat: improving token selection modal

### LINKED ISSUE #1821: Namadillo: Show balances in fee modal
The fee selection modal in Namadillo 1.15.1 only shows the gas fee in different assets, but not the balances I have of those assets. I need to go back to the overview to check if which assets can cover the fee. 

Image

I suggest to add a column that shows my current balances. 

Maybe add a checkbox to hide assets that do not have enough balance to cover the fee.

pedrorezende:
Resolves #1821


euharrison (approved):
Cool!

Issue #1589, opened 2025-01-29
Allow users to copy both shielded and transparent address from the top Nav


ChrisHoltDesign:
Currently users can only copy there transparent address from the namadillo top Nav.

Create a dropdown from the account name / copy button that displays both the transaparent and shielded address.

A stretch goal for this would also allow users to randomise the shielded address to create sub accounts in the same UI

<img width="1093" alt="Image" src="https://github.com/user-attachments/assets/1046760b-2a2e-4425-ac4e-de6a5f523956" />

brentstone:
Oh I like the idea of Namadillo generating a new random shielded address for the user every time (if they'd like, that kind of default behavior might spook the most novice users though)

Issue #1554, opened 2025-01-15
Hide the switch account button if has only one account


euharrison:
Hide the switch account button if has only one account

![Screenshot 2025-01-15 at 17 26 28](https://github.com/user-attachments/assets/b3005bc6-4586-48bd-adcd-c605b15a0771)
![Screenshot 2025-01-15 at 17 26 31](https://github.com/user-attachments/assets/6a64f380-163b-4640-b231-3b54825aa59a)

PR #1866, opened 2025-03-08
feat: add button to copy shielded address

euharrison:
On hover the account name, shows the account addresses to be copied

Also, hide the switch button in case the user has only one account (accounts with the same alias)

It also support shielded only accounts imported with spending key

Closes https://github.com/anoma/namada-interface/issues/1589
Closes https://github.com/anoma/namada-interface/issues/1554

Screenshot 2025-03-08 at 18 27 18

Screenshot 2025-03-08 at 18 27 30


mateuszjasiuk (approved):
Nice!

Issue #1697, opened 2025-02-21
Multiple assets with same name and different trace are shown on the Select Asset dropdown


neocybereth:
<img width="518" alt="Image" src="https://github.com/user-attachments/assets/cba779af-7b20-40d1-bc11-1712d46a970b" />

I'm seeing this in the work on #1687. 

From this log you can see that it's the same asset just coming from different traces:

<img width="864" alt="Image" src="https://github.com/user-attachments/assets/3a5903df-2f7a-4a76-8df7-df9e91373cdd" />

Sidenote, this trace seems like it's a bug with double `transfer/channel-7`
```
'transfer/channel-7/transfer/channel-7/uosmo'}
```

neocybereth:
@brentstone @euharrison any particular way you'd like me to go about solving this? Do I merge the similar traces or is this expected etc?

brentstone:
So hopefully we don't ever run into this on mainnet, it's happening bc on Housefure some older channels expired and we had to open new ones.

@pedrorezende @euharrison def curious what is going on with the `transfer/channel-7/transfer/channel-7/uosmo`.

Ideally in my mind, we would only use an alias for a single instance of a given token type and then use the `transfer/channel-X/yyyy` string otherwise. So in this example, `transfer/channel-7/uosmo` would appear as `Osmosis` while `transfer/channel-0/uosmo` would appear as `transfer/channel-0/uosmo`. Would this be done using the chain registry? How does Namadillo get `Osmosis` for the token anyway?

euharrison:
another ref https://github.com/anoma/namada-interface/issues/1672

euharrison:
Agree, we should render the `tnam...` for `transfer/channel-0/uosmo` as the easiest path, or render `transfer/channel-0/uosmo` if we create this map on the code

Related to the `transfer/channel-7/transfer/channel-7/uosmo`, @Fraccaman @grarco it came from the indexer, do you have any idea why it's happening?

euharrison:
@neocybereth I opened the PR https://github.com/anoma/namada-interface/pull/1868 feel free to comment on it if you had other ideas for this

PR #1868, opened 2025-03-10
feat: add warning for tokens from inactive channel

euharrison:
Add warning for tokens from inactive channel

Creates the `<TokenLabel>` and `<InactiveChannelWarning>` component that handle this inactive warning

Also, update the Asset Selector to render symbol instead of the asset name

To test, if you don't have an account with this issue, you can simulate your balances by hacking the fetch balance. On Housefire:
```tsx
// apps/namadillo/src/atoms/accounts/services.ts
const r = (await api.apiV1AccountAddressGet(account.address)).data;
const otherOsmo = r.find(
  (i) => i.tokenAddress === "tnam1phks0geerggjk96ezhxclt6r5tdgu3usa5zteyyc"
);
if (otherOsmo) {
  otherOsmo.minDenomAmount = "2000";
} 
return r;

// apps/namadillo/src/atoms/balance/atoms.ts
shieldedBalance.push({
  address: "tnam1phks0geerggjk96ezhxclt6r5tdgu3usa5zteyyc",
  minDenomAmount: "3000",
});
```

Closes https://github.com/anoma/namada-interface/issues/1697
Closes https://github.com/anoma/namada-interface/issues/1869

Screenshot 2025-03-10 at 11 01 15

Screenshot 2025-03-10 at 11 01 24

Screenshot 2025-03-10 at 11 01 31

Screenshot 2025-03-10 at 11 01 37


apps/namadillo/src/App/Common/TokenLabel.tsx
```diff
@@ -0,0 +1,30 @@
+import { Asset } from "@chain-registry/types";
+import { InactiveChannelWarning } from "App/Common/InactiveChannelWarning";
+import { AssetImage } from "App/Transfer/AssetImage";
+import { ReactNode } from "react";
+import { Address } from "types";
+
+export const TokenLabel = ({
```
pedrorezende, line 7: Can we rename this to TokenCard? Because it's not a label

apps/namadillo/src/App/Common/InactiveChannelWarning.tsx
```diff
@@ -0,0 +1,57 @@
...
+    !ibcChannels?.namadaChannel ||
+    trace.includes(ibcChannels.namadaChannel)
+  ) {
+    return <></>;
+  }
+
+  return <div className="text-xs text-neutral-500">inactive: {trace}</div>;
+};
+
+export const InactiveChannelWarning = ({
```
pedrorezende, line 30: Maybe we can extract the logic of this function and create a new one to return if a given token is active or not. Might be useful in other places

pedrorezende (approved):
LGTM! With a few minor comments

PR #1874, opened 2025-03-10
feat: creating a new version for Claim Rewards modal

### LINKED ISSUE #1819: Namadillo Proposal: Allow more granular claim of staking rewards
Currently users only have the choice to claim all staking rewards from all validators at one time.

In line with what keplr offers some users are asking to be able to claim rewards from certain validators but not others.

- Users would still be offered to 'claim all' in the UI but we can show a table below of each individual earnings that can be claimed individually.

Image

- This would be a 2 stage claim, based on the users choice, opening up a second modal to confirm the claim + allow for gas fee calculations based on the users choice

Image

pedrorezende:
Resolves #1819


packages/components/src/AmountSummaryCard.tsx
```diff
@@ -26,6 +27,7 @@ export const AmountSummaryCard = ({
   callToAction,
   className = "",
   isLoading = false,
+  isSuccess = false,
```
euharrison, line 30: This is breaking the number on the card `Available NAM to Stake`, we should add the `isSuccess` there

Screenshot 2025-03-10 at 18 18 28

PR #1876, opened 2025-03-10
feat: adding pro-mode control and hiding advanced settings panel

### LINKED ISSUE #1875: Namadillo: 'Pro mode' toggle (Open to discussion)
There have been open discussions that the Namadillo UX is currently straddling the line between dev focused UX and trying to be accessible to end users. It's not a bad place to be, but we are making concessions on each end of our audience base. 

Currently we are trying to replicate CLI operations 1:1 in Namadillo which is adding levels of complexity to new users and potentially scaring them away as the UX seems to complex. Given the current landscape of web3 and 'abstraction of tech' being the key narrative to onboarding a broader audience into blockchain, we should make Namadillo as dumb proof and as simple as possible for users to gain privacy. And set the preceident that Namada makes privacy easy vs other protocols.

We do not want to alienate either user base so the proposed idea of a **'pro mode'** toggle in namadillo that can turn on some of the more advanced CLI UX that we do not necessarily want to offer to everyone.

Elements that would be hidden until **'Pro Mode'** is turned on

- Advanced settings to custom Indexers and RPC URLS
- Times displayed in EPOCHS
- Testnet setting functionality
- View and connect individually derived keys
- Allows for multiple derived shielded payment addresses

How this talks to the keychain or vise verser should be discussed and or if the mode needs to be turned on in both instances.

This would work in conjunction with proposed UI plans for Namadillo V2  

Image

pedrorezende:
Resolves #1875

### DIFF
--- a/apps/namadillo/src/App/Settings/SettingsMain.tsx
+++ b/apps/namadillo/src/App/Settings/SettingsMain.tsx
@@ -1,6 +1,14 @@
+import { ToggleButton } from "@namada/components";
+import { IconTooltip } from "App/Common/IconTooltip";
 import { routes } from "App/routes";
-import { indexerHeartbeatAtom, maspIndexerHeartbeatAtom } from "atoms/settings";
+import {
+  indexerHeartbeatAtom,
+  maspIndexerHeartbeatAtom,
+  settingsAtom,
+  updateSettingsProps,
+} from "atoms/settings";
 import { useAtomValue } from "jotai";
+import { FaInfo } from "react-icons/fa6";
 import { GoLinkExternal } from "react-icons/go";
 import { version as sdkVersion } from "../../../../../packages/sdk/package.json";
 import { version } from "../../../package.json";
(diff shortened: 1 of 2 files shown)


apps/namadillo/src/App/Settings/SettingsMain.tsx
```diff
@@ -11,19 +19,65 @@ const { VITE_REVISION: revision = "" } = import.meta.env;
...
+                <>
+                  PRO mode unlocks more technically advanced features in
+                  Namadillo that replicates the UX of the CLI. You can learn
+                  more about these features{" "}
+                  <a
+                    href="https://docs.namada.net/users"
+                    target="_blank"
+                    rel="nofollow noreferrer"
+                  >
+                    here
```
euharrison, line 71: The word `here` is a little bit off because the parent has `display:flex`, maybe wrapping the content in a `div` can fix this 
Screenshot 2025-03-10 at 19 42 44

euharrison: we could also add a `text-yellow hover:underline` or similar to highlight the link

euharrison (approved):

PR #1882, opened 2025-03-11
fix: add z-index to prevent address hiding in IBC transfer

neocybereth:


neocybereth:
Screenshot 2025-03-11 at 1 05 42 PM

pedrorezende (approved):

PR #1889, opened 2025-03-11
ci: making stable branch up to date on every release

pedrorezende:
We now have a stable branch that must be up to date with the latest releases of namada-interface. It will help to automate a few things on CI


euharrison (approved):
so, in the end, we have the name `stable` instead of `release` 😁

PR #1891, opened 2025-03-11
fix: tokens sometimes not rendering correctly after refreshing masp page

pedrorezende:
To reproduce this issue you need to go to the masp list of tokens and refresh the page. Sometimes tokens are not loaded correctly.


euharrison (approved):
🙏

PR #1892, opened 2025-03-11
feat: improve sync indicator error log

euharrison:
Shows more descriptive errors when failed to sync

Screenshot 2025-03-11 at 18 31 31

Screenshot 2025-03-11 at 18 36 00

Screenshot 2025-03-11 at 18 36 24

### DIFF
--- a/apps/namadillo/src/App/Layout/SyncIndicator.tsx
+++ b/apps/namadillo/src/App/Layout/SyncIndicator.tsx
@@ -7,17 +7,43 @@ import {
 import { useAtomValue } from "jotai";
 import { twMerge } from "tailwind-merge";
 
+const formatError = (
+  errors: (string | Error)[],
+  label?: string
+): JSX.Element => {
+  if (!errors.length) {
+    return <></>;
+  }
+
+  return (
+    <div>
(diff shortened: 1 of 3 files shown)


apps/namadillo/src/App/Layout/SyncIndicator.tsx
```diff
@@ -7,17 +7,43 @@ import {
 import { useAtomValue } from "jotai";
 import { twMerge } from "tailwind-merge";
 
+const formatError = (
```
neocybereth, line 10: think we should bring this down into the component itself?

neocybereth (approved):

apps/namadillo/src/App/Layout/SyncIndicator.tsx
euharrison: oh, if we do it, it'll recreate the function every time, but we don't need it :)

neocybereth: TIL lol...I always just throw it in the component for cleanliness but to save bytes makes total sense to me! 👍

PR #1894, opened 2025-03-12
fix: fix notification toast pending

neocybereth:
Screenshot 2025-03-12 at 12 31 50 PM

### DIFF
--- a/apps/namadillo/src/atoms/notifications/functions.ts
+++ b/apps/namadillo/src/atoms/notifications/functions.ts
@@ -11,11 +11,9 @@ export const createNotificationId = (
   if (Array.isArray(data)) {
     return data
       .map((tx) => {
-        if ("innerTxHashes" in tx && Array.isArray(tx.innerTxHashes)) {
+        if (tx.hash) return tx.hash;
+        else if ("innerTxHashes" in tx && Array.isArray(tx.innerTxHashes))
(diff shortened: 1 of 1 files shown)


neocybereth:
@euharrison @pedrorezende here's the fix for the notification. 

```
  const onTransferSuccess = ({
    detail: tx,
  }: CustomEvent<TransferTransactionData>): void => {
    if (!tx.hash) return;
    const id = createNotificationId(tx.hash);
```

Screenshot 2025-03-12 at 1 12 08 PM

So in `onTransferSuccess` we're creating the notification ID with the hash b/c, as you can see in the photo, that's all we have access to, not the `innerTxHashes`.  With the pending transactions though we had access to `innerTxHashes` so the pending's notification ID was the `innerTxHash` and the successful's notification ID was the `hash` (thank you for the lead @euharrison)

neocybereth:
Screenshot 2025-03-12 at 1 06 06 PM

pedrorezende:
Re: the snippet you mentioned, I think we would need to use some time soon to refactor this, as we're having issues with tx being broadcasted, but not actually applied on chain. The same when a transaction times out and is actually applied on chain. Rethinking this logic will be necessary to fix these bugs that might mislead people in the usage of Namadillo

pedrorezende:
@neocybereth I'll go ahead and merge this one, ok? I think it will be important for the next release

PR #1896, opened 2025-03-12
feat: Only display unbond in the top overview when a user is unbonding #1743

### LINKED ISSUE #1743: Only display unbond in the top overview when a user is unbonding
I think the UI element that shows unbonding NAM can be removed unless a user has actioned an unbond.

Current view: Would be enabled when an unbond is active: 

Image

The following view would be the default:

Image

I would also flip so that shielded NAM is first

neocybereth:
Closes #1743

### DIFF
--- a/apps/namadillo/src/App/AccountOverview/NamBalanceContainer.tsx
+++ b/apps/namadillo/src/App/AccountOverview/NamBalanceContainer.tsx
@@ -108,7 +108,8 @@ export const NamBalanceContainer = (): JSX.Element => {
     shieldedAmountQuery,
     shieldedNamAmount,
   } = useBalances();
-
+  const isUnbondingOrWithdrawable =
+    unbondedAmount.gt(0) || withdrawableAmount.gt(0);
   return (
     <>
       {maspEnabled && (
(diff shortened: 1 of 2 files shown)


neocybereth:
Screenshot 2025-03-12 at 5 55 25 PM

neocybereth:
@ChrisHoltDesign

apps/namadillo/src/App/AccountOverview/NamBalanceContainer.tsx
```diff
@@ -102,66 +108,106 @@ export const NamBalanceContainer = (): JSX.Element => {
...
+                  <ListItemContainer
+                    isEnabled={shieldingRewardsEnabled}
+                    className="flex flex-1"
+                  >
+                    <ShieldedRewardsBox
+                      shieldedRewardsAmount={shieldedRewards.amount}
+                    />
+                  </ListItemContainer>
+                </>
+              : <>
```
euharrison, line 185: @neocybereth this looks like duplicated, is it only to add the `ListItemContainer` or is there another reason? If it's only the wrapper, we have some flexibility on the layout to preserve the best code we believe. Sometimes the layout screenshot will not be 100% pixel perfect. If this is case, can you remove this condition and apply the same layout for both views?

Screenshot 2025-03-12 at 11 22 59

euharrison (approved):
Approved, but if you could check and address the comments before merging :)

apps/namadillo/src/App/AccountOverview/NamBalanceContainer.tsx
```diff
@@ -102,66 +108,106 @@ export const NamBalanceContainer = (): JSX.Element => {
...
+              as="ul"
+              gap={2}
+              className={clsx("w-full", {
+                "order-1": maspEnabled && shieldingRewardsEnabled,
+              })}
             >
-              <ShieldedRewardsBox
-                shieldedRewardsAmount={shieldedRewards.amount}
+              <NamBalanceListItem
+                className={clsx({ "h-full": !isUnbondingOrWithdrawable })}
```
euharrison, line 134: I believe we don't need these class as a conditional, we could simply always apply it 

https://github.com/user-attachments/assets/ba57b49c-2402-4879-82d0-71b5f88d81ed

neocybereth: Nice one!

apps/namadillo/src/App/AccountOverview/NamBalanceContainer.tsx
neocybereth: Awesome good to know 👍 ...updated.

PR #1906, opened 2025-03-14
fix: pending notification not being removed on IBC Withdraw

### LINKED ISSUE #1903: IBC withdraw showing Pending notification indefinitely

neocybereth:
Closes #1903

### DIFF
--- a/apps/namadillo/src/atoms/notifications/functions.ts
+++ b/apps/namadillo/src/atoms/notifications/functions.ts
@@ -2,10 +2,10 @@ export const notificationIdSeparator = ";";
 import { TxProps } from "@namada/types";
 
 export const createNotificationId = (
-  data?: TxProps["hash"] | TxProps | TxProps["hash"][]
+  data?: TxProps["hash"] | TxProps["hash"][]
 ): string => {
   if (!data) return Date.now().toString();
   if (typeof data === "string") return data;
   if (Array.isArray(data)) return data.join(notificationIdSeparator);
-  return data.hash;
+  return data;
 };
--- a/apps/namadillo/src/atoms/notifications/functions.ts
+++ b/apps/namadillo/src/atoms/notifications/functions.ts
@@ -6,8 +6,6 @@ export const createNotificationId = (
 ): string => {
   if (!data) return Date.now().toString();
   if (typeof data === "string") return data;
-  if (Array.isArray(data)) {
-    return data.join(notificationIdSeparator);
-  }
+  if (Array.isArray(data)) return data.join(notificationIdSeparator);
   return data.hash;
 };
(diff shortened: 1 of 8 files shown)


neocybereth:
@euharrison @pedrorezende here's the fix for the notification. I'm not sure if this effects anything else downstream so just want to make sure you're aware of the change here.

neocybereth:
Screenshot 2025-04-01 at 1 04 42 PM


@euharrison @pedrorezende here's a temporary fix. Pretty ugly and just extends the existing hack we had, but I think this will have to do for now until we rewrite the `useTransaction` hook

apps/namadillo/src/atoms/notifications/functions.ts
```diff
@@ -1,13 +1,18 @@
...
 export const notificationIdSeparator = ";";
 
 export const createNotificationId = (
-  data?: TxProps | TxProps[] | string
+  data?: TxProps | TxProps[] | string | EncodedTxData<TxProps>
 ): string => {
   if (!data) return Date.now().toString();
   if (typeof data === "string") return data;
-
+  if ((data as EncodedTxData<TxProps>).type === "buildIbcTransfer") {
```
pedrorezende, line 11: Wouldn't be better to check first if "type" exists on data? I'm afraid this would throw an error trying to access `type` from an array

neocybereth: Good call, I updated that.

euharrison:
While I'm fixing this pending notification from  Claim & Stake https://github.com/anoma/namada-interface/issues/1907 I tested this PR and I'm receiving this error

Screenshot 2025-04-02 at 19 40 26

I think we should avoid as much as possible this pattern of `as Type` because it'll lead to these cases

I don't know yet what is the reason for this, but posting here in case this is useful somehow

euharrison:
In general, I think this `transactionId` as a string is creating a nightmare for us. Maybe it'll be better to handle this as an object, like `createNotificationId(...): { hash: string, innerTxHashes: string[] }` and we can have the freedom to decide how to clean the notifications based on this

neocybereth:
@euharrison @pedrorezende I've updated the `createNotificationId` function to only take in the hash array or hash string type. Now all individual places that call this can handle exactly what they pass in. For most of them it's the transaction hash but for IBC notifications (withdraws in particular) its the innerTxHashes b/c of the API

apps/namadillo/src/atoms/notifications/functions.ts
```diff
@@ -1,21 +1,11 @@
-import { TxProps } from "@namada/types";
-
 export const notificationIdSeparator = ";";
+import { TxProps } from "@namada/types";
 
 export const createNotificationId = (
-  data?: TxProps | TxProps[] | string
+  data?: TxProps["hash"] | TxProps["hash"][]
```
euharrison, line 5: Funny, I was thinking that you will allow passing a bunch of different types and handle them all here returning `string[]` instead of preserving the `data.join(";")`

euharrison (approved):
Cool! It's working fine on my tests ✅

Issue #1332, opened 2024-11-27
UI to allow users to 'unshield' direct from the shield pool put over IBC


ChrisHoltDesign:
From the UI side:
- in order to 'unshield' via IBC we just need to a 'shielded / transparent' source toggle to the 'From Namada Flow'

<img width="1030" alt="Image" src="https://github.com/user-attachments/assets/fd8ea5e8-26e5-4a4f-9971-39bda115c1cb" />

ChrisHoltDesign:
UI Changed needed to the IBC tab

- Add the toggle to allow to switch between shield and unshield. Shield would be the default 

<img width="1030" alt="Screenshot 2024-11-27 at 16 58 08" src="https://github.com/user-attachments/assets/8e50b8b9-710e-4b89-b362-75ae4d96b633">

ChrisHoltDesign:
We would then re-enable the unshield selection modal

<img width="597" alt="Image" src="https://github.com/user-attachments/assets/050aa8e0-49a4-4c52-90dc-39a8fb0f057d" />

mateuszjasiuk:
Design for refund keys:

![Image](https://github.com/user-attachments/assets/05215609-9016-4202-8085-5be6efa6744c)

PR #1920, opened 2025-03-18
feat: ibc unshielding

mateuszjasiuk:
Resolves https://github.com/anoma/namada-interface/issues/1332

- [x] refund target 
- [x] retest migrations

> The flow looks like this atm:
> - before we build the tx we generate random keypair and the address using OsRng. We store that in the extension local storage(this is semi-persistent, gets cleared after an hour)
> - after build and right before broadcasting we get the account info from local storage and we save it in the IndexedDB in the extension as a new account(same as any other account) but with a new type "disposable". Accounts with type "disposable" are filtered out by default(in the extension), but there is an option in settings to show them in the list of accounts.
> - if tx fails on namada side, we remove disposable account from the accounts list
> - if tx is successful on namada side, namadillo queries indexer to see if tx was also successfully applied on the destination chain.
>   - If it was, we remove disposable account from the accounts list
>   - If it wasn't, we just keep the account
> 
>As this account is stored as any other account it will display in namadillo. The one problem is that it does not have "shielded" counterpart, so we have to take this into consideration. Also when you take a look at tx details, refund_target is just a sender address. 

## Testing
**Use our internal infrastructure as it has the newest indexer**

### Happy path
- go to ibc->from namada->shielded and send some token(make sure gas fee is enough)
- check if balances changed correctly

### Sad path :(
Option 1:

1. go to `IbcWithdraw.tsx` and change `receiver`  in `performWidthdraw` call to some random string
2. go to ibc->from namada->shielded and send some token(make sure gas fee is enough)
3. wait for fail
4. once tx failed you should see disposable accounts toggle in the accounts list view in the extension
5. toggle **on** and see if there is a new disposable account(there is a 30 sec offset, so if tx failed before you might need to wait a bit )
6. select account, go to namadillo and see if it has balance - it should xd

Option 2:
1. go to `IbcWithdraw.tsx` and set `timeoutSecOffset` to `0`  in `performWidthdraw`  call
2. to 6. as above

### Testing accounts clear:
- Call `await this.namada.getSigner().clearDisposableKeypair("YOUR ADDR");` from namadillo to see if you can not remove NON DISPOSABLE address
- To check if someone does not try to persist address as disposable one, you can:
1. Call `await this.namada.getSigner().genDisposableKeypair();` from namadillo
2. In extension terminal: 
```js
let addr = "addr from genDisposableKeypair() output";
await chrome.storage.local.set({"Namada::LocalStorage/namadaExtensionDisposableSigners":
    {[addr]: {...(await chrome.storage.local.get("Namada::LocalStorage/namadaExtensionDisposableSigners"))["Namada::LocalStorage/namadaExtensionDisposableSigners"][addr], privateKey: "5f04046fd2d58e5ba3327131918980cdd4ff4a9afa46897ffc0d0c416759d509"}}});
```
3, In namadillo: `await this.namada.getSigner().persistDisposableKeypair("addr from genDisposableKeypair() output");`

You can also set gas limit to some lower value and see if this does not produce disposable account in extension


jurevans (approved):
LGTM!

jurevans (approved):
Nice!

PR #1921, opened 2025-03-18
fix: update password checker when second input is used first

### LINKED ISSUE #1916: [Namadillo]: Keychain - "Password are not matching" is not refreshed
### What happened?

"Password are not matching" is not refreshed when you add the passwords in the opposite order

Image

### Steps to Reproduce

1. create your account
2. add the second instance of the password text
3. add the 1st instance of the password text

### Relevant log output

```shell

```

### Debug Information

```shell

```

neocybereth:
Closes #1916

### DIFF
--- a/apps/extension/src/Setup/Common/Password.tsx
+++ b/apps/extension/src/Setup/Common/Password.tsx
@@ -14,8 +14,8 @@ const validatePassword = (
   password: string,
   passwordMatch: string
 ): boolean => {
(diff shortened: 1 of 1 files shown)


neocybereth:
Screenshot 2025-03-19 at 12 53 22 PM

mateuszjasiuk (approved):
Tested! LGTM!

PR #1926, opened 2025-03-20
SDK: Support deadline in broadcast Tx

### LINKED ISSUE #1925: SDK: Add deadline override to broadcast_tx
This issue relates to an issue where broadcast Tx is timing out. We currently don't have a way to specify deadline, so this issue is to implement broadcast where deadline can be specified, therefore treating this as a feature though it relates to a reported bug.

More details _TBD_

jurevans:
Resolves  #1925 

- [ ] Implement `broadcastTx` in a way where deadline can be specified
- [ ] Determine this is the proper behavior for MaspTx worker as well


mateuszjasiuk (approved):
LGTM!

Issue #1791, opened 2025-02-28
Extension: UX Improvements


jurevans:
**Ledger Import / MASP Signing**

- Zip32 form for Ledger either needs field borders, or a different background as input background is black, it's difficult to see the fields
- When signing MASP-related Tx, the extension shows `Connecting to Ledger,` when it should really prompt the user to wait to approve on Ledger in this step
- Ledger key import steps need a loading icon spinner thing
  - When the warning shows on Ledger import (that shielded keys will be available 3.0.0+), message + loading icon should be centered vertically
- Viewing Key/Spending Key field (in general) can _sometimes_ trigger a scrollbar - maybe we should consider a monospaced font, or allow for more tolerance here to avoid this?

**ETC**

- If connecting to Namadillo when the Keychain is locked, after entering password, the Approve Origin/Chain view is briefly visible (doesn't seem to be consistent in all browsers, but certainly is noticeable on Chromium)

PR #1927, opened 2025-03-20
FIX: Keychain - Don't display approve connection form if already approved

jurevans:
Relates to #1791 

- [x] Fix issue where `Approve Connection` form displays briefly even though domain/chain ID are already approved
  - To test - Lock keychain, and _from Namadillo_, from an approved domain/chain ID, click `Connect Keychain` - upon successful authentication, you shouldn't see the `Approve Connection` form flash before popup closes (this now displays only if the check approved site has happened before determining if anything should be displayed)
  - Remove Site in keychain (unapprove it), lock keychain, then connect again from Namadillo - should see the form as expected


mateuszjasiuk (approved):
Tested! LGTM!

PR #1928, opened 2025-03-21
chore: fix clippy warnings

mateuszjasiuk:
- fixed clippy warnings
- added build target to `.cargo/config.toml`
- set `"web"` as default feature and added `--no-default-features` in build script, because of that:
  - LSP correctly displays clippy warnings in code
  - I had to add `--no-default-features` so we do not build both web and node at the same time
  - for developing node features we can temporarily replace "web" with "node" in `Cargo.toml` - this only affects LSP code highlighting/checking


jurevans (approved):
Nice!

PR #1934, opened 2025-03-24
fix: Namadillo Status Indicator conflates RPC and Indexer status

### LINKED ISSUE #1164: Namadillo Status Indicator conflates RPC and Indexer status
The Status Indicator label says "Fully synced" when the RPC node is caught up, but the height and epoch are taken from the Namada Indexer, which can be inactive or catching up while the RPC is still running and caught up.

neocybereth:
Closes #1164


neocybereth:
Screenshot 2025-03-25 at 12 44 35 PM
Screenshot 2025-03-25 at 12 44 42 PM

PR #1945, opened 2025-03-28
fix: check that the chain in custom address matches address prefix

neocybereth:

### DIFF
--- a/apps/namadillo/src/App/Transfer/CustomAddressForm.tsx
+++ b/apps/namadillo/src/App/Transfer/CustomAddressForm.tsx
@@ -1,25 +1,15 @@
 import { Input, Stack } from "@namada/components";
 
-import { Chain } from "@chain-registry/types";
 import { chain as osmosis } from "chain-registry/mainnet/osmosis";
-import { Dispatch, SetStateAction, useMemo, useState } from "react";
+import { useMemo, useState } from "react";
 import namadaShieldedSvg from "./assets/namada-shielded.svg";
 import namadaTransparentSvg from "./assets/namada-transparent.svg";
-
-const bech32PrefixMap: Record<string, string> = {
-  osmo: "osmo",
-  znam: "namada",
-  tnam: "namada",
-  cosmos: "cosmos",
-};
 
 type CustomAddressFormProps = {
   onChangeAddress?: (address: string) => void;
   customAddress?: string;
   memo?: string;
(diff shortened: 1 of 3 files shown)


apps/namadillo/src/App/Transfer/TransferModule.tsx
```diff
@@ -471,6 +472,7 @@ export const TransferModule = ({
             )}
             isShieldedAddress={destination.isShieldedAddress}
             isShieldedTx={isShieldedTx}
+            setIsDisabled={setIsDisabled}
```
neocybereth, line 475: Feels kinda weird to prop drill just for the CustomAddressForm here.

But I didn't see any other way to disable the `Submit` from an address error.

apps/namadillo/src/App/Transfer/CustomAddressForm.tsx
```diff
@@ -1,50 +1,102 @@
...
+    return Object.keys(bech32PrefixMap).find((p) => address.startsWith(p));
+  };
+
+  const checkIfAddressMatchesChain = (
+    address: string,
+    chain: Chain | undefined
+  ): boolean => {
+    if (!chain || !address) return false;
+    const prefix = getAddressPrefix(address);
+    return prefix ? bech32PrefixMap[prefix] === chain.bech32_prefix : false;
```
euharrison, line 52: Could something like this be simpler? 
```tsx
address.startsWith(chain.bech32_prefix)
```

apps/namadillo/src/App/Transfer/TransferModule.tsx
euharrison: I don't think it should be a blocker, but in case you want to explore alternatives here, maybe you could try to create a new option for `ValidationResult` that handle this case. It should disable the submit button

Screenshot 2025-03-31 at 17 13 02

euharrison (approved):
Some comments, but not a blocker. The feature looks good! 👍

apps/namadillo/src/App/Transfer/TransferModule.tsx
neocybereth: I updated to use this and it's wayyyy better...so much less surface area of the code touched.

Thank you for the suggestion.

neocybereth: Screenshot 2025-04-01 at 12 18 53 PM

PR #1946, opened 2025-03-28
feat: SDK - Add lookup for TxResponse result codes

### LINKED ISSUE #1898: Namadillo/SDK: Map readable responses to ResultCode Enum (Errors)
Currently, we return the full output when the SDK returns any errors.

We can make use of the `code` param of `TxResponse` and map readable errors to this:

```rust
pub enum ResultCode {
    // WARN: These codes shouldn't be changed between version!
    // =========================================================================
    /// Success
    #[default]
    Ok = 0,
    /// Error in WASM tx execution
    WasmRuntimeError = 1,
    /// Invalid tx
    InvalidTx = 2,
    /// Invalid signature
    InvalidSig = 3,
    /// The block is full
    AllocationError = 4,
    /// Replayed tx
    ReplayTx = 5,
    /// Invalid chain ID
    InvalidChainId = 6,
    /// Expired tx
    ExpiredTx = 7,
    /// Exceeded gas limit
    TxGasLimit = 8,
    /// Error in paying tx fee
    FeeError = 9,
    /// Invalid vote extension
    InvalidVoteExtension = 10,
    /// Tx is too large
    TooLarge = 11,
    /// Tx code is not allowlisted
    TxNotAllowlisted = 12,
    // =========================================================================
    // WARN: These codes shouldn't be changed between version!
}
````

### PR DESCRIPTION
**Pull request by jurevans**
Resolves #1898 

See `ResultCode` definition at:

https://github.com/anoma/namada/blob/427a57cab8c93a18f293b00e9f54e6e9da5ac147/crates/tx/src/data/mod.rs#L42-L90

- [x] Provide `ResultCode` definitions in `@namada/types`
- [x] Provide result code text lookup in SDK
- [x] Hook up to Namadillo when a `TxResponse` is returned from chain

To update these error messages, we simply need to make the change here: https://github.com/anoma/namada-interface/blob/1f1962bd14f7134c25a8bd6739f53937eae79ac1/packages/types/src/tx/types.ts#L102-L116

Broadcast error messages returned from Namada SDK can now be parsed as JSON. The response before was being formatted in `broadcast_tx` in Namada SDK, so we'd end up with an error like this:

```
Encountered error while broadcasting transaction: StringTracer: server error: {"codespace":"","code":8,"data":"","log":"Mempool validation failed: Wrapper transaction exceeds its gas limit","hash":"11AE6D481A56ABB762509198DE7F5088EF1772A281F0FFCEDD86E6273F813AC7"}
```

Deadline timeout should return a normal, non-JSON string.

Introduced a custom error `BroadcastTxError` that can parse a `TxResponse` from the SDK, otherwise will return the original error string for non-TxResponse results.

### Testing

- Trigger a gas error by decreasing gas limit to an unreasonable amount - should get appropriate message
- Trigger a deadline error by updating `deadline` to `BigInt(1)`, should get appropriate message in notification

Tested with normal accounts (batch tx) and Ledger (with multiple Txs)

### Screenshots

Gas limit exceeded:
image

Deadline timeout:
image

### DIFF
--- a/apps/namadillo/src/lib/query.ts
+++ b/apps/namadillo/src/lib/query.ts
@@ -234,8 +234,7 @@ export const broadcastTxWithEvents = async <T>(
       if (result.status === "fulfilled") {
         return result.value.commitments;
       } else {
-        // Throw wrapper error if encountered
-        throw new Error(`Broadcast Tx failed! ${result.reason}`);
+        throw new Error(result.reason.toString());
       }
     });
 
--- a/packages/sdk/src/rpc/rpc.ts
+++ b/packages/sdk/src/rpc/rpc.ts
@@ -6,6 +6,7 @@ import {
   TransferToEthereum,
 } from "@namada/shared";
 import {
+  BroadcastTxError,
   DatedViewingKey,
   TxResponseMsgValue,
   TxResponseProps,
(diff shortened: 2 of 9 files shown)


mateuszjasiuk (approved):
Tested LGTM!

PR #1951, opened 2025-04-01
feat: Overview Dashboard v2

pedrorezende:
This PR contains a new version of the overview page, or the main dashboard for Namadillo. 

image


apps/namadillo/src/App/routes.ts
```diff
@@ -19,6 +19,7 @@ export const routes = {
   masp: "/masp",
   maspShield: "/masp/shield",
   maspUnshield: "/masp/unshield",
+  shieldAssets: "/masp/shield-assets",
```
pedrorezende, line 22: @euharrison you were right :P

apps/namadillo/src/App/AccountOverview/ShieldedAssetsOverview.tsx
```diff
@@ -0,0 +1,62 @@
...
+          isShielded={true}
+          footerButtons={
+            <>
+              <ActionButton
+                outlineColor="yellow"
+                size="xs"
+                className="w-auto px-4"
+              >
+                <span className={clsx(textContainerClassList)}>
+                  Shielded Transfer
```
neocybereth, line 31: Screenshot 2025-04-01 at 5 43 40 PM

Neither of these buttons work for me. I def have balance to `Unshield`

apps/namadillo/src/App/AccountOverview/StakeYourNamCard.tsx
```diff
@@ -0,0 +1,34 @@
...
+  const location = useLocation();
+  return (
+    <div
+      className={clsx(
+        "flex items-center justify-between gap-12 text-sm text-cyan bg-neutral-900 rounded-sm px-6",
+        "py-6"
+      )}
+    >
+      <span className="max-w-[15ch] text-center leading-tight">
+        Stake your Transparent NAM
```
neocybereth, line 17: Screenshot 2025-04-01 at 5 43 53 PM

This button doesnt bring up the Staking route/modal. It just scrolls to the top of the page.

apps/namadillo/src/App/AccountOverview/UnshieldedAssetsOverview.tsx
```diff
@@ -0,0 +1,64 @@
+import { ActionButton, Panel, SkeletonLoading } from "@namada/components";
+import { routes } from "App/routes";
+import { transparentTokensAtom } from "atoms/balance";
+import { useAmountsInFiat } from "hooks/useAmountsInFiat";
+import { useAtomValue } from "jotai";
+import { useLocation, useNavigate } from "react-router-dom";
+import { StakeYourNamCard } from "./StakeYourNamCard";
+import { TotalBalanceCard } from "./TotalBalanceCard";
+import { UnshieldedAssetTable } from "./UnshieldedAssetTable";
+
```
neocybereth, line 10: Screenshot 2025-04-01 at 5 47 23 PM


Can we make these two icons have a tooltip?

It's not really clear to users what they do without clicking them and navigating away to a new page.

neocybereth (commented):
https://github.com/user-attachments/assets/eb1993aa-b0a3-451c-b097-eb2f14ba63ab


Any chance would remember the tab they last clicked? 

For instance:

1) User clicks the "Unshielded Assets" tab 
2) User then clicks "Stake" and are taken to the Stake page
3) User comes back to the Overview page

In this case it should show the "Unshielded Assets" tab IMO, but it currently goes back to the default left tab Shielded Assets (MASP)

apps/namadillo/src/App/AccountOverview/UnshieldedAssetsOverview.tsx
euharrison: @ChrisHoltDesign what do you think? Maybe change it to text directly, like "Shield"? 
This is also related with the new Transfer area which could have different actions and drop (make it invisible for the user) things like 'Ibc transfer' etc

euharrison:
Yeah, I agree preserving the tab, I had the same experience 👍 

An easy way to do it is using the `useUrlState` which is like a regular `useState` but saving the value on the url with query params. There are a few examples of this use case on the transfer forms

apps/namadillo/src/App/AccountOverview/AssetsOverviewPanel.tsx
```diff
@@ -0,0 +1,37 @@
...
+import { useRequiresNewShieldedSync } from "hooks/useRequiresNewShieldedSync";
+import { useState } from "react";
+import { ShieldedAssetsOverview } from "./ShieldedAssetsOverview";
+import { UnshieldedAssetsOverview } from "./UnshieldedAssetsOverview";
+
+export const AssetsOverviewPanel = (): JSX.Element => {
+  const requiresNewShieldedSync = useRequiresNewShieldedSync();
+
+  // Always select the unshielded tab as default if shielded sync is required
+  const [currentTab, setCurrentTab] = useState(requiresNewShieldedSync ? 1 : 0);
```
euharrison, line 12: @pedrorezende what do you think about using the `useUrlState` here?

euharrison (approved):
Only the `useUrlState` comment, but it's not a blocker :)

apps/namadillo/src/App/AccountOverview/AssetsOverviewPanel.tsx
neocybereth: I dont think `useUrlState` is going to even accomplish what we want here unfortunately. 

B/c basically everytime you update the route, by clicking on any buttons, the state is lost.

apps/namadillo/src/App/AccountOverview/UnshieldedAssetsOverview.tsx
neocybereth: Screenshot 2025-04-08 at 8 48 49 AM

PR #1953, opened 2025-04-01
bug: Keychain - Disable shielded key import for nano S

### LINKED ISSUE #1952: [Keychain]: Bug - Disable shielded keys import for Nano-S only
As the Ledger Nano-S does not and will not support shielded features, we should not automatically attempt to import shielded keys when this device is present.

- Can we detect this specific model from the info query? We should be able to, otherwise:
- Gracefully fail the query for shielded keys and allow user to continue

jurevans:
Resolves #1952 

- [x] If device is a Nano S, disable shielded key import
- [x] Update Ledger validation to also check against this (zip32 or masp functions only work if both device and version is correct)
- [x] Update warning copy on import

### Testing

Build/install extension from this branch, then, assuming you are on > v3.0.0 of the Namada App:

Nano S:
- In setup, import your Ledger
- You should see the warning in the screenshot below
- You should be able to successfully import just the transparent keys

Not-Nano S:
- In setup, import your Ledger
- You should not see any warning
- You are guided through 3 steps: Transparent import, Viewing key import, Proof Gen / Auth key import

### Notes

See: https://github.com/LedgerHQ/ledger-live/blob/06cfbf04e2d99e4e7916ace7a10cff8d37c5c38a/libs/ledgerjs/packages/types-devices/src/index.ts#L4

Ledger Live has enumerated the available device IDs here:
```typescript
/**
 * DeviceModelId is a unique identifier to identify the model of a Ledger hardware wallet.
 */
export enum DeviceModelId {
  blue = "blue",
  nanoS = "nanoS",
  nanoSP = "nanoSP",
  nanoX = "nanoX",
  stax = "stax",
  europa = "europa",
}
```

### Screenshots

**NOTE** I forced this message by setting the blacklisted device as `nanoSP` -  this should be only be displayed for the `nanoS`:

image

### DIFF
--- a/apps/extension/src/Setup/Ledger/LedgerConnect.tsx
+++ b/apps/extension/src/Setup/Ledger/LedgerConnect.tsx
@@ -59,8 +59,8 @@ export const LedgerConnect: React.FC<Props> = ({
       if (returnCode !== LedgerError.NoErrors) {
         throw new Error(errorMessage);
       }
-      const canImportShielded = await ledger?.isZip32Supported();
-      setIsZip32Supported(canImportShielded);
+      const isMaspSupported = await ledger?.isZip32Supported();
+      setIsZip32Supported(isMaspSupported);
 
       setIsLedgerConnecting(true);
       setCurrentApprovalStep(1);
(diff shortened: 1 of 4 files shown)


mateuszjasiuk (approved):
Tested and LGTM! Changed `deviceId` from `nanoS` to `nanoSP` for testing!

Issue #1907, opened 2025-03-14
Namadillo: Claim and stake time out error, shows weird NAM value


ChrisHoltDesign:
- Tried to claim and stake
- Times out
- The display then shows **NaN NAM**

Stake claim balance went to zero and was partially applied and part went to non staked balance

<img width="624" alt="Image" src="https://github.com/user-attachments/assets/a909065a-3fd3-4af7-bfb7-4742c03b568c" />

Rigorously:
I claimed and staked with Gas Amount set to 50,000. Namadillo showed the grey "in progress" popup indefinitely. 

The [transaction was successful according to the Valopers explorer](https://namada.valopers.com/transactions/89008e77e0eb973a9fc9f2ddb9145077a0a7648a5f0815ebbce810cc9907128c). 

Using the command line interface to compare the balance before and after the transaction however indicates that claim was successful, but staking was not.

euharrison:
@Rigorously I had the same issue on old versions of Namadillo + Indexer

But I tested now on https://interface.housefire.tududes.com/ (1.18.0 + 2.30) and changing the gas fee to something really low (500), it dispatches the error notification and show the error on the interface

Can you test it again?, In case you still have the same issue, do you mind sharing which Namadillo are you using? Or versions?

In case the bug persist, I'll create a new issue so we don't mix with the `NaN` issue :) 

![Image](https://github.com/user-attachments/assets/03590616-39f4-4e58-82d3-9c7ded144aff)

euharrison:
Issue for wrong gas price for Claim & Stake https://github.com/anoma/namada-interface/issues/1961

PR #1957, opened 2025-04-02
fix: fix claim and stake NaN issue

euharrison:
Fix claim and stake NaN issue.

Closes https://github.com/anoma/namada-interface/issues/1907

This is happening in case the transaction returns an error because the stake didn't applied, but the claim succeed. This causes the return the error message, but the validator rewards is empty: `rewards[r.validator] === undefined`

To simulate this, you can:
1 - Replace the validator address for anything else `rewards['something-else']` on `ClaimRewardsSubmitModalStage.tsx:39`
2 - In case you want to submit the transaction, you can simulate an error by adding `throw new Error(`Broadcast Tx failed! Mock timeout error`);` on `src/lib/query.ts:233`

Screenshot 2025-04-02 at 20 47 40


mateuszjasiuk (approved):
LGTM! :)

PR #1958, opened 2025-04-03
fix: make sure to refresh after an unshield

### LINKED ISSUE #1902: After unshield, we need to update the transparent balance
- I have some OSMO on MASP
- I successfully unshielded some osmo  (green notification and screen update)
- I navigated to the Overview, and the balance wasn't updated
- I refreshed the browser and I was able to see the new balance

We need to refetch the balance, and make sure that all `queryKey` and `queryDependencies` are set correctly for the Overview page

neocybereth:
Closes #1902

### DIFF
--- a/apps/namadillo/src/App/Masp/MaspUnshield.tsx
+++ b/apps/namadillo/src/App/Masp/MaspUnshield.tsx
@@ -8,8 +8,12 @@ import {
   TransferModule,
 } from "App/Transfer/TransferModule";
 import { allDefaultAccountsAtom } from "atoms/accounts";
-import { namadaShieldedAssetsAtom } from "atoms/balance/atoms";
+import {
+  namadaShieldedAssetsAtom,
+  shieldedBalanceAtom,
+} from "atoms/balance/atoms";
 import { chainParametersAtom } from "atoms/chain/atoms";
+import { shouldUpdateBalanceAtom } from "atoms/etc";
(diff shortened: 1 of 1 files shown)


euharrison (approved):
I'm ok with this PR, but I would like to instigate us to think more about our solutions in general

My main concern is: if keep adding custom code for each transaction, we'll always miss border cases

For example:
- Why only unshield needs this manual update and shield or shielded transfers don't need it?
- Why the `onTransferSuccess` on `useTransactionCallbacks.tsx` is not handling this case?

Maybe answering these questions helps us to have a better control over the app as a whole

ps: If this is related to the issue I created, maybe it was already fixed and if we can't reproduce it anymore, we could simply close it

neocybereth:
> I'm ok with this PR, but I would like to instigate us to think more about our solutions in general
> 
> My main concern is: if keep adding custom code for each transaction, we'll always miss border cases
> 
> For example:
> 
> * Why only unshield needs this manual update and shield or shielded transfers don't need it?
> * Why the `onTransferSuccess` on `useTransactionCallbacks.tsx` is not handling this case?
> 
> Maybe answering these questions helps us to have a better control over the app as a whole
> 
> ps: If this is related to the issue I created, maybe it was already fixed and if we can't reproduce it anymore, we could simply close it

Yeah I need to actually get enough tokens to be able to `unshield` b/c this could be fixed by my last fix here:

PR #1959, opened 2025-04-03
fix: update logic for ibc transfers from NAM

neocybereth:
Transparent transfers were checking for custom chains per [this](https://github.com/anoma/namada-interface/pull/1945) PR, but that's only necessary on IBC transfers *FROM* Namada

### DIFF
--- a/apps/namadillo/src/App/Transfer/TransferModule.tsx
+++ b/apps/namadillo/src/App/Transfer/TransferModule.tsx
@@ -122,12 +122,8 @@ type ValidationResult =
   | "NoLedgerConnected"
   | "Ok";
 
-/**
- * Checks if a custom address matches the specified destination chain
- * and validates shielded/transparent address requirements
- * Only used for IBC transfers *FROM* Namada
- */
-const addressMatchesCustomChain = ({
+// Check if the provided address is valid for the destination chain and transaction type
+const isValidDestinationAddress = ({
   customAddress,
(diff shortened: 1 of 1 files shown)


apps/namadillo/src/App/Transfer/TransferModule.tsx
```diff
@@ -202,16 +208,16 @@ export const TransferModule = ({
   }, [source.selectedAssetAddress, source.availableAmount, displayGasFee]);
 
   const validationResult = useMemo((): ValidationResult => {
+    console.log(destination.customAddress ?? "", destination.chain);
```
mateuszjasiuk, line 211: log :)

mateuszjasiuk (commented):
I think this looks good overall!
It would be nice to add this logic also in the Transfer tab as right now we can do this:

obraz

neocybereth:
> I think this looks good overall! It would be nice to add this logic also in the Transfer tab as right now we can do this:
> 
> obraz

Awesome, I agree. I'll merge this one and open that as a separate issue so this bug fix is resolved.

apps/namadillo/src/App/Transfer/TransferModule.tsx
neocybereth: thank you!

mateuszjasiuk (approved):

PR #1962, opened 2025-04-03
fix: fix z index

euharrison:
- Fix z-index of Shielded Sync overlay
- Fix hit area conflict on Staking buttons because of position relative/absolute conflicts
- Fix left nav top of 113px (instead of 112px)

https://github.com/user-attachments/assets/ffda4e9c-d3f3-4374-8c14-42bc48f27af0


neocybereth (approved):
LGTM!

Issue #1922, opened 2025-03-19
Keychain: Handle multiple chain IDs


jurevans:
I think it would be an excellent improvement to track approved chain IDs per domain, instead of what we currently do which is to allow a single signing-enabled chain ID at any given time. The representation in state would essentially look like the following:

```typescript
export type ApprovedDomainsStore = Record<string, string[]>;
```

```json
{
  "https://interface.namada.tududes.com/": ["namada.5f5de2dd1b88cba30586420", "housefire-alpaca.cc0d3e0c033be"],
  "https://interface.campfire.tududes.com/": ["campfire-square.ff09671d333707", "housefire-alpaca.cc0d3e0c033be"],
  "https://interface.housefire.tududes.com/": [""housefire-alpaca.cc0d3e0c033be"],
}
```

Similar to Keplr, our _Connected Sites_ view should allow revocation of any given chain ID within an approved domain, or removal of an entire domain at the top level, basically this:

- [ ] Domain 1
  - [ ] Chain ID 1
  - [ ] Chain ID 2
- [ ] Domain 2
  - [ ] Chain ID 3
  - [ ] Chain ID 1

PR #1968, opened 2025-04-07
feat: Keychain - Display chain name for known chains

### LINKED ISSUE #1923: Keychain: Make it more clear which network is connected when signing & approving connection
Since we have names for long-running testnets, we could display this when user _approves_ a connection to an app, as well as when they are signing. This would help prevent the user unintentionally signing a Tx destined for the wrong network.

We could hardcode known network names (matching our registry) for at least these chains:

- **Namada Mainnet** - `namada.5f5de2dd1b88cba30586420`
- **Housefire Testnet** - `housefire-alpaca.cc0d3e0c033be`
- **Campfire Testnet** - `campfire-square.ff09671d333707`

Also, it was suggested that we also show the origin which requested the signing.

This could really be extended to Namadillo as well. It would be nice for it to be extremely obvious which network you are connected to across the board.

jurevans:
Resolves #1923 

If `chainId` is known (mainnet or long-running testnet) display name, otherwise, just chain ID, and show origin (the origin that issues the call to the keychain) on signing approvals.

- [x] Add lookup for chain name based on chain ID
- [x] Add chain name / chain ID and origin to `ApproveConnection`
- [x] Add origin to `ApproveSignTx`
- [x] Add chain name to `Network` if a definition exists

**NOTE** The reason I added `chainIds` as as `string[]` to `ApproveSignTx` is because, in the case of Ledger, it is _possible_ to submit multiple Txs with different chain IDs, so this will display all of them (this should never really happen, but if say another interface attempts to submit it this way, we will display them). 

### Testing

- Approve Connection: With an unapproved chain/domain, connect to keychain - should see the chain name if it is one of the three known networks, otherwise, just the chain ID
- Submit transactions for signing - should see origin & chain ID
- Also test Ledger with multiple tx (such as bond), ensure that for Namadillo we only see one Chain ID.
- Check Network form, changing chain ID will either show the associated chain, or nothing if it is not a known change (users will likely never use this form, and it will go away with https://github.com/anoma/namada-interface/issues/1922 anyways)
- Submit Sign Arbitrary - should see origin in approval

#### Approve Connection
_Unknown_ chain (this is a test, I made the campfire chain ID invalid)
image

_Known_ chain
image

#### Approve Sign Tx

image

### Network

image

### DIFF
--- a/apps/extension/src/App/Settings/Network.tsx
+++ b/apps/extension/src/App/Settings/Network.tsx
@@ -7,6 +7,7 @@ import {
   Input,
   Stack,
 } from "@namada/components";
+import { NamadaChains } from "@namada/types";
 import { PageHeader } from "App/Common";
 import { GetChainMsg, UpdateChainMsg } from "background/chain";
 import { useRequester } from "hooks/useRequester";
@@ -27,6 +28,7 @@ export const Network = (): JSX.Element => {
   const [errorMessage, setErrorMessage] = useState("");
 
   const shouldDisableSubmit = status === Status.Pending || !chainId;
+  const chainIdentifier = NamadaChains.get(chainId);
 
   useEffect(() => {
     void (async () => {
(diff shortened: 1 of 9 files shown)


neocybereth (approved):
https://heliax.slack.com/archives/C03JY8E9571/p1744035478035359

PR #1973, opened 2025-04-08
feat: namadillo v2 feedback

### LINKED ISSUE #1972: Namadillo V2 feedback
- [x] 'Overview' should be renamed to 'Manage Assets'

Image

----------

- [x] Remove the MASP tab. This has now been ported into the 'Manage Assets' Shielded Tab. The internal shield and unshield can function as normal without needing to highlight.

Image

----------

- [x] Lets remove the cluster icons and the button for now. Too many prompts to shield currently and not all those assets can be shielded

Image

----------

- [ ] We should not have this gap at the bottom. It should fill the screen regardless of assets held

Image

----------

- [x] Lets remove the timeline on the right hand side. We are informing users elsewhere as to what phase we are in and what is enabled

Image

We can replace the gap with a link to discord 

Image

----------

- [x] Shielded Rewards block should be right aligned like the equivalent staking option on transparent

Image

----------

- [x] Lets kill these little icons. I need to draw something properly at a later date

Image

----------

- [x] Lets add in an transfer icon to the shielded assets also next to unshield

Image

----------

- [ ] We should also add the NAM exclusion text in the $ balances on the other tabs, along with the main balances

Image

Image

neocybereth:
Closes #1972

### DIFF
--- a/apps/namadillo/src/App/Common/TableWithPaginator.tsx
+++ b/apps/namadillo/src/App/Common/TableWithPaginator.tsx
@@ -52,7 +52,7 @@ export const TableWithPaginator = <T,>({
         headProps={headProps}
       />
     );
-  }, [rows, tableProps, headProps]);
+  }, [headers, id, rows, tableProps, headProps]);
 
   const pagination = useMemo(() => {
     return (
@@ -66,7 +66,7 @@ export const TableWithPaginator = <T,>({
         }}
       />
     );
-  }, [page, itemList, onPageChange]);
+  }, [page, pageCount, onPageChange, scrollTop]);
 
   if (rows.length === 0) {
     return (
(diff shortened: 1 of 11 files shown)


apps/namadillo/src/App/Common/TableWithPaginator.tsx
```diff
@@ -52,7 +52,7 @@ export const TableWithPaginator = <T,>({
         headProps={headProps}
       />
     );
-  }, [rows, tableProps, headProps]);
+  }, [headers, id, rows, tableProps, headProps]);
```
euharrison, line 55: Thinking loud, maybe we could enable the lint `react-hooks/exhaustive-deps` as warning for this. But we probably need to accept warnings on the build, otherwise it'll be a big refactoring

euharrison (approved):
LGTM!

Just this top margin on the tabs that looks tight. What do you think of increasing it?

Screenshot 2025-04-11 at 12 00 54

PR #1978, opened 2025-04-14
fix: update apr and table height

neocybereth:

### DIFF
--- a/apps/namadillo/src/App/AccountOverview/UnshieldedAssetTable.tsx
+++ b/apps/namadillo/src/App/AccountOverview/UnshieldedAssetTable.tsx
@@ -144,6 +144,7 @@ const TransparentTokensTable = ({
         Tokens
       </div>
       <TableWithPaginator
+        containerClassName="h-screen"
         id="transparent-tokens"
         headers={headers.concat("")}
         renderRow={renderRow}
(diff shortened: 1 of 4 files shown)


apps/namadillo/src/App/AccountOverview/UnshieldedAssetTable.tsx
```diff
@@ -144,6 +144,7 @@ const TransparentTokensTable = ({
         Tokens
       </div>
       <TableWithPaginator
+        containerClassName="h-screen"
```
euharrison, line 147: It looks like it's a bit bigger than necessary. Maybe you will need to add a bunch of `flex:1` to fix this 😬 


https://github.com/user-attachments/assets/b63c2b0c-f2fb-4a01-910a-f4763e5d1580

euharrison (approved):

apps/namadillo/src/App/AccountOverview/UnshieldedAssetTable.tsx
neocybereth: Yeah the former alternative was too small unfortunately 🫠, this will all be coming in the cleanup PR with Chris' other suggestions

PR #1980, opened 2025-04-15
feat: shielded swaps

mateuszjasiuk:


apps/namadillo/src/App/Swap/SwapSuccess.tsx
```diff
@@ -0,0 +1,168 @@
+import { Heading, Stack } from "@namada/components";
+import anime from "animejs";
+import { useSetAtom } from "jotai";
+import { useCallback, useEffect, useRef } from "react";
+import { CgCheck } from "react-icons/cg";
+import { SwapStatus } from "./state";
+import { swapStateAtom, swapStatusAtom } from "./state/atoms";
+
+export const SwapSuccess = (): JSX.Element => {
```
neocybereth: Screenshot 2025-11-07 at 7 17 21 PM

After a successful swap I get the checkmark and then it goes back here. Are we not showing a Transaction Receipt for the swaps?

apps/namadillo/src/App/Swap/SwapReview.tsx
```diff
@@ -0,0 +1,251 @@
...
+    await performSwap({ localRecoveryAddr: walletAddress });
+  }, [walletAddress]);
+
+  const validationResult = useSwapReviewValidation({
+    walletAddress,
+    ledgerAccountInfo,
+  });
+
+  // We stop the ledger status check when the transfer is in progress
+  setLedgerStatusStop(["Building", "AwaitingSignature"].includes(status.t));
```
neocybereth: should we move this to inside a useEffect?

apps/namadillo/src/App/AppRoutes.tsx
```diff
@@ -99,6 +100,11 @@ export const MainRoutes = (): JSX.Element => {
             <Route path={routes.maspUnshield} element={<div />} />
           </Route>
 
+          {/* Swapping */}
+          <>
```
neocybereth: seems we can remove the fragment

apps/namadillo/src/App/Swap/SwapCalculations.tsx
```diff
@@ -0,0 +1,353 @@
...
+  const { data: assetsWithBalance } = useAtomValue(namadaShieldedAssetsAtom);
+  const { data: tokenPrices } = useAtomValue(
+    tokenPricesFamily(sortedAssets.map((a) => a.address))
+  );
+  const feeProps = useTransactionFee(["IbcTransfer"]);
+  const shieldedAccountAddress = useAtomValue(
+    defaultShieldedAccountAtom
+  )?.address;
+
+  useSwapSimulation();
```
neocybereth: should we put this in a useEffect with empty dependency array?

apps/namadillo/src/App/Swap/SwapReview.tsx
```diff
@@ -0,0 +1,251 @@
...
+            outlineColor="yellow"
+            backgroundColor="transparent"
+            backgroundHoverColor="yellow"
+            textColor="yellow"
+            textHoverColor="black"
+            onClick={() => setStatus(SwapStatus.idle())}
+          >
+            Back
+          </ActionButton>
+        )}
```
neocybereth: can make these ternary instead of two `&&`

apps/namadillo/src/App/Swap/SwapSource.tsx
```diff
@@ -0,0 +1,94 @@
+import { AmountInput, Text } from "@namada/components";
+import { AvailableAmountFooter } from "App/Common/AvailableAmountFooter";
+import { SelectedAsset } from "App/Common/SelectedAsset";
+import { TokenAmountCard } from "App/Common/TokenAmountCard";
+import BigNumber from "bignumber.js";
+import clsx from "clsx";
+import { Asset } from "types";
+
+export type SwapSourceProps = {
+  walletAddress?: string;
```
neocybereth: Now with these types they're all optional and become `string | undefined` but some are probably non-optional eh? :D

apps/namadillo/src/App/Swap/state/atoms.ts
```diff
@@ -0,0 +1,182 @@
...
+        swapState.mode === "sell" ?
+          toBaseAmount(sellAsset, swapState.sellAmount!)
+        : swapState.mode === "buy" ?
+          toBaseAmount(buyAsset, swapState.buyAmount!)
+        : toBaseAmount(buyAsset, BigNumber(1));
+
+      const simulateSell =
+        swapState.mode === "sell" || swapState.mode === "none";
+
+      const params: Record<string, string> =
```
neocybereth: can remove this casting and update the type in `fetchQuote` if you want....up to you

neocybereth (changes requested):

apps/namadillo/src/App/Swap/SwapInProgress.tsx
```diff
@@ -0,0 +1,99 @@
+import { Heading, Stack } from "@namada/components";
+import anime from "animejs";
+import { clearDisposableSigner } from "atoms/transfer/services";
+import { useAtom } from "jotai";
```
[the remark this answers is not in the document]
mateuszjasiuk: I can move it up, it's 200px from top as in figma

apps/namadillo/src/App/Swap/SwapSuccess.tsx
mateuszjasiuk: No we are not, here is the design https://www.figma.com/design/NFyHbLZXBSl3aUsMxtffvV/Namada-End-User-Interface-V1.0---Phase-1--Block-Party?node-id=10113-2166&t=XcYgWVYTci6BBFNL-0. I assume it's because we do not consider swap as a kind of transfer

apps/namadillo/src/App/Swap/SwapModule.tsx
```diff
@@ -0,0 +1,42 @@
...
+
+        {["Confirming", "Broadcasting"].includes(status.t) && (
+          <SwapInProgress />
+        )}
+        {status.t === "Completed" && <SwapSuccess />}
+
+        {!["Broadcasting", "Confirming", "Completed", "Error"].includes(
+          status.t
+        ) && (
+          <p className="w-full mt-6 text-center font-light">
```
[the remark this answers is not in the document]
mateuszjasiuk: I can ask Chris or Diego

apps/namadillo/src/App/Swap/SwapCalculations.tsx
mateuszjasiuk: We can't cause we break react hook rules, I don't think it's that bad tho as we use `useEffect` inside `useSwapSimulation`

apps/namadillo/src/App/Swap/SwapModule.tsx
```diff
@@ -0,0 +1,42 @@
+import { Panel } from "@namada/components";
+import { useAtomValue } from "jotai";
+import { swapStatusAtom } from "./state/atoms";
+import { SwapCalculations } from "./SwapCalculations";
+import { SwapHeader } from "./SwapHeader";
+import { SwapInProgress } from "./SwapInProgress";
+import { SwapReview } from "./SwapReview";
+import { SwapSuccess } from "./SwapSuccess";
+
+// TODO: for better state isolation, we should create a new store for the swap module
```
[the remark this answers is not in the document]
mateuszjasiuk: Nah, this is something that we should probably do everywhere but we do not do this anywhere. Basically in jotai you can wrap components in `<Provider />` which would make component inside use different store. This is good for isolating state, now out global state is mixed up with feature states

apps/namadillo/src/App/Swap/SwapCalculations.tsx
```diff
@@ -0,0 +1,353 @@
...
+
+const SwapCalculationsFooter = ({
+  feeProps,
+  sellAmountPerOneBuy,
+  selectedAsset,
+  selectedTargetAsset,
+  tokenPrice,
+  quote,
+}: SwapCalculationsFooterProps): JSX.Element => {
+  // Quote cache, prevents blinking when quote is temporarily undefined
```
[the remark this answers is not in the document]
neocybereth: Sounds good to me, I trust your judgement.

apps/namadillo/src/App/Swap/SwapSource.tsx
mateuszjasiuk: I see your point but `string | undefined` is not a solution. It makes passing a prop to component mandatory, but it does not change the fact that the value can be undefined. So prop is not optional but the value inside prop still is. It only helps in scenarios in which someone forgets to pass a prop. Component still has to handle undefined. Components should not have so many potentially undefined props, because then you never know what you really get at which point. We should probably use more of this:
https://jotai.org/docs/guides/async#suspense

apps/namadillo/src/App/Swap/SwapInProgress.tsx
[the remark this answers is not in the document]
mateuszjasiuk: changed `mt` from 200 to 150px

apps/namadillo/src/App/Swap/SwapModule.tsx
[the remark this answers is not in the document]
mateuszjasiuk: added

PR #1991, opened 2025-04-28
WIP: Namadillo - Return original error if it is not JSON parseable

jurevans:
Fixes an issue where the utility mapping custom errors attempts to parse JSON when there is no JSON.

### Screenshot
image


mateuszjasiuk (approved):
LGTM!

PR #1999, opened 2025-05-06
SDK: Updating toolchain, SDK, Namadillo chain params

jurevans:
Updates introduced:

- Toolchain version: 1.85.1
- `rust-toolchain.toml` - nightly-2025-03-27
- `namada_tx` = libs-v0.149.1
- `namada_sdk` = libs-v0.149.1
- `tendermind-config` = 0.40.4

These are needed to support the latest MASP-indexer on the testnets!

### TODO
We need an update in the Indexer client with the updated type for `checksums` in `Parameters`. I bumped the version, but it still does not contain the new fields, so I introduced a temporary type to cast it as.

### Screenshots

This PR also fixes issue with blank Keychain approval:

image

### DIFF
--- a/.github/workflows/verify-pull-request.yml
+++ b/.github/workflows/verify-pull-request.yml
@@ -56,6 +56,9 @@ jobs:
       - name: Install yarn dependencies
         uses: ./.github/actions/yarn-cache
 
+      - name: Setup Rust toolchain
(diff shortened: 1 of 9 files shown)


neocybereth (approved):
LGTM

PR #2012, opened 2025-05-08
fix: fix new ibc tokens not showing

euharrison:
Fix the new tokens missing from the chain/asset selector when executing a transfer.

- Add the missing chains
- Update the `namada-chain-registry` to pull the new assets from there, instead of merging with the osmosis one
- Rename some variables to help to understand the code
- Penumbra is out of this PR because they have another issue. They are listed as `non-cosmos` chain and they don't have a `chain.json` file. We need to handle this differently. https://github.com/cosmos/chain-registry/tree/master/_non-cosmos/penumbra

https://github.com/user-attachments/assets/5c8750f6-7dc6-4937-9a67-3afbe00a36cc

### DIFF
--- a/apps/namadillo/package.json
+++ b/apps/namadillo/package.json
@@ -10,7 +10,7 @@
     "@chain-registry/client": "^1.53.5",
     "@cosmjs/encoding": "^0.32.3",
     "@keplr-wallet/types": "^0.12.136",
-    "@namada/chain-registry": "^1.0.0",
+    "@namada/chain-registry": "github:anoma/namada-chain-registry#5a884bf6343f5130c4a419901a87ea86df53dd83",
     "@namada/indexer-client": "2.5.4",
     "@tailwindcss/container-queries": "^0.1.1",
     "@tanstack/query-core": "^5.40.0",
--- a/apps/namadillo/package.json
+++ b/apps/namadillo/package.json
@@ -10,7 +10,7 @@
     "@chain-registry/client": "^1.53.5",
     "@cosmjs/encoding": "^0.32.3",
     "@keplr-wallet/types": "^0.12.136",
-    "@namada/chain-registry": "github:anoma/namada-chain-registry#5a884bf6343f5130c4a419901a87ea86df53dd83",
+    "@namada/chain-registry": "^1.2.0",
     "@namada/indexer-client": "2.5.4",
     "@tailwindcss/container-queries": "^0.1.1",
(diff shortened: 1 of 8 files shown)


apps/namadillo/src/atoms/integrations/atoms.ts
```diff
@@ -122,12 +122,6 @@ export const availableChainsAtom = atom((get) => {
   return getKnownChains(settings.advancedMode).map(({ chain }) => chain);
 });
 
-// Lists only the available assets list
-export const availableAssetsAtom = atom((get) => {
```
euharrison, line 126: This is dead code, not used anywhere

apps/namadillo/src/atoms/integrations/functions.ts
```diff
@@ -56,7 +59,15 @@ registry.assets.push(
...
-const mainnetChains: ChainRegistryEntry[] = [celestia, cosmos, osmosis, stride];
+const mainnetChains: ChainRegistryEntry[] = [
+  celestia,
+  cosmos,
+  osmosis,
+  stride,
+  neutron,
+  noble,
+  nyx,
+];
```
euharrison, line 70: The magic is here, lets remember to add the new chains here as well when he have even more tokens. Or think in a smart solution to get this dynamically from somewhere

neocybereth: Can we also add a note above this like: "This is the array we must update to add new assets" or something similar

neocybereth:
Select Asset dropdown doesnt seem to register Neutron assets: 
Screenshot 2025-05-09 at 12 55 13 PM
Screenshot 2025-05-09 at 12 55 07 PM

apps/namadillo/package.json
```diff
@@ -10,7 +10,7 @@
     "@chain-registry/client": "^1.53.5",
     "@cosmjs/encoding": "^0.32.3",
     "@keplr-wallet/types": "^0.12.136",
-    "@namada/chain-registry": "^1.0.0",
+    "@namada/chain-registry": "github:anoma/namada-chain-registry#5a884bf6343f5130c4a419901a87ea86df53dd83",
```
jurevans, line 13: quick question: Should we go ahead and merge the release branch on the registry, once everything we need is in place? Then we can just update this to that tag

euharrison: good idea!

PR #2020, opened 2025-05-14
feat: improve error message on wasm errors

mateuszjasiuk:
Before we did not return error when `tx_response.code != 0`. Because of that tx response error were not handled and we would get timeout notification instead of error.

How to test:
try to claim and stake all(with default gas or lower it a bit) and see if you get error like this:
image


jurevans (approved):
Nice, LGTM!

neocybereth (approved):

PR #2032, opened 2025-05-21
chore: update namada chain registry

mateuszjasiuk:
For new tia on housefire.

Also changed `availableAssets` because it would query indexer for tokens, but if there are no token indexed yet(like new TIA), then it's not possible to make first ibc transfer into namada.


jurevans (approved):
LGTM!

PR #2033, opened 2025-05-22
feat: masp reward calculator

neocybereth:

### DIFF
--- a/apps/namadillo/src/App/Sidebars/MaspRewardCalculator.tsx
+++ b/apps/namadillo/src/App/Sidebars/MaspRewardCalculator.tsx
@@ -1,0 +1,113 @@
+import { Panel, SkeletonLoading, StyledSelectBox } from "@namada/components";
+import { AssetImage } from "App/Transfer/AssetImage";
+import { maspRewardsAtom } from "atoms/chain";
+import clsx from "clsx";
+import { useAtomValue } from "jotai";
+import { useState } from "react";
+import { MaspAssetRewards } from "types";
+
+export const MaspRewardCalculator = (): JSX.Element => {
+  const rewards = useAtomValue(maspRewardsAtom);
+  const [amount, setAmount] = useState<string>("");
+  const [selectedAsset, setSelectedAsset] = useState<
+    MaspAssetRewards | undefined
+  >(rewards.data?.[0] || undefined);
+
+  const estimatedRewards =
+    amount ?
+      (parseFloat(amount) * Number(selectedAsset?.maxRewardRate)).toFixed(2)
+    : "0.00";
+
+  return (
+    <Panel className={clsx("flex flex-col pt-2 pb-2 px-2")}>
+      <h2 className="uppercase text-[13px] text-center font-medium pb-0 pt-2">
+        MASP REWARDS CALCULATOR
+      </h2>
+      <div className="mt-3 flex flex-col gap-3">
+        {rewards.isLoading && (
+          <div className="flex flex-col gap-1">
+            <SkeletonLoading height="40px" width="100%" />
+            <SkeletonLoading height="40px" width="100%" />
+            <SkeletonLoading height="40px" width="100%" />
+          </div>
+        )}
+        {rewards.data && (
+          <>
+            <StyledSelectBox
+              value={selectedAsset?.asset.base || rewards.data[0].asset.base}
+              id="validator-filter"
+              containerProps={{
+                className: clsx(
+                  "text-sm flex-1 border-none bg-neutral-900 w-full rounded-sm",
+                  "px-4"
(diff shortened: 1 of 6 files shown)


neocybereth:
Screenshot 2025-05-22 at 2 42 31 PM
Screenshot 2025-05-22 at 2 42 38 PM

apps/namadillo/src/App/Sidebars/MaspRewardCalculator.tsx
```diff
@@ -0,0 +1,113 @@
...
+import { MaspAssetRewards } from "types";
+
+export const MaspRewardCalculator = (): JSX.Element => {
+  const rewards = useAtomValue(maspRewardsAtom);
+  const [amount, setAmount] = useState<string>("");
+  const [selectedAsset, setSelectedAsset] = useState<
+    MaspAssetRewards | undefined
+  >(rewards.data?.[0] || undefined);
+
+  const estimatedRewards =
```
mateuszjasiuk, line 16: Unfortunately it's not that straightforward :/ At some point you have to call `simulateShieldedRewards` function othrwise the result will be incorrect.
Also it looks like `simulate_shielded_rewards` function in `sdk/mod.rs` is a bit bugged, but easy to fix:

```rs
+        let epoch = rpc::query_masp_epoch(self.namada.client()).await?;
+
         let (_, masp_value) = shielded
             .convert_namada_amount_to_masp(
                 self.namada.client(),
-                // Masp epoch should not matter
-                MaspEpoch::zero(),
+                epoch,
```

neocybereth: Opened a [PR ](https://github.com/anoma/namada-sdkjs/pull/3)here. Thank you

apps/namadillo/src/App/Sidebars/MaspRewardCalculator.tsx
```diff
@@ -0,0 +1,261 @@
...
+        );
+        return;
+      }
+
+      setIsCalculating(true);
+      try {
+        const rewardsResult = await simulateShieldedRewards(
+          chainId,
+          assetAddress,
+          amount
```
mateuszjasiuk, line 75: So amount here is direct value from the input field(which is fine!) but `simulateShieldedRewards` expects amount in min(base) denomination. You can use `toBaseAmount` from utils to convert it

apps/namadillo/src/App/Sidebars/MaspRewardCalculator.tsx
```diff
@@ -0,0 +1,261 @@
...
+        console.error(
+          "Could not find address for asset:",
+          selectedAsset.asset.symbol
+        );
+        return;
+      }
+
+      setIsCalculating(true);
+      try {
+        const rewardsResult = await simulateShieldedRewards(
```
mateuszjasiuk, line 72: I'm curious if it would be beneficial to debounce this call, calculating rewards might take few seconds and debounce my reduce lag a bit. WDYT?

mateuszjasiuk (dismissed):
We have to change two things to make it work :)

First is in the comment and 2nd is that you fixed the `sumulate_shielded_rewards` in https://github.com/anoma/namada-sdkjs/pull/3 which is good, but namadillo does not use that package yet. So you also have to change it in `/packages/shared/lib/src/sdk/mod.ts` aswell. I missed it when reviewing that PR, sorry for that.

apps/namadillo/src/App/Sidebars/MaspRewardCalculator.tsx
neocybereth: Yup can do

apps/namadillo/src/App/Sidebars/MaspRewardCalculator.tsx
neocybereth: resolved

apps/namadillo/src/App/Sidebars/MaspRewardCalculator.tsx
```diff
@@ -0,0 +1,270 @@
...
+import {
+  chainAssetsMapAtom,
+  chainParametersAtom,
+  maspRewardsAtom,
+} from "atoms/chain";
+import { simulateShieldedRewards } from "atoms/staking/services";
+import BigNumber from "bignumber.js";
+import clsx from "clsx";
+import { useAtomValue } from "jotai";
+import { debounce } from "lodash";
```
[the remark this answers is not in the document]
neocybereth: "lodash.debounce": "^4.0.8",
    "lodash.isequal": "^4.5.0",
    
It's *only* the debounce and isEqual functions...not the whole library

PR #2049, opened 2025-05-28
fix: adjust gas for ibc deposits and terminate workers to free up memory

mateuszjasiuk:
- remove worker test
- terminate worker instances
- reduce estimated gas for ibc deposit slightly


pedrorezende (approved):
LGTM

jurevans (approved):
LGTM!

PR #2066, opened 2025-06-02
feat: update syncing view

neocybereth:
- Brings back syncing bar, gets rid of pulsing sync
- Updates syncing bar so that quick syncs are not shown
- Adds data to the Application sync (Heartbeat, Account Overview, Staking, Governance)
- Adds a container for errors to so that long errors dont overflow the container

### DIFF
--- a/apps/namadillo/src/App/Common/PulsingRing.tsx
+++ b/apps/namadillo/src/App/Common/PulsingRing.tsx
@@ -4,12 +4,9 @@ import { useScope } from "hooks/useScope";
 import { useRef } from "react";
 import { twMerge } from "tailwind-merge";
 
-type PulsingRingProps = { className?: string; size?: "small" | "large" };
+type PulsingRingProps = { className?: string };
 
-export const PulsingRing = ({
-  className,
-  size = "large",
-}: PulsingRingProps): JSX.Element => {
+export const PulsingRing = ({ className }: PulsingRingProps): JSX.Element => {
   const containerRef = useRef<HTMLDivElement>(null);
 
   useScope(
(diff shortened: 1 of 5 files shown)


apps/namadillo/src/App/Masp/ShieldedSyncProgress.tsx
```diff
@@ -0,0 +1,44 @@
...
+  const syncProgress = useAtomValue(shieldedSyncProgress);
+  const { isFetching } = useAtomValue(shieldedBalanceAtom);
+  const [showShieldedSync, setShowShieldedSync] = useState(false);
+
+  const roundedProgress = useMemo(() => {
+    // Only update when the progress changes by at least 1%
+    return Math.min(Math.floor(syncProgress * 100), 100);
+  }, [Math.floor(syncProgress * 100)]);
+
+  useEffect(() => {
```
neocybereth, line 16: This is to make sure we don't show the quick syncs.

apps/namadillo/src/App/Common/PulsingRing.tsx
```diff
@@ -4,12 +4,9 @@ import { useScope } from "hooks/useScope";
 import { useRef } from "react";
 import { twMerge } from "tailwind-merge";
 
-type PulsingRingProps = { className?: string; size?: "small" | "large" };
+type PulsingRingProps = { className?: string };
```
neocybereth, line 7: just returns this component back to it's original state

apps/namadillo/src/App/Masp/ShieldedSyncProgress.tsx
```diff
@@ -0,0 +1,44 @@
...
+    let timeout: ReturnType<typeof setTimeout> | undefined;
```
mateuszjasiuk, line 17: It looks like it shows the nodejs specific return type because of the node types that we include. Not sure if we need those but it's ok for now :)

mateuszjasiuk (approved):
Tested and looks good!

PR #2068, opened 2025-06-02
fix: increase additional percentage for simulated ibc transfer gas fee

mateuszjasiuk:
We match Keplr with this


neocybereth (approved):
LGTM

PR #2069, opened 2025-06-04
ci: removing stable branch integration

pedrorezende:
We're removing the last part failing, cause we won't stable branch for now


neocybereth (approved):
LGTM

PR #2074, opened 2025-06-05
feat: Fathom integration

pedrorezende:
We're adding the possibility of adding a [Fathom](https://app.usefathom.com/) tracker on Namadillo, so might an anonymous overview of the app usage.

### DIFF
--- a/apps/namadillo/src/hooks/useFathomTracker.tsx
+++ b/apps/namadillo/src/hooks/useFathomTracker.tsx
@@ -1,13 +1,41 @@
 import { defaultServerConfigAtom } from "atoms/settings";
-import * as Fathom from "fathom-client";
+import { load, trackEvent, trackPageview } from "fathom-client";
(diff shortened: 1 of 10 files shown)


apps/namadillo/src/hooks/useFathomTracker.tsx
```diff
@@ -0,0 +1,53 @@
...
+
+      trackPageview({
+        url: pathname,
+        referrer: document.referrer,
+      });
+    }
+  }, [fathomSiteId, initialized, pathname]);
+
+  return {
+    trackEvent: (eventName: string) => {
```
euharrison, line 47: We could have two hooks, one for the load, and another for the trackEvent, and have a global variable like `isEnabled` based on the initialization.
 I also ask myself if we execute the `trackEvent` without the `load` if it'll not just ignore everything

euharrison (approved):

PR #2075, opened 2025-06-05
feat: Account for changes in SDK libs v0.150.1

jurevans:
This PR simply updates `namada_sdk` and `namada_tx` to `v0.150.1`, and accounts for these changes in `@namada/shared`

### DIFF
--- a/packages/shared/lib/src/sdk/tx.rs
+++ b/packages/shared/lib/src/sdk/tx.rs
@@ -3,13 +3,14 @@ use std::str::FromStr;
 
 use gloo_utils::format::JsValueSerdeExt;
 use namada_sdk::borsh::{self, BorshDeserialize, BorshSerialize};
+use namada_sdk::collections::HashSet;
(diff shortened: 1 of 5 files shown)


packages/shared/lib/src/sdk/tx.rs
```diff
@@ -70,7 +71,11 @@ impl SigningData {
             None => None,
         };
 
-        let fee_payer = signing_tx_data.fee_payer.to_string();
+        let fee_payer = match signing_tx_data.fee_payer {
+            Either::Left(f) => f.0.to_string(),
+            Either::Right(_) => panic!("Fee payer must be present in SigningData"),
```
mateuszjasiuk, line 76: You can use this:
```rs
            Either::Right(_) => return Err(JsError::new("Fee payer must be a public key")),
```
panics in wasm can leak memory :rip:

jurevans: thanks!

mateuszjasiuk (approved):
Nice!

Issue #2096, opened 2025-06-07
Show hover-over popover on all Shielded Sync animations


mauricewbr:
Similar to the Shielded Sync animation in the navigation bar top-right, I suggest we also show the hover-over popover on the other Shielded Sync animations, e.g. next to the Total Balance.

PR #2101, opened 2025-06-07
feat: add tooltip to pulsing ring indicator

mauricewbr:
#2096


pedrorezende:
Closing since it was addressed on #2117

PR #2114, opened 2025-06-08
fix: Connect wallet larger click surface 

### LINKED ISSUE #2113: [Namadillo]: Connect wallet larger click surface
On the `Connect Wallet` function, the surface exposed as a button should be large and rectangular encompassing `Keplr` and the space in between `Connect`. Currently, one must hover over the `Connect` text itself only.

Image

neocybereth:
Closes #2113

### DIFF
--- a/apps/namadillo/src/App/Transfer/WalletCard.tsx
+++ b/apps/namadillo/src/App/Transfer/WalletCard.tsx
@@ -23,27 +23,32 @@ export const WalletCard = ({
     return wallet.downloadUrl.chrome;
   };
 
+  const handleClick = (): void => {
+    if (!installed) {
+      window.open(getDownloadUrl(), "_blank", "noopener,noreferrer");
+      return;
+    }
+    if (installed && !connected) return onConnect?.();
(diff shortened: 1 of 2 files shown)


apps/namadillo/src/App/Transfer/WalletCard.tsx
```diff
@@ -23,27 +23,32 @@ export const WalletCard = ({
...
+      window.open(getDownloadUrl(), "_blank", "noopener,noreferrer");
+      return;
+    }
+    if (installed && !connected) return onConnect?.();
+    if (installed && connected) return onSelect?.();
+  };
+
   return (
-    <div className="flex justify-between items-center w-full">
+    <div
```
euharrison, line 36: accessibility wise, we need to the tag `<a>` and `<button>`, so user can navigate with tab, right click and open it in a new tab, etc

we can use this pattern if it helps
```tsx
const Component = installed 
  ? ((props) => <a href={getDownloadUrl()} {...props} />) 
  : ((props) => <button onClick={handleClick} {...props} />)
  
<Component ....
```

PR #2115, opened 2025-06-08
fix: automagically select first available asset when selecting a chain

### LINKED ISSUE #2105: Auto select the token when selecting the chain
Image

neocybereth:
Closes #2105 

- makes sure fee is taken into account
- makes sure the rest of the component is aware of the selected asset


neocybereth:
https://github.com/user-attachments/assets/acea6c81-002d-44f8-992f-d9ad7ae2e4e5

pedrorezende (approved):
LGTM! Just fyi, we're investigating a solution of having only the token selection control available for users (until we have the designs of v2 implemented)

PR #2116, opened 2025-06-08
fix: remove sidebar shield sync indicator

euharrison:
Remove sidebar shield sync indicator that should have been done on https://github.com/anoma/namada-interface/pull/2084


pedrorezende (approved):
LGTM

PR #2117, opened 2025-06-09
fix: update total balance banner sync dot and transfer source masp sync info

### LINKED ISSUE #2096: Show hover-over popover on all Shielded Sync animations

neocybereth:
Closes #2096


pedrorezende (approved):
LGTM

PR #2137, opened 2025-06-13
fix: update txn amounts for shielding/unshielding transfers

### LINKED ISSUE #2136: History not showing amount transferred just fees paid.
### What happened?

Image

### Steps to Reproduce

1) Go to history
2) check transfer data

### Relevant log output

```shell

```

### Debug Information

```shell

```

neocybereth:
Closes #2136

### DIFF
--- a/apps/namadillo/src/App/Transactions/TransactionCard.tsx
+++ b/apps/namadillo/src/App/Transactions/TransactionCard.tsx
@@ -86,25 +86,25 @@ const getTransactionInfo = (
 
   const parsed = typeof tx.data === "string" ? JSON.parse(tx.data) : tx.data;
   const sections: RawDataSection[] = Array.isArray(parsed) ? parsed : [parsed];
+  const target = sections.find((s) => s.targets?.length);
+  const source = sections.find((s) => s.sources?.length);
 
-  let receiver: string | undefined;
   let amount: BigNumber | undefined;
+  let receiver: string | undefined;
 
-  // Find first section with targets or sources that has amount
-  const targetSection = sections.find((sec) => sec.targets?.[0]?.amount);
-  const sourceSection = sections.find((sec) => sec.sources?.[0]?.amount);
-
-  if (targetSection?.targets?.[0]) {
-    amount = new BigNumber(targetSection.targets[0].amount);
-    receiver = targetSection.targets[0].owner;
+  if (target?.targets) {
+    const mainTarget = target.targets.reduce((max, cur) =>
+      new BigNumber(cur.amount).isGreaterThan(max.amount) ? cur : max
+    );
+    amount = new BigNumber(mainTarget.amount);
+    receiver = mainTarget.owner;
+  }
+  // fall back to sources only when we had no targets
+  if (!amount && source?.sources?.[0]) {
+    amount = new BigNumber(source.sources[0].amount);
   }
 
-  if (!amount && sourceSection?.sources?.[0]) {
-    amount = new BigNumber(sourceSection.sources[0].amount);
-  }
-
-  // Find sender from any section with sources
-  const sender = sections.find((sec) => sec.sources?.[0]?.owner)?.sources?.[0]
+  const sender = sections.find((s) => s.sources?.[0]?.owner)?.sources?.[0]
     ?.owner;
 
   return amount ? { amount, sender, receiver } : undefined;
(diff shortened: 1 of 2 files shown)


apps/namadillo/src/App/Transactions/TransactionCard.tsx
```diff
@@ -86,25 +86,25 @@ const getTransactionInfo = (
...
+      new BigNumber(cur.amount).isGreaterThan(max.amount) ? cur : max
+    );
+    amount = new BigNumber(mainTarget.amount);
+    receiver = mainTarget.owner;
   }
-
-  if (!amount && sourceSection?.sources?.[0]) {
-    amount = new BigNumber(sourceSection.sources[0].amount);
+  // fall back to sources only when we had no targets
+  if (!amount && source?.sources?.[0]) {
```
euharrison, line 103: @neocybereth shouldn't we do the same sum with reduce with the sources?

apps/namadillo/src/App/Transactions/TransactionCard.tsx
```diff
@@ -86,25 +86,25 @@ const getTransactionInfo = (
...
+    amount = new BigNumber(source.sources[0].amount);
   }
 
-  // Find sender from any section with sources
-  const sender = sections.find((sec) => sec.sources?.[0]?.owner)?.sources?.[0]
+  const sender = sections.find((s) => s.sources?.[0]?.owner)?.sources?.[0]
```
euharrison, line 107: shouldn't we move the `?.sources?.[0]` inside the sender? could the sender be still undefined?
```
  const sender = sections.find((s) => s.sources?.[0]?.owner?.sources?.[0])
```

I didn't test on real world, only thinking about the pure js logic

apps/namadillo/src/App/Transactions/TransactionCard.tsx
```diff
@@ -125,7 +125,9 @@ export const TransactionCard = ({
     : getTransactionInfo(transaction);
   const baseAmount =
     asset && txnInfo?.amount ?
-      toDisplayAmount(asset, txnInfo.amount)
+      isBondingTransaction ? toDisplayAmount(asset, txnInfo.amount)
+      : isNamadaAsset(asset) ? txnInfo.amount
+      : toDisplayAmount(asset, txnInfo.amount)
```
euharrison, line 130: this is a ternary hell to read, maybe we could create a helper function with "return earlier" to handle this

apps/namadillo/src/App/Transactions/TransactionCard.tsx
neocybereth: Can confirm, sources only ever have one option.

apps/namadillo/src/App/Transactions/TransactionCard.tsx
neocybereth: There's always a sender. 

So if I did: 

```  const sender = sections.find((s) => s.sources?.[0]?.owner?.sources?.[0])
```

I'd still have to write sender?.source?.[0]?.owner to extract the value, so that's what I'm doing above but shorthand. Make sense?

apps/namadillo/src/App/Transactions/TransactionCard.tsx
```diff
@@ -136,6 +132,14 @@ export const TransactionCard = ({
   const validators = useAtomValue(allValidatorsAtom);
   const validator = validators?.data?.find((v) => v.address === receiver);
 
+  const getBaseAmount = (): BigNumber | undefined => {
+    if (asset && txnInfo?.amount) {
+      if (isBondingTransaction) return toDisplayAmount(asset, txnInfo.amount);
+      if (isNamadaAsset(asset)) return txnInfo.amount;
```
euharrison, line 138: ok, I checked the indexer endpoint and it's incompatible, transparent transfers are saved as base 6 and bond are base 0. I think we could request for the indexer team to update this later and use only a common pattern. At least we have the fix on the front :)

euharrison (approved):
LGTM! only code comments, feature is working 👍

apps/namadillo/src/App/Transactions/TransactionCard.tsx
euharrison: I just mean, if we add the `?.` on `sources?.[0]`, we are expecting that, in a border case, we could have `sources === undefined`

so, if this is the case, should we consider another `s.sources?.[0]?.owner` to be searched?

I do understand that we do not expect this on happy paths, but if we are 100% sure about all cases, we could simply remove all `?`, no? As we don't, we could be aligned on code on the same idea

PR #2170, opened 2025-06-18
fix: shielded balance comma separator

euharrison:
On home, show the comma separator for Shielded balance

Before

Screenshot 2025-06-18 at 15 15 00

After 

Screenshot 2025-06-18 at 15 14 29


neocybereth (approved):
LGTM

PR #2176, opened 2025-06-19
feat: shielded sync retry

mateuszjasiuk:
Retry strategies  for shielded sync


neocybereth:
https://github.com/user-attachments/assets/000aabcf-c7f6-461b-baf1-57e8e7465a2d

Just FYI I was playing with this to see if it'd resolve the IBC Deposit issue and found this. Whenever you go to Deposit and switch assets is stays stucks on an old asset as the destination chain.

PR #2182, opened 2025-06-20
fix: displaying housefire nam in mainnet osmosis assets

mateuszjasiuk:
How to test:
- send some nam on housefire to osmosis
- go to ibc deposit and check if you can see nam(you should be able to)
- change providers to mainnet
- go to ibc deposit and check if you can see nam(you shouldnt be able to)


jurevans (approved):
LGTM!

PR #2181, opened 2025-06-20
fix: update shielded txn receipt

neocybereth:

PR #2183, opened 2025-06-20
fix: denominate nam before ibc withdraw

mateuszjasiuk:
How to test:
- try to ibc transparent withdraw 100 nam
- see if no error
- see if 100 was subtracted from your balance
- see if tx history shows correct values(it does not for sproutstake indexer, but our internal one works fine)
- try the same with ibc shielded withdraw

### DIFF
--- a/apps/namadillo/src/atoms/transfer/services.ts
+++ b/apps/namadillo/src/atoms/transfer/services.ts
@@ -42,11 +42,11 @@ export type WorkerTransferParams = {
 
 const workerBuildTxPair = async <T>({
   rpcUrl,
-  token,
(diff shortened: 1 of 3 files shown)


neocybereth (approved):
Screenshot 2025-06-20 at 11 45 21 PM
Screenshot 2025-06-20 at 11 46 27 PM


One is a Withdraw from Transparent and the other from Shielded. Don't mind the TransferSource component that's fixed by [this](https://github.com/anoma/namada-interface/pull/2181)

jurevans (approved):
LGTM!

PR #2189, opened 2025-06-23
fix: update unshield button with proper route

### LINKED ISSUE #2168: When you click 'unshield' i should get the choice of IBC or Internal + Prefill module
- When clicking 'unshield' on an asset it should bring up the IBC/ Internal selector
- It should then prefill that selected token in the corresponding transfer UI

Image

Image

neocybereth:
Closes #2168


neocybereth:
https://github.com/user-attachments/assets/500b57db-c848-4919-a191-8bce22e7f6ef

mateuszjasiuk (approved):

PR #2202, opened 2025-07-01
fix: remove deposit and withdraw tabs for transaction animation

neocybereth:


neocybereth:
https://github.com/user-attachments/assets/20e83e6d-dcff-48bc-8967-ba2d75e78850

mateuszjasiuk (approved):

PR #2212, opened 2025-07-04
fix: fees in ibc deposits

mateuszjasiuk:


neocybereth (approved):
On mobile so can't test but code looks fine to me

PR #2214, opened 2025-07-07
fix: bonded amount dollar value

mateuszjasiuk:
Also fixes a problem in transfer tab with shielded transfers etc.


neocybereth (approved):

neocybereth:
LGTM

PR #2261, opened 2025-07-28
fix: add staking amount to Total Balance banner

### LINKED ISSUE #2255: [Namadillo] Display total balance including staked NAM
The Total Balance in Namada data in the `Manage Assets` Home tab does not account for staked NAM. It should, and then perhaps the liquid / unstaked value can be displayed to the right of it in parenthese or something. 

Image

neocybereth:
Closes #2255


mateuszjasiuk (commented):
This looks good,, I'm just curious if we should not change the amounts in the staking tab tho. Because atm it only shows bonded amount and for total we also take unbonds and withdraws into consideration.
Image
@brentstone @ChrisHoltDesign

neocybereth:
> This looks good,, I'm just curious if we should also change the amounts in the staking tab tho. Because atm it only shows bonded amount and for total we also take unbonds and withdraws into consideration. Image @brentstone @ChrisHoltDesign

image

image

PR #2280, opened 2025-08-18
fix: add alias - shielded for all gas spendnig keys

neocybereth:

### DIFF
--- a/apps/namadillo/src/App/Common/TransactionReceipt.tsx
+++ b/apps/namadillo/src/App/Common/TransactionReceipt.tsx
@@ -66,8 +66,6 @@ const TransferTransactionReceipt = ({
   }, [transaction]);
 
   const getEncodedViewingKey = (address: string): string => {
-    const canDecode = PseudoExtendedKey.can_decode(address);
-    if (!canDecode) return "";
     const decodedPseudokey = PseudoExtendedKey.decode(address);
     const encodedViewingKey = decodedPseudokey.to_viewing_key().encode();
     return encodedViewingKey;
--- a/packages/shared/lib/src/types/masp.rs
+++ b/packages/shared/lib/src/types/masp.rs
@@ -99,7 +99,7 @@ impl PseudoExtendedKey {
 
     pub fn can_decode(encoded: String) -> bool {
         let decoded = hex::decode(encoded);
-        Ok(decoded.is_ok())
+        decoded.is_ok()
     }
 
     pub fn decode(encoded: String) -> Result<PseudoExtendedKey, JsError> {


apps/namadillo/src/App/Common/TransactionReceipt.tsx
```diff
@@ -75,11 +80,17 @@ const TransferTransactionReceipt = ({
       transaction.destinationAddress || ""
     );
   }, [transaction]);
+  const decodedPseudokey = PseudoExtendedKey.decode(
+    transaction.sourceAddress || ""
+  );
+  const encodedViewingKey = decodedPseudokey.to_viewing_key().encode();
```
mateuszjasiuk, line 78: to_viewing_key does not exist ;)

apps/namadillo/src/App/Common/TransactionReceipt.tsx
```diff
@@ -60,9 +72,17 @@ const TransferTransactionReceipt = ({
       transaction.destinationAddress || ""
     );
   }, [transaction]);
+  const decodedPseudokey = PseudoExtendedKey.decode(
```
mateuszjasiuk, line 75: This can throw. Also I think maybe instead of try catching it's worth adding `can_decode` function to the PseudoExtendedKey struct and just return a boolean

mateuszjasiuk (changes requested):

packages/shared/lib/src/types/masp.rs
```diff
@@ -97,6 +97,11 @@ impl PseudoExtendedKey {
         hex::encode(borsh::to_vec(&self.0).expect("Serializing PseudoExtendedKey should not fail!"))
     }
 
+    pub fn can_decode(encoded: String) -> Result<bool, JsError> {
```
mateuszjasiuk, line 100: You can just return `bool` here, should be ok!

apps/namadillo/src/App/Common/TransactionReceipt.tsx
```diff
@@ -48,6 +49,11 @@ const TransferTransactionReceipt = ({
     return chain;
   };
 
+  const isExtendedKey = (address: string): boolean => {
+    if (!address) return false;
```
mateuszjasiuk, line 53: Shouldn't `can_decode` return false for empty string anyway?

apps/namadillo/src/App/Common/TransactionReceipt.tsx
```diff
@@ -61,8 +67,20 @@ const TransferTransactionReceipt = ({
     );
   }, [transaction]);
 
+  const getEncodedViewingKey = (address: string): string => {
+    const canDecode = PseudoExtendedKey.can_decode(address);
+    if (!canDecode) return "";
```
mateuszjasiuk, line 72: Maybe instead of checking `canDecode` twice you can just move this logic:
```
isExtendedKey(transaction.sourceAddress || "") ?
                  getEncodedViewingKey(transaction.sourceAddress || "")
                : transaction.sourceAddress
```
to a separate function?

mateuszjasiuk (approved):
Nice! I think we can wrap `getChain`, `isExtendedKey ` and `getEncodedViewingKey` in the useCallback also

EDIT: btw I think you need to rebase the branch with main for CI to pass

mateuszjasiuk:
Also now that refactor is merged you need to add some of the changes to this [repo ](https://github.com/namada-net/namada-sdkjs)

PR #2288, opened 2025-08-20
refactor: use @namada/sdk package in the extension

mateuszjasiuk:
- replaces imports
- fixes tests
- removes all the wasm build code
- removes shared, crypto and sdk packages

### DIFF
--- a/apps/extension/src/Setup/Ledger/LedgerConnect.tsx
+++ b/apps/extension/src/Setup/Ledger/LedgerConnect.tsx
@@ -2,7 +2,7 @@ import { chains } from "@namada/chains";
 import { ActionButton, Alert, Image, Stack } from "@namada/components";
 import {
   ExtendedViewingKey,
-  initSdk,
+  init,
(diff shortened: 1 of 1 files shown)


apps/extension/src/Setup/Ledger/LedgerConnect.tsx
```diff
@@ -80,8 +80,11 @@ export const LedgerConnect: React.FC<Props> = ({
         setCurrentApprovalStep(3);
         const { ak, nsk } = await ledger.getProofGenerationKey(path);
 
-        // SDK wasm init must be called
-        await initWasm();
+        // We only do this to initialize wasm, props don't matter
```
jurevans, line 83: Just a note, though I think this is fine: There is an `init` that is directly exported (same with `*/inline`) that doesn't initialize Sdk, but since we don't use the inline imports in extension, you'd have to pass the wasm as an arg :D Maybe that can be improved in the future, but let's keep it like this for now!

jurevans (approved):
Nice, LGTM!

PR #2290, opened 2025-08-20
fix: build fixes

neocybereth:


mateuszjasiuk:
We will remove crypto soon enough anyway :)

PR #2297, opened 2025-09-15
fix: fix for clickjacking that mirrors namada.net

neocybereth:
Mirror's Harri's changes here except for Docker instead of a static host (Netlify): https://github.com/heliaxdev/namada.net/pull/167/files


euharrison (approved):
Nice! I didn't have the opportunity to run the docker to test. But the code looks good

Please, test before publishing a new version 🙏

neocybereth:
Screenshot 2025-09-18 at 9 04 25 AM

Tested various transactions and clicked around on all sorts of views and nothing seems to be blocked on the dev environment with this docker image.

Issue #2319, opened 2025-11-08
[Namadillo]: /shield route in Shield Assets button


zenodeapp:
### What happened?

Hey,

<img width="232" height="187" alt="Image" src="https://github.com/user-attachments/assets/2fd1a51f-c39f-4229-b034-7cf6171b2275" />

This button goes to an invalid route (also a couple more, but see the PRs and info below for more detail). It did get me to wonder if you guys are keeping the `/masp/shield` route for a reason? Or should actually everything else get rerouted to `/shield` instead to keep it all aligned with the rest of the routes?

### Steps to Reproduce

1. Manage Assets
2. Unshielded Assets-tab
3. Shield Assets button goes to the incorrect route (404-error).

### Relevant log output

```shell

```

### Debug Information

```shell
## Navigator
------------------
User Agent: Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/142.0.0.0 Safari/537.36
Screen Resolution: 2560px x 1289px / DPI: 1
Sat Nov 08 2025 04:48:45 GMT+0100 (Central European Standard Time)

## Version Info
------------------
Namadillo Version: 1.32.1
SDK Version: 0.22.0
Keychain Version: 0.8.1
Indexer Version: 4.1.0
Masp Indexer Version: 1.4.7

## MASP Status
------------------
MASP Indexer URL: undefined
Last Shielded Sync: Sat Nov 08 2025 05:14:59 GMT+0100 (Central European Standard Time)
Is Shielded Sync complete? Yes

## Indexer Status
------------------
Indexer URL: 
Epoch: 7463
Block height: 3373507
Crawler Info: [{"name":"chain","timestamp":1762575298,"lastProcessedBlockHeight":3373508},{"name":"governance","timestamp":1762575294,"lastProcessedBlockHeight":null},{"name":"parameters","timestamp":1762575294,"lastProcessedBlockHeight":null},{"name":"pos","timestamp":1762575298,"lastProcessedBlockHeight":null},{"name":"rewards","timestamp":1762575294,"lastProcessedBlockHeight":null},{"name":"transactions","timestamp":1762575294,"lastProcessedBlockHeight":3373507}]

## Other Info
------------------
RPC Url: undefined
Account type: mnemonic
```

zenodeapp:
[...]

PR #2320, opened 2025-11-08
fix: /shield and /unshield routes

zenodeapp:
One of the ways for fixing #2319.

- This goes the /shield /unshield route omitting /masp.

This keeps it in alignment with the rest of the routes. But if the preference goes to keeping /masp then there's #2321.


neocybereth (approved):
looks great! nice work.

PR #2323, opened 2025-11-08
fix: no duplicate address options and make sure alias matches for cus…

neocybereth:
Screenshot 2025-11-08 at 10 59 20 PM
Screenshot 2025-11-08 at 10 59 13 PM
Screenshot 2025-11-08 at 10 59 23 PM


mateuszjasiuk (approved):
LGTM!
Maybe it's also worth filtering out selected account from the list of recent ones?

PR #2343, opened 2025-11-20
fix: various fixes test suite

neocybereth:
- [x] Stride selectable
- [x] Make Keplr assets not show in destination when Keplr token is source
- [x] Add better caching for fetching of Keplr assets


apps/namadillo/src/integrations/utils.ts
```diff
@@ -44,6 +44,7 @@ export const findRegistryByChainId = (
 };
 
 export const getChainFromAddress = (address: string): Chain | undefined => {
+  const chains = getAvailableChains();
```
neocybereth: Stride wasn't selectable b/c here I was using the `chains` from the registry and it was getting matched with the `stratos` chain instead of `stride`.

PR #2358, opened 2025-12-05
feat: add data collection permissions to the Firefox manifest

mateuszjasiuk:


jurevans (approved):
Looks good! This should suffice I think

